PR #1, opened 2019-06-17
Create environments for kubernetes and openshift manifests

metalmatze:
Instead of generating everything into `/manifests` I have now split both environments up into _kubernetes_ and _openshift_. _openshift_ depends on _kubernetes_, so for most changes we might only need to update the Kubernetes environment.

In the end, we generate a folder `/environments/kubernetes/manifests/` that contains a working set of manifests for Kubernetes and we generate a `Template` file for OpenShift that has templated `${NAMESPACE}`, `${IMAGE}` and others. I know, we template the template for the OpenShift one and hopefully this is something we can get rid of in the future, but for now it's a requirement to have these.

/cc @aditya-konarde @squat @brancz @ant31 

TODO:
* [x] Pass all S3 configuration via env vars for automatic Vault rotation and not via Thanos config
  * [ ] ~Remove objectStorageConfig~
* [ ] ...


environments/openshift/main.jsonnet
```diff
@@ -0,0 +1,52 @@
+local list = import 'telemeter/lib/list.libsonnet';
```
brancz: This is kind of a weird dep but I’m ok with it as long as we acknowledge it.

environments/openshift/main.jsonnet
```diff
@@ -0,0 +1,52 @@
...
+          for name in std.objectFields(t.querier)
+        } + {
+          ['store-' + name]: t.store[name]
+          for name in std.objectFields(t.store)
+        };
+
+        list.asList('thanos', objects, [
+          {
+            name: 'NAMESPACE',
+            value: 'telemeter',
```
squat: How do these values interact with the config values? Should we have:
```
value: $._config.namespace,
```
as the default value for the variable as we do in the telemeter repo?

metalmatze: Yes. But for simplicity reasons I've set `$._config.namespace` to `${NAMESPACE}`. Especially overwriting the `dnssrv+` flags becomes quite tricky otherwise.
In the end, for all kubernetes/ksonnet things, we should probably not even use `$._config`.

brancz:
I quite like this approach for kubernetes things. For monitoring-mixins we'll have to think of something else, as there are intentionally shared fields for those. For monitoring-mixins I'm currently thinking `$._config.<mixin-name>` for namespacing each config.

metalmatze:
Yes. This was specifically for Kubernetes.
The monitoring mixin approach of using `$._config.<mixin-name>` is something that I've done a few times already. Guess it only makes sense to go forward with that.

brancz:
Do you or @ant31 want to create a PR to propose the namespaced config to the monitoring-mixin design documentation?

metalmatze:
We should discuss the monitoring mixins in person before and do some more PoC things. Happy to have @ant31 do it, as I am probably quite biased. Let's discuss this somewhere else! :blush: 

@aditya-konarde, PTAL and let me know if there's anything missing from the OpenShift template that you need.

brancz:
I think being opinionated is our strength :)

environments/openshift/manifests/thanos-template.yaml
```diff
@@ -0,0 +1,187 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: thanos
+objects:
+- apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    labels:
+      app: thanos-querier
```
squat: We need to update these labels to the new standard. Can be a follow up.

PR #6, opened 2019-07-09
Add Telemeter to the stack

metalmatze:
Currently we need to run patched version of Telemeter and Thanos for the forwarding and hashrings to work.

/cc @kakkoyun @brancz @aditya-konarde


brancz:
what are the thanos patches that we need to apply on top of current thanos master?

brancz:
everything looks good, just trying to understand the different images (telemeter can be just the latest image once forwarding is merged, no?)

metalmatze:
> what are the thanos patches that we need to apply on top of current thanos master?

We want to make use of the new hashring capabilities that was added to the receive component. But you're right, that's not really the concern of this PR and we can still use the upstream images tagged with `master-%y-%m-%d-%hash`. I'll do that with a new PR that adds the thanos-receive-controller! :+1:

environments/kubernetes/telemeter.libsonnet
```diff
@@ -30,4 +30,16 @@
...
   },
+} + {
+  local ts = super.telemeterServer,
+  telemeterServer:: {
+    [k]: ts[k]
+    for k in std.objectFields(ts)
+    // This array must be sorted for `std.setMember` to work.
+    if !std.setMember(k, [
+      'secret',
+      'serviceMonitor',
```
brancz: why remove this serviceMonitor? We still want to scrape the telemeter servers' metrics no?

metalmatze: Yes. But I talked to @aditya-konarde and we want to add these ServiceMonitors to their app-sre Prometheus. :blush:

brancz: but we would still want the ServiceMonitor to be created in this namespace, no?

metalmatze: It's a limitation of OpenShift 3.11 Prometheus Operator being on v0.22

brancz:
lgtm

PR #7, opened 2019-07-11
Add Telemeter to Openshift Templates

metalmatze:
The two files for telemeter and kube-thanos expose their own OpenShift templates:
```
environments/openshift/telemeter.libsonnet
environments/openshift/kube-thanos.libsonnet
```

Then in `environments/openshift/main.libsonnet` we just merge both Templates together.

/cc @brancz @aditya-konarde

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -187,12 +187,159 @@ objects:
         - emptyDir: {}
           name: data
     volumeClaimTemplates: []
+- apiVersion: v1
+  kind: Service
+  metadata:
+    annotations:
+      service.alpha.openshift.io/serving-cert-secret-name: telemeter-server-shared
+    labels:
+      k8s-app: telemeter-server
+    name: telemeter-server
+    namespace: ${NAMESPACE}
+  spec:
+    clusterIP: None
+    ports:
+    - name: external
+      port: 8443
+      targetPort: external
+    - name: internal
+      port: 8081
+      targetPort: internal
+    - name: cluster
+      port: 8082
+      targetPort: cluster
+    selector:
+      k8s-app: telemeter-server
+- apiVersion: v1
+  kind: ServiceAccount
+  metadata:
+    name: telemeter-server
+    namespace: ${NAMESPACE}
+- apiVersion: apps/v1beta2
+  kind: StatefulSet
+  metadata:
+    name: telemeter-server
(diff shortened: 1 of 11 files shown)


brancz:
if you've tested this, then this lgtm! :+1:

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -0,0 +1,369 @@
...
+    name: telemeter-server
+    namespace: ${NAMESPACE}
+- apiVersion: apps/v1beta2
+  kind: StatefulSet
+  metadata:
+    name: telemeter-server
+    namespace: ${NAMESPACE}
+  spec:
+    podManagementPolicy: Parallel
+    replicas: 3
```
aditya-konarde: Should we parameterize this? Telemeter in production needs to be using 10 replicas, for example.

metalmatze: This was not parameterized in the current Telemeter version. We should definitely do that. Can we do that in a follow-up PR? :)

https://github.com/openshift/telemeter/blob/25f46281db14a545d07b1b9fcca7105b4e27c629/manifests/server/list.yaml#L101

brancz:
this has my blessing if it has @aditya-konarde 's :)

aditya-konarde:
LGTM :)

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -0,0 +1,432 @@
...
+      app: thanos-receive
+- apiVersion: apps/v1
+  kind: StatefulSet
+  metadata:
+    labels:
+      app: thanos-receive
+    name: thanos-receive
+    namespace: ${NAMESPACE}
+  spec:
+    replicas: ${{THANOS_RECEIVE_REPLICAS}}
```
squat: This seems funny. Shouldn’t there be just one set of braces?

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -0,0 +1,432 @@
...
+      app: thanos-store
+- apiVersion: apps/v1
+  kind: StatefulSet
+  metadata:
+    labels:
+      app: thanos-store
+    name: thanos-store
+    namespace: ${NAMESPACE}
+  spec:
+    replicas: ${{THANOS_STORE_REPLICAS}}
```
squat: Here too

PR #10, opened 2019-07-15
Add the thanos-receive-controller with a catch all tenancy

metalmatze:
Add the thanos-receive-controller to operator the current thanos-receive statefulset with a catch-all tenancy.

/cc @squat @kakkoyun @aditya-konarde @brancz

### DIFF
--- a/components/thanos-receive-controller.libsonnet
+++ b/components/thanos-receive-controller.libsonnet
@@ -3,6 +3,50 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
 {
   thanos+:: {
     receiveController: {
+      serviceAccount:
+        local sa = k.core.v1.serviceAccount;
+
+        sa.new() +
+        sa.mixin.metadata.withName('thanos-receive-controller') +
+        sa.mixin.metadata.withNamespace('observatorium'),
+
+      role:
+        local role = k.rbac.v1.role;
+        local rules = role.rulesType;
+
+        role.new() +
+        role.mixin.metadata.withName('thanos-receive-controller') +
+        role.mixin.metadata.withNamespace('observatorium') +
+        role.withRules([
+          rules.new() +
+          rules.withApiGroups(['']) +
+          rules.withResources([
+            'configmaps',
+          ]) +
+          rules.withVerbs(['list', 'watch', 'get', 'create', 'update']),
+          rules.new() +
+          rules.withApiGroups(['apps']) +
+          rules.withResources([
+            'statefulsets',
+          ]) +
+          rules.withVerbs(['list', 'watch', 'get']),
+        ]),
+
+      roleBinding:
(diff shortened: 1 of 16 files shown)


tenants.libsonnet
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    hashring: 'thanos-receive',
```
squat: let's give this hashring a more descriptive name as every hashring will be made of `thanos-receive`. maybe `default` as it is the catch-all

metalmatze:
Pushed all the updates including renaming the statefulset to `thanos-receive-default`, add ServiceAccount, Role and RoleBinding and then update to Thanos `v0.6.0-rc.0` as default version.

/cc @squat

components/thanos-receive-controller.libsonnet
```diff
@@ -0,0 +1,97 @@
...
+        ) +
+        service.mixin.metadata.withNamespace('observatorium'),
+
+      deployment:
+        local deployment = k.apps.v1.deployment;
+        local container = deployment.mixin.spec.template.spec.containersType;
+        local env = container.envType;
+
+        local c =
+          container.new($.thanos.receiveController.deployment.metadata.name, 'quay.io/observatorium/thanos-receive-controller:latest') +
```
bwplotka: Can we pin to any version at some point?

squat: yes would be nice  to add proper master+hash tagging to the build system like thanos has

environments/kubernetes/kube-thanos.libsonnet
```diff
@@ -38,7 +43,11 @@ local deployment = k.apps.v1.deployment;
         service.mixin.metadata.withNamespace(namespace),
       statefulSet+: {
         metadata+: {
+          name: 'thanos-receive-default',
```
bwplotka: I think making store `sts.mixin.spec.withReplicas(3),` is not that helpful as you have headless service with `dnssrv` option. This means that you treat each store as separate source that will be touch 100% time. If request will bring it down all would go down. However it increase availability on other crashes / restarts (: 

We explain this more: https://github.com/improbable-eng/thanos/issues/1083

environments/kubernetes/kube-thanos.libsonnet
```diff
@@ -38,7 +43,11 @@ local deployment = k.apps.v1.deployment;
...
           namespace: namespace,
+          labels+: {
+            'controller.receive.thanos.io': 'thanos-receive-controller',
```
bwplotka: Might be worth to configure it explicitly

metalmatze:
Thanks for the review @bwplotka. I've created some issues to track those comments.

components/thanos-receive-controller.libsonnet
bwplotka: Feel free to copy the logic for that from Makefile + .circleci (:

PR #16, opened 2019-07-16
Remove unnecessary metrics from whitelist

smarterclayton:
The removed metrics are either no longer produced (payload_errors)
or are or should be redundant with existing metrics (up and ALERTS
for image registry and etcd).

Depends on https://github.com/openshift/telemeter/pull/197 and
https://github.com/openshift/telemeter/pull/198


metalmatze:
This should still be done in the Telemeter repository, right @squat? Then we can bump the dependency in this repo like we did in https://github.com/observatorium/configuration/pull/12.

squat:
Yes @metalmatze that is the correct flow. This changed depended on openshift/telemeter#197 and openshift/telemeter#197 so it didn't bump jsonnet yet. Now that those pull requests are in, we can update this PR to actually be a jsonent + generation change. I'll push to this branch directly.

PR #26, opened 2019-07-23
environments/openshift: Use telemeter-server serviceAccount for Thanos querier

metalmatze:
As it seems complicated to get another ServiceAccount, let's just use telemeter-server as serviceAccount to make the SAR. 

/cc @aditya-konarde @squat @brancz


brancz:
lgtm for now, let's see how we can use local subject access reviews in the future

PR #30, opened 2019-07-25
Bug 1731232: Add cluster infrastructure metrics for capture

smarterclayton:
These metrics assist in triaging problems by cloud provider and
enablement of tech preview features.


metalmatze:
I've generated the latest `observatorium-template.yaml`. Just run `./build.sh` in the root in the future :)

metalmatze:
Ah wait. You still need to go through openshift/telemeter and add the metrics there. Then we can bump it's dependency by running `jb update` and then `./build.sh` in this repo.

squat:
closing in favor of #32

PR #31, opened 2019-07-26
Generate SRE ServiceMonitors for staging and production

metalmatze:
This has potential to be looped over in jsonnet.
Feel free to improve this. Otherwise I'm continuing it in the next days.

/cc @squat @aditya-konarde


environments/sre/servicemonitors/observatorium-thanos-querier-production.servicemonitor.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    prometheus: app-sre
+  name: observatorium-thanos-querier
```
aditya-konarde: I realized that since these are landing in the same namespace in the end, we'd need to suffix the `name` field with `-stage` and `production`

PR #42, opened 2019-08-15
Add dashboards for Thanos

kakkoyun:


environments/sre/grafana.jsonnet
```diff
@@ -0,0 +1,31 @@
...
+  _config+:: {
+    thanosQuerier: 'thanos-querier',
+    thanosStore: 'thanos-store',
+    thanosReceive: 'thanos-receive',
+    thanosRule: 'thanos-rule',
+    thanosCompact: 'thanos-compact',
+    thanosSidecar: 'thanos-sidecar',
+
+    thanosQuerierSelector: 'job="%s"' % self.thanosQuerier,
+    thanosStoreSelector: 'job="%s"' % self.thanosStore,
```
brancz: Any reason why we template here instead of directly specifying the selector?

kakkoyun: It's for reusing, I needed names of jobs for pod selector and when I introduced those I've just reused them 
https://github.com/metalmatze/kube-thanos/blob/a7da914e6dbff3472031e169131cb88efa6cda26/jsonnet/thanos-mixin/dashboards/querier.libsonnet#L119

kakkoyun:
cc @aditya-konarde 

I'd love to hear your opinions as well

environments/sre/grafana.jsonnet
metalmatze: I think it's fine to redefine them here. That's one of the cases where we actually want to override and be in control.

PR #43, opened 2019-08-16
Copy modified Thanos Queriers to separate tenants

metalmatze:
This copies the Querier Deployment and Service and modifies the labels to separate tenants when querying.

/cc @squat @kakkoyun @aditya-konarde @brancz

### DIFF
--- a/environments/openshift/kube-thanos.libsonnet
+++ b/environments/openshift/kube-thanos.libsonnet
@@ -115,6 +115,44 @@ local list = import 'telemeter/lib/list.libsonnet';
           volume.fromSecret('secret-querier-tls', 'querier-tls'),
           volume.fromSecret('secret-querier-proxy', 'querier-proxy'),
         ]),
+
+      datahubService:
+        $.thanos.querier.service {
+          metadata: {
+            name: 'thanos-querier-datahub',
+            labels+: {
+              'observatorium.io/tenant': 'datahub',
+            },
+          },
+          spec+: {
+            selector+: {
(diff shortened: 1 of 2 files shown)


brancz:
did we ever actually decide to have a separate querier for datahub?

environments/openshift/kube-thanos.libsonnet
```diff
@@ -115,6 +115,44 @@ local list = import 'telemeter/lib/list.libsonnet';
...
+          metadata+: {
+            name: 'thanos-querier-datahub',
+            labels+: {
+              'observatorium.io/tenant': 'datahub',
+            },
+          },
+          spec+: {
+            selector+: {
+              matchLabels+: {
+                'observatorium.io/tenant': 'datahub',
```
squat: right now the other deployment will select these pods too because the other selector is not specific enough

environments/openshift/kube-thanos.libsonnet
```diff
@@ -115,6 +115,44 @@ local list = import 'telemeter/lib/list.libsonnet';
...
+        $.thanos.querier.service {
+          metadata: {
+            name: 'thanos-querier-datahub',
+            labels+: {
+              'observatorium.io/tenant': 'datahub',
+            },
+          },
+          spec+: {
+            selector+: {
+              'observatorium.io/tenant': 'datahub',
```
squat: same with the other querier service: these pods will also be added to the endpoints

PR #47, opened 2019-08-28
Update Telemeter and Thanos images

metalmatze:
Update images to more recent versions.


squat:
:/ I think this will break staging because we havent yet rolled out the updated secrets to match the fields expected by the new telemeter server deployment config

PR #49, opened 2019-08-28
Set proper resources requests and limits for Thanos components

metalmatze:
We want to set proper resource requests and limits for Thanos components so we limit the damage we can make to the entire cluster.
Values used are coming from the upstream kube-thanos project and will be overridden in the downstream deployment specific to the cluster.


environments/openshift/kube-thanos.libsonnet
```diff
@@ -234,6 +244,18 @@ local list = import 'telemeter/lib/list.libsonnet';
...
           value: 'telemeter-thanos-stage-s3',
         },
+        { name: 'THANOS_QUERIER_CPU_REQUEST', value: '100m' },
+        { name: 'THANOS_QUERIER_CPU_LIMIT', value: '"1"' },
+        { name: 'THANOS_QUERIER_MEMORY_REQUEST', value: '500Mi' },
+        { name: 'THANOS_QUERIER_MEMORY_LIMIT', value: '1Gi' },
+        { name: 'THANOS_STORE_CPU_REQUEST', value: '100m' },
+        { name: 'THANOS_STORE_CPU_LIMIT', value: '"1"' },
+        { name: 'THANOS_STORE_MEMORY_REQUEST', value: '500Mi' },
+        { name: 'THANOS_STORE_MEMORY_LIMIT', value: '1Gi' },
```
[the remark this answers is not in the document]
brancz: very likely yes

metalmatze:
FYI Before we merge this PR, we should make sure to set the proper limits in our cluster.

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -35,6 +35,13 @@ objects:
             name: grpc
           - containerPort: 9090
             name: http
+          resources:
```
aditya-konarde: Looks like the current configuration hardcoded CPU requests and limits to the querier container, and parameterized ones to the proxy in front of the querier.

metalmatze: Good catch. Updating.

PR #50, opened 2019-08-30
Add a PVC to Thanos Store

metalmatze:
We want to save locally on a PVC the data that the Thanos Store gets from object storage.

/cc @kakkoyun @squat @aditya-konarde @brancz

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -429,9 +429,17 @@ objects:
           - mountPath: /var/thanos/store
             name: data
             readOnly: false
-        volumes:
-        - emptyDir: {}
-          name: data
+        volumes: []
+    volumeClaimTemplates:
+    - metadata:
+        name: data
+      spec:
+        accessModes:
+        - ReadWriteOnce
+        resources:
+          requests:
+            storage: 50Gi
+        storageClassName: gp2-encrypted
 - apiVersion: v1
   data:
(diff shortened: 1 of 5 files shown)


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -429,9 +429,17 @@ objects:
           - mountPath: /var/thanos/store
             name: data
             readOnly: false
-        volumes:
-        - emptyDir: {}
-          name: data
+        volumes: []
+    volumeClaimTemplates:
+    - metadata:
+        name: data
```
aditya-konarde: Nit: Can we name this something more explicit than 'data'? Perhaps 'observatorium-thanos-store-data'

brancz: it's already in an observatorium namespace, I think that would be duplicate, but `<statefulsetname>-data` would make sense I think

aditya-konarde: Yeah `<statefulsetname>-data` sounds perfect. Just want to avoid the possibility of multiple components calling their PVC `data`

metalmatze: Made the change upstream. Will bump here once merged upstream and it'll update without any other changes. :+1: 

https://github.com/metalmatze/kube-thanos/pull/33

brancz:
lgtm, merge whenever you're ready

PR #51, opened 2019-09-02
Add alerts and dashboards for Thanos receive controller

kakkoyun:
This PR;
* adds new alerts and dashboards for Thanos Receive Controller,
* updates error panels for Thanos dashboards,
* improves existing alerts

cc @metalmatze @squat @aditya-konarde

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,1 @@
 vendor/
-*.servicemonitor
-*.configmap
-*.prometheusrules
-*-serviceAccount
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 vendor/
+*.servicemonitor
+*.configmap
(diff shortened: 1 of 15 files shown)


.gitignore
```diff
@@ -1 +1,5 @@
 vendor/
+*.servicemonitor
+*.configmap
+*.prometheusrules
+*-serviceAccount
```
metalmatze: What are these files? I don't think we generate them anyway, so that we need to ignore them, do we?

kakkoyun: Maybe it was wrong to put them here. So the thing is, they're generated interim files in our build process, sometimes `build.sh` doesn't remove them in my local setup, I just wanted to put them here to prevent accidental commits. Better to remove them.

PR #53, opened 2019-09-05
Add Thanos compactor to Kubernetes and OpenShift

metalmatze:
We want to add the Thanos compactor to our stack to compact all the blocks that we already have in object storage as well as future ones.

/cc @squat @kakkoyun @bwplotka @brancz

### DIFF
--- a/environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
+++ b/environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
@@ -4,7 +4,7 @@ metadata:
   labels:
     app.kubernetes.io/name: thanos-compactor
   name: thanos-compactor
-  namespace: monitoring
+  namespace: observatorium
 spec:
   replicas: 1
   selector:
@@ -21,8 +21,8 @@ spec:
         - compact
         - --wait
         - --retention.resolution-raw=16d
-        - --retention.resolution-5m=42d
-        - --retention.resolution-1h=180d
+        - --retention.resolution-5m=16d
+        - --retention.resolution-1h=16d
         - --objstore.config=$(OBJSTORE_CONFIG)
         env:
         - name: OBJSTORE_CONFIG
--- a/environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
+++ b/environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
@@ -1,0 +1,45 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    app.kubernetes.io/name: thanos-compactor
+  name: thanos-compactor
+  namespace: monitoring
+spec:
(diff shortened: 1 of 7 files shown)


environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
```diff
@@ -0,0 +1,45 @@
...
+        - --retention.resolution-5m=42d
+        - --retention.resolution-1h=180d
+        - --objstore.config=$(OBJSTORE_CONFIG)
+        env:
+        - name: OBJSTORE_CONFIG
+          valueFrom:
+            secretKeyRef:
+              key: thanos.yaml
+              name: thanos-objectstorage
+        image: quay.io/thanos/thanos:v0.6.1
```
bwplotka: v0.7.0?

environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
```diff
@@ -0,0 +1,45 @@
...
+        name: thanos-compactor
+        ports:
+        - containerPort: 10902
+          name: http
+        resources:
+          limits:
+            cpu: 500m
+            memory: 2Gi
```
bwplotka: so here is 2GI but the jsonnet specified 5GB :thinking: 

>   { name: 'THANOS_COMPACTOR_MEMORY_LIMIT', value: '5Gi' },

environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
metalmatze: Yes. We've moved to v0.7.0 on upstream kube-thanos but still override the image in this repo: https://github.com/observatorium/configuration/blob/master/environments/kubernetes/kube-thanos.libsonnet#L19

I guess it's safe to move to v0.7.0 however that should be another PR, imho.

environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
metalmatze: That's because this file `environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml` is a generic combined Kubernetes deployment for Observatorium, where's the `{ name: 'THANOS_COMPACTOR_MEMORY_LIMIT', value: '5Gi' },` is for OpenShift and more closely to what we actually want on our deployments already. Although we still override these values on a per cluster basis later.

bwplotka: Thanks!

environments/kubernetes/manifests/thanos-compactor-statefulSet.yaml
bwplotka: :+1:

PR #58, opened 2019-09-09
Add auth proxy for cache, change cache container port

aditya-konarde:
Changes: 
- Add an auth proxy container for the cache container
- Change the cache port to `9090` to be consistent with the querier. 

Related: PR for adding a route redirect reference on the SA used for the auth proxy is already created here: https://github.com/openshift/telemeter/pull/233

### DIFF
--- a/environments/openshift/kube-thanos.libsonnet
+++ b/environments/openshift/kube-thanos.libsonnet
@@ -272,6 +272,13 @@ local list = import 'telemeter/lib/list.libsonnet';
       roleBinding+: setSubjectNamespace(super.roleBinding) + roleBinding.mixin.metadata.withNamespace(namespace),
     },
     querierCache+: {
+      // The proxy secret is there to encrypt session created by the oauth proxy.
+      proxySecret:
+        secret.new('querier-proxy', {
+          session_secret: std.base64($.thanos.variables.proxyConfig.sessionSecret),
+        }) +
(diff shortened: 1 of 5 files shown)


brancz:
I'm curious .. why not use sso.redhat.com ?

aditya-konarde:
@brancz I'm just trying to stay consistent with what we currently have for the scope of the PR. 

Changing the authentication mode from namespace RBAC to sso.redhat.com needs to be service-wide and affects the querier, prometheus-telemeter and the cache. We'll also need to create app-interface integrations related to that. 

Does that answer your question? I don't think planning a move to sso.redhat.com is in the scope of this PR.

brancz:
That's fine, I more meant, if we're adding auth, then why not directly do what we'll end up doing anyways? Not blocking this at all.

environments/openshift/kube-thanos.libsonnet
```diff
@@ -272,6 +272,13 @@ local list = import 'telemeter/lib/list.libsonnet';
       roleBinding+: setSubjectNamespace(super.roleBinding) + roleBinding.mixin.metadata.withNamespace(namespace),
     },
     querierCache+: {
+      // The proxy secret is there to encrypt session created by the oauth proxy.
+      proxySecret:
+        secret.new('querier-proxy', {
```
aditya-konarde: This needs to be called `querier-cache-proxy`

PR #64, opened 2019-09-10
Added jaeger all in one with badger storage

jpkrohling:
This is a first draft to get Jaeger all in one in Observatorium. This brings a simple Jaeger setup using the `all-in-one` strategy and uses the badger storage type.

This does *not* reflect a production installation as the operator would do. Namely, I removed all the ports that might not be necessary by Thanos. I assume that it will use a Jaeger Agent as sidecar, and this sidecar will then use gRPC to talk to the collectors.

The headless service is to enable the gRPC client side load balancing to work. No services with IP are provided, as I don't think Istio (or other types of ingresses) will be necessary, but I can certainly add an IP-service if needed.

I added a `Makefile` and added a couple of lines to the readme, based on my needs as someone trying to understand how it all works. Let me know if those aren't appropriate and I'll revert those changes.

cc @aditya-konarde for the main parts, and cc @burmanm for the badger-specific settings.

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/environments/kubernetes/main.jsonnet
+++ b/environments/kubernetes/main.jsonnet
@@ -1,6 +1,7 @@
 local app =
   (import 'kube-thanos.libsonnet') +
-  (import 'telemeter.libsonnet');
+  (import 'telemeter.libsonnet') +
+  (import 'jaeger.libsonnet');
 
 { ['thanos-querier-' + name]: app.thanos.querier[name] for name in std.objectFields(app.thanos.querier) } +
 { ['thanos-receive-' + name]: app.thanos.receive[name] for name in std.objectFields(app.thanos.receive) } +
@@ -8,4 +9,5 @@ local app =
 { ['thanos-store-' + name]: app.thanos.store[name] for name in std.objectFields(app.thanos.store) } +
 { ['thanos-receive-controller-' + name]: app.thanos.receiveController[name] for name in std.objectFields(app.thanos.receiveController) } +
 { ['thanos-querier-cache-' + name]: app.thanos.querierCache[name] for name in std.objectFields(app.thanos.querierCache) } +
-{ ['telemeter-' + name]: app.telemeterServer[name] for name in std.objectFields(app.telemeterServer) }
+{ ['telemeter-' + name]: app.telemeterServer[name] for name in std.objectFields(app.telemeterServer) } +
+{ ['jaeger-' + name]: app.jaeger[name] for name in std.objectFields(app.jaeger) }
--- a/environments/kubernetes/manifests/jaeger-deployment.yaml
+++ b/environments/kubernetes/manifests/jaeger-deployment.yaml
@@ -20,10 +20,17 @@ spec:
         - --badger.directory-key=/var/jaeger/store/keys
         - --badger.directory-value=/var/jaeger/store/values
         - --badger.ephemeral=false
+        - --collector.queue-size=4000
         env:
         - name: SPAN_STORAGE_TYPE
           value: badger
         image: jaegertracing/all-in-one:1.14.0
+        livenessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /
+            port: "14269"
+            scheme: HTTP
         name: jaeger-all-in-one
         ports:
         - containerPort: 14250
@@ -32,6 +39,19 @@ spec:
           name: admin-http
         - containerPort: 16686
           name: query
+        readinessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /
+            port: "14269"
+            scheme: HTTP
+        resources:
+          limits:
+            cpu: "4"
+            memory: 2Gi
+          requests:
+            cpu: "1"
+            memory: 256Mi
         volumeMounts:
         - mountPath: /var/jaeger/store
           name: jaeger-store-data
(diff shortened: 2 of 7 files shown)


jpkrohling:
Is the build supposed to work? I can't reproduce the failure locally, even when using the same build image (`quay.io/coreos/jsonnet-ci`).

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	rm -rf environments/kubernetes/manifests
+	rm -rf environments/openshift/manifests
+	rm -rf environments/sre/servicemonitors
+	rm -rf environments/sre/prometheusrules
+	rm -rf environments/sre/grafana
+
+.PHONY: build
+build:
+	@echo Building...
+	@./build.sh > build.log 2>&1
```
metalmatze: If we really want a Makefile, we should get rid of the build.sh entirely.

Makefile
```diff
@@ -0,0 +1,30 @@
+.DEFAULT_GOAL := install
+
+.PHONY: check
+check:
+	@echo Checking...
+	@git diff -s --exit-code . || (echo "Build failed: one of more files aren't staged to be committed" && exit 1)
```
metalmatze: Could we simply show the diff and not suppress its errors?

environments/kubernetes/main.jsonnet
```diff
@@ -1,11 +1,13 @@
 local app =
   (import 'kube-thanos.libsonnet') +
-  (import 'telemeter.libsonnet');
+  (import 'telemeter.libsonnet') +
+  (import '../../components/jaeger-collector.libsonnet');
```
metalmatze: Can we create `jaeger-collector.libsonnet` where we import `../../components/jaeger-collector.libsonnet` so that we can do the overrides for Kubernetes there.
In the future the `../../components/jaeger-collector.libsonnet` should come from somewhere upstream.

environments/kubernetes/main.jsonnet
```diff
@@ -1,11 +1,13 @@
...
 
 { ['thanos-querier-' + name]: app.thanos.querier[name] for name in std.objectFields(app.thanos.querier) } +
 { ['thanos-receive-' + name]: app.thanos.receive[name] for name in std.objectFields(app.thanos.receive) } +
 { ['thanos-compactor-' + name]: app.thanos.compactor[name] for name in std.objectFields(app.thanos.compactor) } +
 { ['thanos-store-' + name]: app.thanos.store[name] for name in std.objectFields(app.thanos.store) } +
 { ['thanos-receive-controller-' + name]: app.thanos.receiveController[name] for name in std.objectFields(app.thanos.receiveController) } +
 { ['thanos-querier-cache-' + name]: app.thanos.querierCache[name] for name in std.objectFields(app.thanos.querierCache) } +
-{ ['telemeter-' + name]: app.telemeterServer[name] for name in std.objectFields(app.telemeterServer) }
+{ ['telemeter-' + name]: app.telemeterServer[name] for name in std.objectFields(app.telemeterServer) } +
+{ ['jaeger-' + name]: app.jaeger[name] for name in std.objectFields(app.jaeger) }
```
metalmatze: Missing newline.

environments/kubernetes/manifests/jaeger-deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+    matchLabels:
+      app.kubernetes.io/name: jaeger-all-in-one
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: jaeger-all-in-one
+    spec:
+      containers:
+      - args:
+        - --badger.directory-key=/var/jaeger/store/keys
```
burmanm: ``--badger.ephemeral=false`` is required, otherwise these parameters are ignored. What sort of TTL has been discussed? Default is 3 days.

metalmatze: 3 days should be more than necessary. I would even say 24h is totally fine. Happy to try 3d first though.

Makefile
jpkrohling: In general, I personally prefer to wrap more complex tasks in shell scripts, as there are things that are insane to accomplish in Makefiles. I think I'll end up removing the `Makefile` for this PR, as it's really not within the scope of the change.

environments/kubernetes/main.jsonnet
jpkrohling: Thanks for the clarification! I've noticed the pattern, but wasn't sure what's the purpose.

environments/kubernetes/manifests/jaeger-deployment.yaml
jpkrohling: Cool, I thought the flag would be adjusted to `false` when the key/values options were set.

Makefile
jpkrohling: The diff would be huge, as it will only happen when the generated manifests aren't updated. We have a similar target in the Jaeger Operator, and when it fails, we know it's because the generated files aren't up to date.

How about we start with this silent approach, and change it to a verbose mode if we need?

environments/kubernetes/main.jsonnet
jpkrohling: How does it look now?

Makefile
metalmatze: We have had these cases and I like seeing the diff on CI, as it allows me to pin point what's actually wrong on CI.
It should be too big normally: https://cloud.drone.io/observatorium/configuration/135/1/2

jpkrohling: I'll move this to another PR. For now, I just removed it from here.

PR #66, opened 2019-09-11
Fix Dashboards job selector issues

kakkoyun:
cc @squat @metalmatze


kakkoyun:
@metalmatze This is actually a forgotten configuration change, I have tested boards using this same exact configuration. Since Observatorium can run multiple replicas for most of the components, we needed a way to select all available jobs. This is already valid for `receive` component. I've just reflected the same idea to all other components. If this causes any issues, we can use an exact match operator for components other than `receive`.

PR #67, opened 2019-09-12
Added Makefile

jpkrohling:
Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,4 +39,4 @@ install-tools:
 all: check vendor build
 
 .PHONY: ci
-ci: all
+ci: all check # we run the check twice, as the manifests might have changed as part of the build
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,42 @@
+JSONNET_FILES=$(shell find . -type f -not -path './vendor/*' \( -name '*.libsonnet' -o -name '*.jsonnet' \))
+
+.DEFAULT_GOAL := all
+
+.PHONY: check
+check:
+	@echo Checking...
+	@echo $(JSONNET_FILES) | xargs -L1 jsonnetfmt -i
+	@git diff -s --exit-code . || (echo "Build failed: one of more files aren't staged to be committed." && git diff && exit 1)
+
+.PHONY: clean
+clean:
+	@echo Cleaning...
+	@rm -rf environments/kubernetes/manifests
+	@rm -rf environments/openshift/manifests
+	@rm -rf environments/sre/servicemonitors
+	@rm -rf environments/sre/prometheusrules
+	@rm -rf environments/sre/grafana
+
+.PHONY: build
+build:
+	@echo Building...
+	@./build.sh
+
+.PHONY: vendor
+vendor:
+	@echo Bringing dependencies...
+	@jb install
+
+.PHONY: install-tools
+install-tools:
+	@## this target is missing the installation of jsonnetfmt, which doesn't seem to have a binary available (!!)
+	@go get \
+		github.com/google/go-jsonnet/cmd/jsonnet \
+		github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb \
+		github.com/brancz/gojsontoyaml
+
+.PHONY: all
+all: check vendor build
+
+.PHONY: ci
+ci: all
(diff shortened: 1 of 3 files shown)


LiliC:
Yes please! :) This was very helpful, when I looked how to contribute 👍 

cc @metalmatze @kakkoyun

kakkoyun:
lgtm, @metalmatze should take a look at it, since he has already started a discussion around it at #64.

Makefile
```diff
@@ -0,0 +1,42 @@
+JSONNET_FILES=$(shell find . -type f -not -path './vendor/*' \( -name '*.libsonnet' -o -name '*.jsonnet' \))
+
+.DEFAULT_GOAL := all
+
+.PHONY: check
+check:
+	@echo Checking...
+	@echo $(JSONNET_FILES) | xargs -L1 jsonnetfmt -i
+	@git diff -s --exit-code . || (echo "Build failed: one of more files aren't staged to be committed." && git diff && exit 1)
```
metalmatze: As commented in an earlier PR, can we not suppress these diffs, I'd like to see what's going on.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+	@rm -rf environments/kubernetes/manifests
+	@rm -rf environments/openshift/manifests
+	@rm -rf environments/sre/servicemonitors
+	@rm -rf environments/sre/prometheusrules
+	@rm -rf environments/sre/grafana
+
+.PHONY: build
+build:
+	@echo Building...
+	@./build.sh
```
metalmatze: If we want this Makefile we should replace the `build.sh` with actual make targets like:

```Makefile
environments/kubernetes/manifests:
jsonnet -J vendor -m environments/kubernetes/manifests environments/kubernetes/main.jsonnet | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml; rm -f {}' -- {}

# for the rest of the build.sh the same
...
```

Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+.PHONY: vendor
+vendor:
+	@echo Bringing dependencies...
+	@jb install
```
metalmatze: We currently don't suppress the output of jb anywhere.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+	@go get \
+		github.com/google/go-jsonnet/cmd/jsonnet \
+		github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb \
+		github.com/brancz/gojsontoyaml
+
+.PHONY: all
+all: check vendor build
+
+.PHONY: ci
+ci: all check # we run the check twice, as the manifests might have changed as part of the build
```
metalmatze: We usually check at the end of running everything. I don't really get why we would check at the beginning? :thinking:

Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+.PHONY: vendor
+vendor:
+	@echo Bringing dependencies...
+	@jb install
+
+.PHONY: install-tools
+install-tools:
+	@## this target is missing the installation of jsonnetfmt, which doesn't seem to have a binary available (!!)
+	@go get \
```
metalmatze: We could also check of the binaries being available and always run this target when simply running `make`.

Makefile
jpkrohling: It is doing that, but only if it fails. Look at the `&& git diff && exit 1` at the end of the command.

Makefile
jpkrohling: Fail fast :)

Makefile
jpkrohling: I'd rather keep the Makefile as a wrapper for more complex scripts. When we have simple commands, like `jb install`, it's OK to have it here, but once it gets more complex, it's easier to maintain shell scripts than make targets.

kakkoyun:
@metalmatze @jpkrohling Shall we push this through? I could help with the fixes if you don't have time.

jpkrohling:
From my side: yes, please!

jpkrohling:
Closing, as I don't think this will be merged anytime soon. Not in this form, at least :-)

PR #72, opened 2019-09-24
Use CircleCI and run KinD clusters there for e2e tests

metalmatze:
/cc @brancz @wmuha @ant31

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,28 @@
+jobs:
+  build:
+    working_directory: /go/src/github.com/observatorium/configuration
+    docker:
+      - image: quay.io/coreos/jsonnet-ci
+    steps:
+      - checkout
+      - run: find . -type f -not -path './vendor/*' \( -name '*.libsonnet' -o -name '*.jsonnet' \) | xargs -L1 jsonnetfmt --test
+      - run: curl -o ./jb -L "https://github.com/jsonnet-bundler/jsonnet-bundler/releases/download/v0.2.0/jb-$(go env GOOS)-$(go env GOARCH)" && chmod +x ./jb
+      - run: ./jb install
+      - run: ./build.sh
+      - run: git diff --exit-code
+
+  tests:
+    machine: true
+    steps:
+      - checkout
+      - run: |
+          ./tests/kind.sh
+          ./tests/deploy.sh
+          ./tests/test.sh
+
+workflows:
+  version: 2
+  build_and_test:
+    jobs:
+      - build
+      - tests
--- a/environments/testing/secret.yaml
+++ b/environments/testing/secret.yaml
@@ -1,0 +1,16 @@
+# type: s3
+# config:
+#   bucket: thanos
+#   endpoint: minio.minio.svc.cluster.local:9000
+#   insecure: true
+#   access_key: minio
+#   secret_key: minio123
+
+apiVersion: v1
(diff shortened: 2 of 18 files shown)


metalmatze:
@ant31, can you enable CircleCI for this repo? Thanks :)

brancz:
Let us know when you think this is ready for a review :)

environments/testing/secret.yaml
```diff
@@ -0,0 +1,16 @@
...
+#   access_key: minio
+#   secret_key: minio123
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: thanos-objectstorage
+  namespace: observatorium
+type: Opaque
+data:
```
brancz: Any reason why not to make use of [stringData](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually) here?

.circleci/config.yml
```diff
@@ -0,0 +1,28 @@
+jobs:
+  build:
+    working_directory: /go/src/github.com/observatorium/configuration
+    docker:
+      - image: quay.io/coreos/jsonnet-ci
+    steps:
+      - checkout
+      - run: find . -type f -not -path './vendor/*' \( -name '*.libsonnet' -o -name '*.jsonnet' \) | xargs -L1 jsonnetfmt --test
+      - run: curl -o ./jb -L "https://github.com/jsonnet-bundler/jsonnet-bundler/releases/download/v0.2.0/jb-$(go env GOOS)-$(go env GOARCH)" && chmod +x ./jb
```
squat: This works for now :+1:  we should probably introduce a Makefile soon to encapsulate all of this logic and have it less spread out

environments/testing/secret.yaml
squat: yeah +1 no need to have this bas64 encoded in this test file

tests/deploy.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
squat: All three of these scripts are quite tightly coupled (binary locations, logic, actual things deployed, etc), for easier navigability and maintainability I would suggest to merge them into one single scripts called tests/e2e.sh

environments/testing/secret.yaml
brancz: or ever for that fact if it's really just text data :man_shrugging:

metalmatze: Will update that one now, yep :+1:

tests/deploy.sh
metalmatze: The idea right now is, that one can easily run `deploy.sh` once, if they already have a kind cluster running. Then run the tests multiple times, without running all of it from the beginning to the end.

squat: Ahh interesting. I see how that would be useful. Hmm right now the scripts are so tightly coupled that it seems a bit disingenuous to put them in separate scripts. It would be great for them to be separate functions in the same script. We can definitely merge as is; it’s not a blocker for me. I’ll can make a followup PR to show what I mean

brancz:
Let's move ahead and adapt in follow ups :)

tests/deploy.sh
metalmatze: Let's do it in a follow up and discuss how we might be able to have one script and maybe flags to run individual functions only.

PR #73, opened 2019-09-25
Add dedicated Prometheus for Subscription Labels

kakkoyun:
This PR adds new Prometheus for Observatorium.

### DIFF
--- a/environments/openshift/main.jsonnet
+++ b/environments/openshift/main.jsonnet
@@ -15,14 +15,14 @@ local app =
         objects:
           $.thanos.template.objects +
           $.telemeterServer.list.objects +
-          $.prometheus.template.objects,
+          $.prometheusAms.template.objects,
 
         parameters:
           $.thanos.template.parameters +
           $.telemeterServer.list.parameters + [
             { name: 'TELEMETER_FORWARD_URL', value: '' },
           ] +
-          $.prometheus.template.parameters,
+          $.prometheusAms.template.parameters,
       },
   } + {
     template+: {
--- a/environments/openshift/main.jsonnet
+++ b/environments/openshift/main.jsonnet
@@ -5,6 +5,7 @@ local list = import 'telemeter/lib/list.libsonnet';
 
 local app =
   (import 'kube-thanos.libsonnet') +
+  (import 'prometheus.libsonnet') +
   (import 'telemeter.libsonnet') +
   {
     local thanos = super.thanos,
@@ -13,13 +14,15 @@ local app =
       list.asList('observatorium', {}, []) + {
         objects:
           $.thanos.template.objects +
-          $.telemeterServer.list.objects,
+          $.telemeterServer.list.objects +
+          $.prometheus.template.objects,
 
         parameters:
           $.thanos.template.parameters +
           $.telemeterServer.list.parameters + [
             { name: 'TELEMETER_FORWARD_URL', value: '' },
-          ],
+          ] +
+          $.prometheus.template.parameters,
       },
   } + {
     template+: {
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -914,12 +914,6 @@ objects:
   metadata:
     name: prometheus-ams-${NAMESPACE}
   rules:
-  - apiGroups:
-    - ""
-    resources:
-    - nodes/metrics
-    verbs:
-    - get
   - nonResourceURLs:
     - /metrics
     verbs:
@@ -936,6 +930,46 @@ objects:
   - kind: ServiceAccount
     name: prometheus-ams
     namespace: ${NAMESPACE}
+- apiVersion: v1
+  data:
+    nginx.conf: |
+      daemon off;
+      worker_processes 1;
+      error_log /dev/stderr;
+      pid /tmp/nginx.pid;
+
+      events {
+          worker_connections 1024;
+      }
+
+      http {
+        log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                          '$status $body_bytes_sent "$http_referer" '
+                          '"$http_user_agent" "$http_x_forwarded_for"';
+
+        upstream app {
+          server http://thanos-receive.${NAMESPACE}.svc.cluster.local:10902/api/v1/receive;
+          keepalive 64;
+        }
+
+        server {
+          listen *:8080;
+          server_name _;
+          access_log  /dev/stdout  main;
+          error_log /dev/stderr;
+
+          location / {
+            proxy_set_header THANOS-TENANT ${THANOS_TENANT};
+            proxy_pass http://app;
+          }
+        }
+      }
+  kind: ConfigMap
+  metadata:
+    labels:
+      prometheus: ams
+    name: remote-write-proxy-config
+    namespace: ${NAMESPACE}
 - apiVersion: monitoring.coreos.com/v1
   kind: Prometheus
   metadata:
@@ -949,15 +983,15 @@ objects:
     baseImage: ${PROMETHEUS_IMAGE}
     containers:
     - args: []
-      image: ${PROXY_IMAGE}:${PROXY_IMAGE_TAG}
-      name: proxy
+      image: quay.io/app-sre/observatorium-receive-proxy:latest
+      name: remote-write-proxy
       ports:
       - containerPort: 8080
-        name: https
+        name: http
       volumeMounts:
       - mountPath: /nginx.conf
         name: prometheus-remote-write-proxy-config
-        readOnly: false
+        readOnly: true
     nodeSelector:
       kubernetes.io/os: linux
     podMonitorSelector: {}
@@ -992,7 +1026,7 @@ objects:
     - configMap:
         items:
         - nginx.conf
-        name: nginx-config
+        name: remote-write-proxy-config
       name: prometheus-remote-write-proxy-config
 - apiVersion: rbac.authorization.k8s.io/v1
   kind: RoleBinding
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -909,27 +909,6 @@ objects:
         - name: telemeter-server-tls
           secret:
             secretName: telemeter-server-shared
-- apiVersion: rbac.authorization.k8s.io/v1
-  kind: ClusterRole
-  metadata:
-    name: prometheus-ams-${NAMESPACE}
-  rules:
-  - nonResourceURLs:
-    - /metrics
-    verbs:
-    - get
-- apiVersion: rbac.authorization.k8s.io/v1
-  kind: ClusterRoleBinding
-  metadata:
-    name: prometheus-ams-${NAMESPACE}
-  roleRef:
-    apiGroup: rbac.authorization.k8s.io
-    kind: ClusterRole
-    name: prometheus-ams-${NAMESPACE}
-  subjects:
-  - kind: ServiceAccount
-    name: prometheus-ams
-    namespace: ${NAMESPACE}
 - apiVersion: v1
   data:
     nginx.conf: |
@@ -1013,7 +992,8 @@ objects:
       fsGroup: 2000
       runAsNonRoot: true
       runAsUser: 1000
-    serviceAccountName: prometheus-ams
+    serviceAccount: prometheus-telemeter
+    serviceAccountName: prometheus-telemeter
     serviceMonitorNamespaceSelector: {}
     serviceMonitorSelector: {}
     version: ${PROMETHEUS_AMS_IMAGE_TAG}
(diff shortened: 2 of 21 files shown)


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +898,220 @@ objects:
...
+          worker_connections 1024;
+      }
+
+      http {
+        log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                          '$status $body_bytes_sent "$http_referer" '
+                          '"$http_user_agent" "$http_x_forwarded_for"';
+
+        upstream app {
+          server http://thanos-receive.${NAMESPACE}.svc.cluster.local:10902/api/v1/receive;
```
kakkoyun: I anticipated OpenShift templating engine would change this as well.

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +898,220 @@ objects:
...
+        }
+
+        server {
+          listen *:8080;
+          server_name _;
+          access_log  /dev/stdout  main;
+          error_log /dev/stderr;
+
+          location / {
+            proxy_set_header THANOS-TENANT ${THANOS_TENANT};
```
kakkoyun: ditto

environments/openshift/main.jsonnet
```diff
@@ -13,13 +14,15 @@ local app =
       list.asList('observatorium', {}, []) + {
         objects:
           $.thanos.template.objects +
-          $.telemeterServer.list.objects,
+          $.telemeterServer.list.objects +
+          $.prometheus.template.objects,
```
metalmatze: We should name this something way more specific to describe what that Prometheus is for.

kakkoyun: `amsPrometheus`? `subscriptionPrometheus`? `subcriptionSyncer` ? 
What would be the most descriptive?

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +898,220 @@ objects:
...
+    name: prometheus-ams-${NAMESPACE}
+  subjects:
+  - kind: ServiceAccount
+    name: prometheus-ams
+    namespace: ${NAMESPACE}
+- apiVersion: v1
+  data:
+    nginx.conf: |
+      daemon off;
+      worker_processes 1;
```
metalmatze: Why only 1? (I have no background around this, just asking)

environments/openshift/manifests/observatorium-template.yaml
metalmatze: That is correct.

environments/openshift/manifests/observatorium-template.yaml
metalmatze: I think we should hardcode this in here. We are probably going to hardcode the same ID in the Telemeter v1 forwarding.

environments/openshift/manifests/observatorium-template.yaml
kakkoyun: Because it runs in a container, it spawns a single process which runs on foreground, with 1024 workers. If it fails then container exits.

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +898,220 @@ objects:
...
+- apiVersion: monitoring.coreos.com/v1
+  kind: Prometheus
+  metadata:
+    labels:
+      prometheus: ams
+    name: ams
+    namespace: ${NAMESPACE}
+  spec:
+    alerting:
+      alertmanagers: []
```
metalmatze: Thinking about this: We should probably even alert on the scrape target being down and send it to app-sre alertmanager.

environments/openshift/manifests/observatorium-template.yaml
kakkoyun: Could we share this as a jsonnet variable between them to avoid synchronization issue?

environments/openshift/manifests/observatorium-template.yaml
kakkoyun: I have checked this again, `nginx` spawns 1 master process and `n` worker processes, best practice is to spawn 1 worker process per core, so what should it be?

environments/openshift/manifests/observatorium-template.yaml
squat: There is no variable replacement for `${THANOS_TENANT}`, this is not defined as a parameter anywhere. It also doesn't really make sense for it to be a parameter: this configuration stack serves every possible thanos tenant. Really, we need to launch one remote-write-proxy per tenant, in which case the header should be be hardcoded for that tenant.

kakkoyun: I'll use `_id` then as we have it right now.

brancz: this would need to be an id not `_id` right?

kakkoyun: Yes, as I recall that's the default. That's why I suggested that.

environments/openshift/manifests/observatorium-template.yaml
kakkoyun: I discussed this with @aditya-konarde, we'll add it as a next step.

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
         - name: telemeter-server-tls
           secret:
             secretName: telemeter-server-shared
+- apiVersion: rbac.authorization.k8s.io/v1
```
aditya-konarde: Due to a known limitation, we cannot create clusterroles for the rbac API, so I'd suggest you use the same one as infogw-proxy is using, for now.

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
...
+  metadata:
+    labels:
+      prometheus: ams
+    name: ams
+    namespace: ${NAMESPACE}
+  spec:
+    baseImage: ${PROMETHEUS_IMAGE}
+    containers:
+    - args: []
+      image: quay.io/app-sre/observatorium-receive-proxy:latest
```
aditya-konarde: Can we parameterize the image tag instead of using latest? Its okay to start with a hardcoded version, works better than `latest`

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
...
+      name: remote-write-proxy
+      ports:
+      - containerPort: 8080
+        name: http
+      volumeMounts:
+      - mountPath: /nginx.conf
+        name: prometheus-remote-write-proxy-config
+        readOnly: true
+    nodeSelector:
```
aditya-konarde: This is not needed, all of our nodes are Linux :)

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
...
+      kubernetes.io/os: linux
+    podMonitorSelector: {}
```
aditya-konarde: Check if this was supported in Prometheus operator  0.22.2

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
...
+    - url: http://localhost:8080
+      write_relabel_configs:
+        action: keep
+        regex: subscription_labels
+        source_labels:
+        - __name__
+    replicas: 1
+    resources:
+      limits:
+        cpu: ${PROMETHEUS_CPU_LIMIT}
```
aditya-konarde: We might need to be more specific on the parameters here, we don't want to accidentally give it the huge resource requirements that the telemeter-prometheus has

environments/sre/servicemonitors.jsonnet
```diff
@@ -94,6 +100,10 @@ local sm =
     metadata+: { name+: '-stage' },
     spec+: { namespaceSelector+: { matchNames: ['telemeter-stage'] } },
   },
+  'observatorium-prometheus-ams-stage.servicemonitor': prom.prometheusAms.serviceMonitor {
+    metadata+: { name+: '-stage', labels: { prometheus: 'app-sre' } },
+    spec+: { namespaceSelector+: { matchNames: ['telemeter-production'] } },
```
squat: should this be `telemeter-stage`?

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
...
+  - kind: ServiceAccount
+    name: prometheus-ams
+    namespace: ${NAMESPACE}
+- apiVersion: rbac.authorization.k8s.io/v1
+  items:
+  - apiVersion: rbac.authorization.k8s.io/v1
+    kind: RoleBinding
+    metadata:
+      name: prometheus-ams
+      namespace: observatorium
```
squat: these namespaces need to be parameterized, right?

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
...
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: Role
+      name: prometheus-ams
+    subjects:
+    - kind: ServiceAccount
+      name: prometheus-ams
+      namespace: observatorium
```
squat: same^

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -878,6 +878,218 @@ objects:
...
+    - configmaps
+    verbs:
+    - get
+- apiVersion: rbac.authorization.k8s.io/v1
+  items:
+  - apiVersion: rbac.authorization.k8s.io/v1
+    kind: Role
+    metadata:
+      name: prometheus-ams
+      namespace: observatorium
```
squat: same^

environments/openshift/prometheus.libsonnet
```diff
@@ -0,0 +1,271 @@
...
+            volume.fromConfigMap('prometheus-%s' % $.prometheusAms.configmap.metadata.name, $.prometheusAms.configmap.metadata.name, ['nginx.conf']),
+          ],
+        },
+      },
+    serviceMonitor:
+      {
+        apiVersion: 'monitoring.coreos.com/v1',
+        kind: 'ServiceMonitor',
+        metadata: {
+          name: 'prometheus-ams',
```
squat: this looks like the only field in jsonnet that is explicitly assigned the string `'prometheus-ams'`, instead of `'prometheus' + $._config.ams.prometheus.name`

environments/openshift/manifests/observatorium-template.yaml
metalmatze: It's not. But it should be ignored. Better remove it anyway.

environments/openshift/manifests/observatorium-template.yaml
metalmatze: I'm currently thinking, do we even need those at all? We only have that component once and it's a rather temporary thing too.

squat:
it looks like we need to regenerate and commit: https://cloud.drone.io/observatorium/configuration/188/1/2

kakkoyun:
@squat You're a fast!

environments/openshift/prometheus.libsonnet
```diff
@@ -0,0 +1,328 @@
...
+
+    imageRepos+:: {
+      prometheusAms: 'quay.io/prometheus/prometheus',
+      remoteWriteProxy: 'quay.io/app-sre/observatorium-receive-proxy',
+    },
+
+    ams+:: {
+      proxyPort: 8080,
+      remoteWriteTarget: 'http://%s.%s.svc.cluster.local:%d/api/v1/receive' % [
+        $.thanos.receive.service.metadata.name,
```
squat: niceeeee this is using the special thanos-receive service that selects all of the hashrings :)

One question: The thanos-receive service does not have a cluster IP, meaning that when looking up the DNS name `thanos.telemeter-production.svc.cluster.local` you will get back a list of all of the thanos-receive IPs. Do you know how NGINX behaves here with regards to DNS resolution? Does nginx round-robin these IPs? or does it always use the first? etc. 

For this integration, where the amount of traffic is quite low, it's not a big deal, but for integrations with l0ts of data, if NGINX always uses just one IP, then we could take down hashring nodes one by one.

squat: OK, looks like we are fine:, per the NGINX docs:

> A domain name that resolves to several IP addresses defines multiple servers at once.
http://nginx.org/en/docs/http/ngx_http_upstream_module.html#server

and

> By default, requests are distributed between the servers using a weighted round-robin balancing method.
http://nginx.org/en/docs/http/ngx_http_upstream_module.html#upstream

So we'll automatically load-balance across all of the IPs

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -909,6 +909,215 @@ objects:
...
+          worker_connections 1024;
+      }
+
+      http {
+        log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                          '$status $body_bytes_sent "$http_referer" '
+                          '"$http_user_agent" "$http_x_forwarded_for"';
+
+        upstream app {
+          server http://thanos-receive.${NAMESPACE}.svc.cluster.local:10902/api/v1/receive;
```
squat: There is just one more issue I see here. The pods backing this service can change over time, however NGINX only resolves the DNS name one time and caches it. If the thanos-receive pods get cycled and their IPs change, or if the hashring is re-sized, the cached IPs will be incorrect. We could tell NGINX to limit the cache duration of the DNS resolution to maybe 1 minute?

kakkoyun: Thanks for pointing this out. After a quick search, I've realized this might be a problem, I'll look into this. https://www.nginx.com/blog/dns-service-discovery-nginx-plus/#domain-name-proxy_pass

kakkoyun: I have updated the configuration as it suggested in the blog post. I haven't tested it yet.

environments/openshift/prometheus.libsonnet
```diff
@@ -0,0 +1,275 @@
...
+        },
+        spec: {
+          replicas: $._config.ams.prometheus.replicas,
+          version: '${PROMETHEUS_AMS_IMAGE_TAG}',
+          baseImage: '${PROMETHEUS_AMS_IMAGE}',
+          securityContext: {},
+          serviceAccount: 'prometheus-telemeter',
+          serviceAccountName: 'prometheus-telemeter',
+          serviceMonitorSelector: selector.withMatchLabels({
+            'k8s-app': 'prometheus-ams',
```
squat: we no longer use this label name and instead try to use `app.kubernetes.io/name` consistently

environments/openshift/prometheus.libsonnet
```diff
@@ -0,0 +1,275 @@
...
+      },
+    serviceMonitor:
+      {
+        apiVersion: 'monitoring.coreos.com/v1',
+        kind: 'ServiceMonitor',
+        metadata: {
+          name: 'prometheus-' + $._config.ams.prometheus.name,
+          namespace: $._config.namespace,
+          labels: {
+            'k8s-app': 'prometheus-ams',
```
squat: same here

environments/openshift/prometheus.libsonnet
```diff
@@ -0,0 +1,275 @@
+local k3 = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
```
squat: `k3` seems to be entirely unused in this file, can we remove it?

squat:
ok LGTM I'll squash the commits and merge :)

PR #74, opened 2019-09-26
telemeter-server: Bump to include type and encoding fix

brancz:
Bump to deploy:

* https://github.com/openshift/telemeter/pull/243
* https://github.com/openshift/telemeter/pull/242

@metalmatze @ant31 @squat @kakkoyun @aditya-konarde

### DIFF
--- a/environments/kubernetes/manifests/telemeter-statefulSet.yaml
+++ b/environments/kubernetes/manifests/telemeter-statefulSet.yaml
@@ -54,7 +54,7 @@ spec:
             secretKeyRef:
               key: client_id
               name: telemeter-server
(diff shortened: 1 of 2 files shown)


aditya-konarde:
For the OpenShift template, we override the image tags here: https://gitlab.cee.redhat.com/service/saas-telemeter/blob/master/telemeter-services/observatorium.yaml

While your PR works for the Kubernetes manifests, I think you also need to bump the other repo

squat:
force-pushing to reset broken circle-ci test

squat:
The image is already bumped in saas-telemeter, so this is really just for the generated k8s manifests, but still nice to have.

PR #75, opened 2019-09-26
Update retention to be 32d,92d,0d

metalmatze:
As discussed in previous meetings.

/cc @brancz @kakkoyun @squat

### DIFF
--- a/environments/kubernetes/kube-thanos.libsonnet
+++ b/environments/kubernetes/kube-thanos.libsonnet
@@ -58,7 +58,7 @@ local rolebinding = k.rbac.v1.roleBinding;
                     '--wait',
                     '--retention.resolution-raw=32d',
                     '--retention.resolution-5m=92d',
-                    '--retention.resolution-1h=10000d',
+                    '--retention.resolution-1h=0d',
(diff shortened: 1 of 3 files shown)


environments/kubernetes/kube-thanos.libsonnet
```diff
@@ -56,9 +56,9 @@ local rolebinding = k.rbac.v1.roleBinding;
                   args: [
                     'compact',
                     '--wait',
-                    '--retention.resolution-raw=16d',
-                    '--retention.resolution-5m=16d',
-                    '--retention.resolution-1h=16d',
+                    '--retention.resolution-raw=32d',
+                    '--retention.resolution-5m=92d',
+                    '--retention.resolution-1h=10000d',
```
brancz: I would change this to `0d`, which disables retention

brancz:
small comment, otherwise lgtm

metalmatze:
Apparently `0d` disables the retention. Are we still compacting but then keep all of those blocks forever? I'm a bit confused about that part. Thus I set it to 10000d for now :smile:

brancz:
It means it will downsample to 1h and keep samples at 1h resolution forever

PR #80, opened 2019-10-01
Fix labels issues

kakkoyun:


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -944,28 +944,25 @@ objects:
   kind: ConfigMap
   metadata:
     labels:
-      prometheus: ams
+      app.kubernetes.io/name: prometheus-remote-write-proxy
```
brancz: nit: wouldn't this application name be `prometheus` and the app instance be `prometheus-remote-write-proxy`?

kakkoyun: Thanks for pointing that out, I'll send another PR to fix labels with the help of https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

PR #82, opened 2019-10-02
Use Kubernetes recommended labels for Prometheus AMS

kakkoyun:
cc @brancz

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -943,25 +943,31 @@ objects:
   kind: ConfigMap
   metadata:
     labels:
-      app.kubernetes.io/name: prometheus-remote-write-proxy
+      app.kubernetes.io/instance: remote-write-proxy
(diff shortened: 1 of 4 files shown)


brancz:
nice :+1: lgtm

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -1016,13 +1025,15 @@ objects:
     serviceAccountName: prometheus-telemeter
     serviceMonitorSelector:
       matchLabels:
-        app.kubernetes.io/name: prometheus-ams
+        prometheus: ams
```
squat: If you merge this, we'll stop picking up the current subscription labels service monitor because it does not have this label

kakkoyun: Yes, exactly 👍  That's why I created this as a draft for now.

kakkoyun:
@squat Shall we quickly merge this so that we can test? What do you think?

PR #86, opened 2019-10-02
Disable Prometheus AMS debug logs and remove configured remote-write timeout

kakkoyun:
cc @aditya-konarde


environments/openshift/prometheus.libsonnet
```diff
@@ -40,7 +40,6 @@ local list = import 'telemeter/lib/list.libsonnet';
         remoteWrite: [
           {
             url: $._config.ams.remoteWriteProxy,
-            remoteTimeout: '5m',
```
metalmatze: You also don't need this anymore?

kakkoyun: By default we have `30s`, it should be enough. We have added this to uncover real issues behind `499`s.

PR #97, opened 2019-10-15
Remove liveness probes

kakkoyun:
This PR,

* upgrades kube-thanos to the latest version
* remove liveness probes

cc @squat @brancz @bwplotka @aditya-konarde

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -664,7 +664,6 @@ objects:
               path: /-/ready
               port: 10902
               scheme: HTTP
-            initialDelaySeconds: 120
           resources:
             limits:
               cpu: ${THANOS_STORE_CPU_LIMIT}
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -62,11 +62,6 @@ objects:
                 key: aws_secret_access_key
                 name: ${THANOS_S3_SECRET}
           image: ${THANOS_IMAGE}:${THANOS_IMAGE_TAG}
-          livenessProbe:
-            httpGet:
-              path: /-/healthy
-              port: 10902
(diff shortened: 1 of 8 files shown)


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -695,7 +675,7 @@ objects:
           - mountPath: /var/thanos/store
             name: thanos-store-data
             readOnly: false
-        volumes: []
+        volumes: null
```
kakkoyun: @metalmatze This comes from upstream, should we keep this?

metalmatze: yes!

bwplotka:
Can we increase readiness.. just in case?

kakkoyun:
@bwplotka Ahh, thanks for reminding that, I was gonna do that! Coming right up!

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -684,6 +664,7 @@ objects:
               path: /-/ready
               port: 10902
               scheme: HTTP
+            initialDelaySeconds: 120
```
kakkoyun: @bwplotka We observed that indexing takes ~90 seconds for us right now. Is this too much?  What do you think?

bwplotka: no - it's quite quick (: It can take few minutes (10-15min) easily for larger buckets

kakkoyun: Removing it for now, after offline discussions.

metalmatze:
For readiness probes, we want to simply start probing right away. It doesn't hurt. It can fail so often, but the first time it succeeds we can start serving traffic. Waiting 120s would just result in 120s min downtime for maybe no good reason.

bwplotka:
+1 @metalmatze we just need final timeout be 120s

PR #98, opened 2019-10-16
Add PVC to thanos-receive statefulsets

metalmatze:
This will add a 50Gi PVC to all thanos-receive pods.

/cc @kakkoyun @squat @brancz @aditya-konarde @bwplotka

### DIFF
--- a/environments/kubernetes/kube-thanos.libsonnet
+++ b/environments/kubernetes/kube-thanos.libsonnet
@@ -10,9 +10,10 @@ local rolebinding = k.rbac.v1.roleBinding;
 
 (import 'kube-thanos/kube-thanos-querier.libsonnet') +
 (import 'kube-thanos/kube-thanos-store.libsonnet') +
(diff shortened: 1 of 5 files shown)


environments/kubernetes/kube-thanos.libsonnet
```diff
@@ -162,7 +167,7 @@ local rolebinding = k.rbac.v1.roleBinding;
                 ],
 
                 local volume = sts.mixin.spec.template.spec.volumesType,
-                volumes+: [
+                volumes: [
```
bwplotka: so this means we override volumes, right?

kakkoyun: Yes, exactly.

metalmatze: Exactly. We now completely replace whatever the volumes is. We actually need to do this now, as `null + []` doesn't compile. :upside_down_face:

PR #103, opened 2019-10-17
openshift/jaeger: Use gp2-encrypted as storageclass

metalmatze:
/cc @aditya-konarde

### DIFF
--- a/components/jaeger-collector.libsonnet
+++ b/components/jaeger-collector.libsonnet
@@ -7,6 +7,10 @@ local service = k.core.v1.service;
     namespace:: error 'must set namespace for jaeger',
     image:: error 'must set image for jaeger',
     replicas:: 1,
+    pvc:: {
+      class: 'standard',
+      size: '50Gi',
+    },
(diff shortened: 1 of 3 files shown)


components/jaeger-collector.libsonnet
```diff
@@ -7,6 +7,10 @@ local service = k.core.v1.service;
     namespace:: error 'must set namespace for jaeger',
     image:: error 'must set image for jaeger',
     replicas:: 1,
+    pvc:: {
+      class: 'standard',
```
aditya-konarde: Do we know if `standard` is some kinda standard (sorry) storageClass that people use in K8s?

metalmatze: Might be, yes. However it's better than nothing, I believe. So we don't really have a choice, but can tell people to change it similar to the way we set the storage class in openshift.

aditya-konarde: Sounds good :+1:

PR #105, opened 2019-10-28
Add Jaeger Agent as sidecar

metalmatze:
Most reviewers are probably interested in all .yaml files. 

The store container is currently missing the `--tracing.config` flag which needs to be added before merging this PR.
However for a general review this is good to go from my side.

/cc @jpkrohling @bwplotka @brancz @kakkoyun @squat

### DIFF
--- a/components/jaeger-agent.libsonnet
+++ b/components/jaeger-agent.libsonnet
@@ -7,9 +7,9 @@ local containerEnv = container.envType;
   container::
     container.new('jaeger-agent', 'jaegertracing/jaeger-agent:1.14.0') +
     container.withArgs([
-      '--reporter.grpc.host-port=dns:///jaeger-collector-headless.${NAMESPACE}.svc:14250',
+      '--reporter.grpc.host-port=dns:///jaeger-collector-headless.$(NAMESPACE).svc:14250',
       '--reporter.type=grpc',
-      '--jaeger.tags=pod.namespace=${NAMESPACE},pod.name=${POD}',
+      '--jaeger.tags=pod.namespace=$(NAMESPACE),pod.name=$(POD)',
     ]) +
     container.withEnv([
       containerEnv.fromFieldPath('NAMESPACE', 'metadata.namespace'),
--- a/components/jaeger-agent.libsonnet
+++ b/components/jaeger-agent.libsonnet
@@ -1,0 +1,30 @@
+local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
+local deployment = k.apps.v1.deployment;
+local container = deployment.mixin.spec.template.spec.containersType;
+local containerEnv = container.envType;
+
+{
+  container::
+    container.new('jaeger-agent', 'jaegertracing/jaeger-agent:1.14.0') +
+    container.withArgs([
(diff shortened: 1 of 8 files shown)


environments/kubernetes/kube-thanos.libsonnet
```diff
@@ -17,6 +19,30 @@ local rolebinding = k.rbac.v1.roleBinding;
...
+    container.withPorts([
+      container.portsType.newNamed(6831, 'jaeger-thrift'),
+      container.portsType.newNamed(5778, 'configs'),
+    ]),
+
+  local jaegerAgentFlag = |||
+    --tracing.config=
+      type: JAEGER
+      config:
+        agent_host: 'localhost'
```
jpkrohling: If it's using the default Jaeger Go Client, you don't need any config at all, as it defaults to this exact setting. You might want to set the sampling strategy explicitly though, as the default is a probabilistic sampler with a 1 in 1000 chance (0.001).

metalmatze: I believe that I need to provide "something" so that Thanos will enable tracing as it's turned off otherwise. But good to know, then I'll rather adjust the sampling to get everything for now.

environments/kubernetes/kube-thanos.libsonnet
```diff
@@ -17,6 +19,30 @@ local rolebinding = k.rbac.v1.roleBinding;
 (import 'thanos-receive-controller/thanos-receive-controller.libsonnet') +
 (import '../../components/thanos-querier-cache.libsonnet') +
 {
+  local jaegerAgent =
```
brancz: Not super important here, but it seems that it would be very useful to have a mixin (non-monitoring-mixin, just a language paradigm mixin) for this, as this will come up a lot I imagine. Thinking declarative sidecar injections.

metalmatze: For a start I can at least refactor it into another package/directory, so that other _could_ grab that folder only.

brancz: Sounds good!

jpkrohling:
Sorry, my bad, I should have caught that. You probably figured that out already, but you are missing the service name:

https://github.com/jpkrohling/configuration/blob/c0f0faa1d855b6880365f71953a50a78c667e66c/environments/kubernetes/kube-thanos.libsonnet#L58-L60

This is how it looks like in my branch:

```jsonnet
                    '--tracing.config=%s' % std.manifestYamlDoc(traceConfig {
                      config+: {
                        service_name: 'querier',
                      },
                    }),
```

brancz:
We're trying out one more time on another env and then will proceed with the merge.

metalmatze:
Alright. PTAL now it's ready for merging. I've tested the Jaeger Thanos integration on my local cluster.

components/jaeger-agent.libsonnet
```diff
@@ -0,0 +1,31 @@
...
+      container.portsType.newNamed(6831, 'jaeger-thrift'),
+      container.portsType.newNamed(5778, 'configs'),
+    ]),
+
+  thanosFlag:: |||
+    --tracing.config=
+      type: JAEGER
+      config:
+        service_name: %s
+        sampler_type: ratelimiting
```
jpkrohling: nice choice!

PR #108, opened 2019-11-01
Add maxSurge=0 and maxUnavailable=1 to scale down jaeger before deploying

metalmatze:
As we only have one PVC we want to scale down this deployment entirely before running another pod with the PVC

/cc @aditya-konarde @brancz


brancz:
should it have been a statefulset to begin with?

metalmatze:
Maybe, however I was pointed to some Deployments from upstream stuff :man_shrugging:

PR #110, opened 2019-11-05
Added missing jaeger flag to Thanos store.

bwplotka:
Signed-off-by: Bartek Plotka <bwplotka@gmail.com>

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -726,6 +726,13 @@ objects:
           - --grpc-address=0.0.0.0:10901
           - --http-address=0.0.0.0:10902
           - --objstore.config=$(OBJSTORE_CONFIG)
+          - |
+            --tracing.config=
+              type: JAEGER
+              config:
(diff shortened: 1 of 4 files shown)


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -1190,7 +1197,7 @@ parameters:
 - name: NAMESPACE
   value: telemeter
 - name: THANOS_IMAGE
-  value: improbable/thanos
+  value: quay.io/quay/thanos
```
kakkoyun: Just for your information, we override these variables before we deploy them. That's why they are a bit outdated.

kakkoyun:
@bwplotka There are issues with the formatting I believe.

bwplotka:
PTAL all should be good now (:

environments/openshift/manifests/observatorium-template.yaml
metalmatze: And it should be `quay.io/thanos-io/thanos`.

PR #111, opened 2019-11-05
Fixed default args for kube-thanos.

bwplotka:
Signed-off-by: Bartek Plotka <bwplotka@gmail.com>


brancz:
I think this is generated via jsonnet.

PR #113, opened 2019-11-12
Add resource requests and limits to containers without

metalmatze:
Add missing resource requests and limits for containers that don't have that.

One final one is missing, but that's not defined in this repo, but rather a dependency. For that I've created: https://github.com/observatorium/thanos-receive-controller/pull/35


components/jaeger-agent.libsonnet
```diff
@@ -15,6 +15,8 @@ local containerEnv = container.envType;
       containerEnv.fromFieldPath('NAMESPACE', 'metadata.namespace'),
       containerEnv.fromFieldPath('POD', 'metadata.name'),
     ]) +
+    container.mixin.resources.withRequests({ cpu: '32m', memory: '16Mi' }) +
```
brancz: Judging by current usage these would be more applicable:

24 Mi memory request
10m CPU request

PR #115, opened 2019-11-13
Add termination grace period to Thanos pods

kakkoyun:
This PR adds similar changes as in https://github.com/thanos-io/kube-thanos/pull/65, since Observatorium runs master version of Thanos.

### DIFF
--- a/environments/kubernetes/kube-thanos.libsonnet
+++ b/environments/kubernetes/kube-thanos.libsonnet
@@ -4,6 +4,7 @@ local service = k.core.v1.service;
 local configmap = k.core.v1.configMap;
 local sts = k.apps.v1.statefulSet;
 local deployment = k.apps.v1.deployment;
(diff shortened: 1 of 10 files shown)


brancz:
Feel free to merge at your own discretion.

metalmatze:
These changes should be done upstream with kube-thanos. Any reason not to do that?

kakkoyun:
@metalmatze Already done, that's in the issue https://github.com/thanos-io/kube-thanos/pull/65 that I've referred until it gets released this should do the work.

brancz:
Do we have v0.9.0 deployed?

kakkoyun:
@brancz Not, yet. This is not an urgent one, I'm just doing the grunt work.

kakkoyun:
Blockers:
1) CI fails already fixed in another issue, it won't be a problem when we merge it #122, before this.
2) Only safe to run on Thanos v0.9.0 and above :)

PR #118, opened 2019-11-13
jsonnetfile: Bump to newer telemeter for virt_platform_type

smarterclayton:
Picks up https://github.com/openshift/telemeter/pull/263


smarterclayton:
@squat @LiliC

PR #120, opened 2019-11-19
Fix Jaeger component monitoring issues

kakkoyun:
This PR attempts to fix issues with monitoring jaeger components.

* Adds a new admin service for jaeger collector. (https://www.jaegertracing.io/docs/1.15/monitoring/)
* Fixes `ServiceMonitor` to point correct service.
* ~~Adds a `PodMonitor` to monitor `jaeger-agent`s~~
* Adds a `ServiceMonitor` to select and monitor `jaeger-agent`s
(https://www.jaegertracing.io/docs/1.15/deployment/#agent)
* Adds liveness probes for `jaeger-agent`s (https://github.com/jaegertracing/jaeger/blob/master/pkg/healthcheck/handler.go)

### DIFF
--- a/components/jaeger-agent.libsonnet
+++ b/components/jaeger-agent.libsonnet
@@ -17,10 +17,15 @@ local containerEnv = container.envType;
     ]) +
     container.mixin.resources.withRequests({ cpu: '32m', memory: '16Mi' }) +
     container.mixin.resources.withLimits({ cpu: '128m', memory: '64Mi' }) +
+    container.mixin.livenessProbe.withFailureThreshold(5) +
+    container.mixin.livenessProbe.httpGet.withPath('/').withPort(self.metricsPort).withScheme('HTTP') +
     container.withPorts([
       container.portsType.newNamed(6831, 'jaeger-thrift'),
       container.portsType.newNamed(5778, 'configs'),
+      container.portsType.newNamed(self.metricsPort, 'metrics'),
     ]),
+
+  metricsPort:: 14271,
 
   thanosFlag:: |||
     --tracing.config=
@@ -30,4 +35,8 @@ local containerEnv = container.envType;
         sampler_type: ratelimiting
         sampler_param: 2
   |||,
+
+  labels:: {
+    'app.kubernetes.io/tracing': 'jaeger-agent',
+  },
 }
--- a/components/jaeger-agent.libsonnet
+++ b/components/jaeger-agent.libsonnet
@@ -26,7 +26,12 @@ local containerEnv = container.envType;
     ]),
 
   metricsPort:: 14271,
-
+  serviceLabels:: {
+    'app.kubernetes.io/name': 'jaeger-agent',
+  },
+  labels:: {
+    'app.kubernetes.io/tracing': 'jaeger-agent',
+  },
   thanosFlag:: |||
     --tracing.config=
       type: JAEGER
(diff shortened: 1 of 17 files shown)


metalmatze:
Sadly we cannot use PodMonitors yet, as the deployed version of the Prometheus Operator on these cluster doesn't have that CRD yet. 
Why do we need this other new admin service? Is this where the metrics are exposed?

Otherwise this looks good to me.

kakkoyun:
@metalmatze Yes, that’s where metrics and probes exposed. 

About Pod Monitors, my bad I didn’t check the Prometheus Operator version. 

So do you have any clever way other than injecting a service for jaeger agent to each pod?

Or should I create a single over-arching service and inject labels to pods, to select all pods that has jaeger agent?

kakkoyun:
@metalmatze I went with the over-arching service option, I'd like to hear your opinions on this.

@jpkrohling What do you think? Do you have any other recommended strategies?

components/jaeger-agent.libsonnet
```diff
@@ -17,11 +17,21 @@ local containerEnv = container.envType;
     ]) +
     container.mixin.resources.withRequests({ cpu: '32m', memory: '16Mi' }) +
     container.mixin.resources.withLimits({ cpu: '128m', memory: '64Mi' }) +
+    container.mixin.livenessProbe.withFailureThreshold(5) +
```
jpkrohling: Interesting, were you having problems with single samples? If so, perhaps we need to add this kind of configuration by default as well via Jaeger Operator. Or is it a best-practice to set a threshold to a value like this?

components/jaeger-collector.libsonnet
```diff
@@ -75,8 +98,11 @@ local service = k.core.v1.service;
         ) +
         container.withVolumeMounts([mount.new('jaeger-store-data', '/var/jaeger/store')]) +
         container.mixin.readinessProbe.withFailureThreshold(3) +
+        container.mixin.readinessProbe.withPeriodSeconds(30) +
+        container.mixin.readinessProbe.withInitialDelaySeconds(30) +
```
jpkrohling: Are you really getting up to 30s for the Jaeger service to start? That's an awful lot of time :)

kakkoyun: 🙈 Just my default values, I'll fix them.

components/jaeger-agent.libsonnet
kakkoyun: They are just my default values, to begin with. Just premature configuration. But if you have any recommended settings, I'd love to take look at them

jpkrohling: I think that's a good value. We could probably add something similar in the operator, to avoid false-positives.

components/jaeger-collector.libsonnet
jpkrohling: 10s should be more than enough, but not sure how long Jaeger needs to start when using badger with a big pre-existing data set. How long is Jaeger typically taking to start in your production instances?

kakkoyun: We haven't measured that, I'll have a look.

kakkoyun:
cc @bwplotka

kakkoyun:
cc @squat

kakkoyun:
These changes caused a similar issue as in https://github.com/grafana/loki/issues/480

PR #121, opened 2019-11-19
Inject Observatorium tag to dashboards

kakkoyun:


metalmatze:
These are pretty generic for Thanos and not really for Observatorium itself. We also wanted to create Observatorium specific dashboards soon. Do we really want this then? 
Just asking, no hard feelings about this.

kakkoyun:
Same here, I don't have any strong opinions. I just suggested this because it'd make it easier to query dashboards we have for Observatorium. I used "observatorium" as an overarching project name in this context. Since it's just a downstream configuration, I don't see any harm.

PR #122, opened 2019-11-22
Add Thanos Rule

kakkoyun:
This PR,
* upgrades dependencies
* adds/modifies manifests to add Thanos Rule


kakkoyun:
cc @squat

environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -994,6 +1089,8 @@ objects:
           - --whitelist={__name__="node_role_os_version_machine:cpu_capacity_cores:sum"}
           - --whitelist={__name__="node_role_os_version_machine:cpu_capacity_sockets:sum"}
           - --whitelist={__name__="subscription_sync_total"}
+          - --whitelist={__name__="csv_succeeded"}
```
bwplotka: Is this change related?

kakkoyun: This is here because I have also updated dependencies, which comes from https://github.com/openshift/telemeter/pull/253/files. They should've created a PR for here as well.

Should I exclude this change and create another PR for it?

bwplotka: No it's fine, I think thanks for context!

metalmatze:
Is there anything left to do for this one? 😊

kakkoyun:
@metalmatze If you give your blessing, I'll merge it :)

environments/kubernetes/manifests/thanos-receive-statefulSet-default.yaml
```diff
@@ -109,10 +109,6 @@ spec:
           requests:
             cpu: 32m
             memory: 16Mi
-      volumes:
-      - configMap:
-          name: observatorium-tenants-generated
-        name: observatorium-tenants
```
metalmatze: We still want this, no?

kakkoyun: I overrode the volume claims that comes from kube-thanos to fix this issue.

kakkoyun:
Rebased after template splits.

kakkoyun:
cc @metalmatze Could you have another look at it?

metalmatze:
The component should probably be named Ruler when we first merge and introduce this already. :blush:

kakkoyun:
Maybe we should do the renaming, as another PR when upstream proposal also happens. Right now, `thanos` and `kube-thanos` refer it as `thanos rule`. At least I have to change it on `kube-thanos` first.

So maybe we can merge this one since this also has pvc/volume fixes in it.

environments/kubernetes/manifests/thanos-store-statefulSet.yaml
```diff
@@ -95,7 +95,6 @@ spec:
           requests:
             cpu: 32m
             memory: 64Mi
-      volumes: null
```
kakkoyun: @metalmatze Do you think this change would create a problem?

metalmatze: I don't think so, as this has been null before. But I'm not 100% sure.

brancz: I think it should be equivalent.

kakkoyun:
@metalmatze To fix volume issues, this disables https://github.com/thanos-io/kube-thanos/pull/76 hiding the volumes.

squat:
I think this was the right move: https://github.com/observatorium/configuration/pull/122#issuecomment-565504394

otherwise it's quite funny to see half thanos-rule and half-thanos-ruler. It seems like a label mismatch just waiting to happen.

kakkoyun:
@metalmatze @squat  Changes in https://github.com/thanos-io/kube-thanos/pull/77 are also reflected now.

PR #128, opened 2019-12-03
Split observatorium-template into smaller templates

metalmatze:
Basically we want to break up the big gigantic `environments/openshift/manifests/observatorium-template.yaml` into smaller pieces. Thus we now have:

```
environments/openshift/manifests/jaeger-template.yaml
environments/openshift/manifests/prometheus-template.yaml
environments/openshift/manifests/telemeter-template.yaml
environments/openshift/manifests/thanos-template.yaml
```

This will allow us to more independently deploy the individual components. 
For example, updating Thanos won't ever trigger a change in Telemeter or Jaeger.

/cc @brancz @kakkoyun @squat @bwplotka @aditya-konarde


metalmatze:
Agreed @squat. I think the benefit will be not deploying Telemeter when changing something on Thanos or Jaeger.

metalmatze:
Let's make sure to squash when merging.

PR #130, opened 2019-12-04
Increase jaeger agent memory limit

kakkoyun:
This PR increases the memory limits of jaeger-agents.

jaeger-agent, by default, allocates a queue to store spans before it sends it to the collector.
> Each queue is 1000 span batches. Given that each span batch has up to 64KiB worth of spans, each queue can hold up to 64MiB worth of span.

Normally this queue is mostly empty but when the collector is down it quickly reaches saturation. In our setup we have a limit of `64MiB` and it needs slightly more than that, so it gets OOMkilled.

ref: https://www.jaegertracing.io/docs/1.15/performance-tuning/#adjust-server-queue-sizes

cc @aditya-konarde @brancz @bwplotka


aditya-konarde:
Lgtm

PR #135, opened 2019-12-09
Use Jaeger in-memory store for testing if badger is the problem

metalmatze:
Let's use the memory storage for now, to see if we still have problems with that.

/cc @jpkrohling @kakkoyun @squat @brancz @bwplotka

### DIFF
--- a/components/jaeger-collector.libsonnet
+++ b/components/jaeger-collector.libsonnet
@@ -82,10 +82,13 @@ local jaegerAgent = import './jaeger-agent.libsonnet';
       local c =
         container.new($.jaeger.deployment.metadata.name, j.image) +
         container.withArgs([
+          // TODO: Use again, once we move back to badger, currently testing the memory store
           // '--badger.directory-key=/var/jaeger/store/keys',
           // '--badger.directory-value=/var/jaeger/store/values',
           // '--badger.ephemeral=false',
           '--collector.queue-size=4000',
+          '--memory.max-traces 1000',
+
         ],) +
         container.withEnv([
           env.new('SPAN_STORAGE_TYPE', 'memory'),
--- a/components/jaeger-collector.libsonnet
+++ b/components/jaeger-collector.libsonnet
@@ -82,13 +82,13 @@ local jaegerAgent = import './jaeger-agent.libsonnet';
       local c =
         container.new($.jaeger.deployment.metadata.name, j.image) +
         container.withArgs([
-          '--badger.directory-key=/var/jaeger/store/keys',
-          '--badger.directory-value=/var/jaeger/store/values',
-          '--badger.ephemeral=false',
+          // '--badger.directory-key=/var/jaeger/store/keys',
+          // '--badger.directory-value=/var/jaeger/store/values',
+          // '--badger.ephemeral=false',
           '--collector.queue-size=4000',
         ],) +
         container.withEnv([
-          env.new('SPAN_STORAGE_TYPE', 'badger'),
+          env.new('SPAN_STORAGE_TYPE', 'memory'),
         ]) + container.withPorts(
           [
             containerPort.newNamed(14250, 'grpc'),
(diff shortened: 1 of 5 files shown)


environments/kubernetes/jaeger.libsonnet
```diff
@@ -1,6 +1,6 @@
 (import '../../components/jaeger-collector.libsonnet') + {
   jaeger+:: {
     namespace:: 'observatorium',
-    image:: 'jaegertracing/all-in-one:1.14.0',
+    image:: 'jaegertracing/all-in-one:1.15.0',
```
jpkrohling: 1.15.1 is available. Perhaps 1.15, so that the latest 1.15 is used?

environments/openshift/jaeger.jsonnet
```diff
@@ -95,7 +95,7 @@ local app =
   parameters: [
     { name: 'NAMESPACE', value: 'telemeter' },
     { name: 'IMAGE', value: 'jaegertracing/all-in-one' },
-    { name: 'IMAGE_TAG', value: '1.14.0' },
+    { name: 'IMAGE_TAG', value: '1.15.0' },
```
jpkrohling: `s/1.15.0/1.15`, or `s/1.15.0/1.15.1` ?

components/jaeger-collector.libsonnet
```diff
@@ -82,13 +82,13 @@ local jaegerAgent = import './jaeger-agent.libsonnet';
       local c =
         container.new($.jaeger.deployment.metadata.name, j.image) +
         container.withArgs([
-          '--badger.directory-key=/var/jaeger/store/keys',
-          '--badger.directory-value=/var/jaeger/store/values',
-          '--badger.ephemeral=false',
+          // '--badger.directory-key=/var/jaeger/store/keys',
```
bwplotka: I think we have git history for this, but I know it's temporary. If anything, `TODO` would be nice (:

components/jaeger-collector.libsonnet
```diff
@@ -82,13 +82,13 @@ local jaegerAgent = import './jaeger-agent.libsonnet';
...
+          // '--badger.directory-value=/var/jaeger/store/values',
+          // '--badger.ephemeral=false',
```
jpkrohling: You have to set this, otherwise it will OOM quickly :) 

```
--memory.max-traces int
```

environments/kubernetes/jaeger.libsonnet
squat: We should always *fully* specify the semver of the images so that deployments are fully reproducible and will always run the same containers.

jpkrohling: That's fair.

components/jaeger-collector.libsonnet
```diff
@@ -82,13 +82,16 @@ local jaegerAgent = import './jaeger-agent.libsonnet';
...
         container.withArgs([
-          '--badger.directory-key=/var/jaeger/store/keys',
-          '--badger.directory-value=/var/jaeger/store/values',
-          '--badger.ephemeral=false',
+          // TODO: Use again, once we move back to badger, currently testing the memory store
+          // '--badger.directory-key=/var/jaeger/store/keys',
+          // '--badger.directory-value=/var/jaeger/store/values',
+          // '--badger.ephemeral=false',
           '--collector.queue-size=4000',
+          '--memory.max-traces 1000',
```
jpkrohling: The other arguments have `=` between the key and the value. Perhaps it would be good to have it here as well, for consistency's sake?

environments/kubernetes/manifests/jaeger-deployment.yaml
```diff
@@ -21,14 +21,12 @@ spec:
     spec:
       containers:
       - args:
-        - --badger.directory-key=/var/jaeger/store/keys
-        - --badger.directory-value=/var/jaeger/store/values
-        - --badger.ephemeral=false
         - --collector.queue-size=4000
+        - --memory.max-traces 1000
```
jpkrohling: In this case, shouldn't the argument be wrapped in quotes? Something to improve in the generator? :)

squat:
we can close this, we have moved this stuff to another repo

PR #148, opened 2019-12-19
Fix Thanos Ruler Query API service discovery issue

kakkoyun:
cc @squat @metalmatze @bwplotka 

xref: https://github.com/thanos-io/kube-thanos/pull/80


kakkoyun:
It's a bit tiresome, true. However, we want to keep maintaining upstream dependencies (`kube-thanos` in this case) as well. So a small price to pay.

And most of the cases this round trip happens when we actually use immature APIs, so it'll be less.

PR #149, opened 2020-01-01
Consume authorize_url from Secret

maorfr:
This PR intends to take the authorization URL out of the template and into the telemeter-server Secret.

Please let me know if you think this is a good idea!

Can you also point out where I'm missing a change in jsonnet?
Update: depends on https://github.com/openshift/telemeter/pull/288

Thanks!

### DIFF
--- a/environments/kubernetes/manifests/telemeter-secret.yaml
+++ b/environments/kubernetes/manifests/telemeter-secret.yaml
@@ -1,5 +1,6 @@
 apiVersion: v1
 data:
+  authorize_url: ""
   client_id: ""
   client_secret: ""
   oidc_issuer: ""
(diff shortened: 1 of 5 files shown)


metalmatze:
The changes would need to be made in this file in the Telemeter repository:
https://github.com/openshift/telemeter/blob/master/jsonnet/telemeter/server/kubernetes.libsonnet
With the changes being merged over there we can then update the Telemeter dependency in this repository by running `jb update`.

maorfr:
wondering what i'm doing wrong @metalmatze

metalmatze:
Updating all jsonnet dependencies broke the current jsonnet stack.

maorfr:
I can update only telemeter manually. How is the hash calculated?

maorfr:
@metalmatze care to take a look again?

metalmatze:
I am still not sure if we really want this in our templates. @squat what do you think?

squat:
We discussed with Maor a few weeks ago. It is not a change we like but, due to details/limitations of the saas-herder stack [1], it is required in order to enable deploying configmaps and other assets from outside of the app-interface repo. For that reason I think that the benefit outweighs the cost :/

[1]  the tooling requires all https urls to be in a secret, as it incorrectly assumes that it is sensitive information.

maorfr:
these changes are now in master, closing this one.

Issue #123, opened 2019-11-28
Update OpenShift Resources with Jinja Template support


kakkoyun:

kakkoyun:
obsolete.

PR #144, opened 2019-12-19
Add  PrometheusRules to Kubernetes manifests and Openshift templates

metalmatze:
We can add these PrometheusRule files to our OpenShift templates and deploy them like that to our namespace. The Prometheus monitoring those is finally able to pick them up from our namespace now.

/cc @aditya-konarde @squat @kakkoyun

PR #150, opened 2020-01-07
Add Thanos Ruler service monitor

kakkoyun:
This PR,
* adds Thanos Ruler service monitor
* refactors others to use templating

xref: #123


environments/sre/servicemonitors/observatorium-jaeger.servicemonitor.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   - port: admin-http
   namespaceSelector:
     matchNames:
-    - telemeter-production
+    - '{{namespace}}'
```
squat: Hmm I’m not convinced about putting more layers of te playing into this repo. This is no longer usable by the community and also means we cannot use this manifests in e2e tests

kakkoyun: Good point. I'm going to create a dedicated directory for this specific use case and leave others as is.

kakkoyun: I've ended up making them parametric so anyone can easily override `environment` and `namespace` values downstream.

kakkoyun:
As a follow up to this, I'll create another PR to refactor out service monitors as we did in #144, then we won't need additional templating.

kakkoyun:
ping @metalmatze @squat Let's merge this and create another one to move service monitors as #144

environments/sre/servicemonitors/observatorium-jaeger.servicemonitor.yaml
metalmatze: I agree with @squat. The environment should be obsolete entirely, as the Prometheus scrape only their environment now.

squat:
rock on

PR #151, opened 2020-01-07
Update kube-thanos to get anti-affinity for queriers and receivers

metalmatze:
/cc @squat @kakkoyun


brancz:
Is it possible to add node region to this in a generic way?

metalmatze:
@brancz, not sure if there's a standard for these labels on nodes. I'd say spreading out across nodes is a good default. There might be people only running in a single region entirely.

squat:
Yes there are standardized labels for both zone and region. These are discussed in the well-known labels doc from the k8s documentation:

https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone

brancz:
Very cool! I wasn’t aware of it! In that case we should additionally add preferred at scheduling and ignored at runtime for zone mode affinity.

PR #153, opened 2020-01-14
Add Observatorium API manifests

kakkoyun:
cc @brancz


kakkoyun:
@metalmatze how about now?

environments/openshift/observatorium.libsonnet
```diff
@@ -0,0 +1,129 @@
...
+  apiVersion: 'v1',
+  kind: 'Template',
+  metadata: {
+    name: 'observatorium',
+  },
+  objects: [
+    $.observatorium.api[name]
+    for name in std.objectFields($.observatorium.api)
+  ],
+  parameters: [
```
squat: this part about adding the parameters is specific to the templating part; should this go in the observatorium.jsonnet, as that file is for generating the template?

kakkoyun: Yes, actually better idea.

environments/openshift/manifests/observatorium-api-template.yaml
```diff
@@ -0,0 +1,147 @@
...
+      port: 8080
+      targetPort: 8080
+    - name: https
+      port: 8081
+      targetPort: https
+    selector:
+      app.kubernetes.io/name: observatorium-api
+parameters:
+- name: NAMESPACE
+  value: telemeter
```
metalmatze: I don't think this namespace exists anywhere, but it will be overwritten anyway.

PR #154, opened 2020-01-15
Fix namespace issues with service DNS

kakkoyun:
cc @aditya-konarde

### DIFF
--- a/environments/kubernetes/observatorium.libsonnet
+++ b/environments/kubernetes/observatorium.libsonnet
@@ -1,7 +1,6 @@
 (import 'observatorium/observatorium-api.libsonnet') {
   observatorium+:: {
-    local namespace = 'observatorium',
-    namespace:: namespace,
+    namespace:: 'observatorium',
 
     local kt = (import 'kube-thanos.libsonnet') + {
       thanos+:: {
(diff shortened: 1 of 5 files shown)


environments/kubernetes/observatorium.libsonnet
```diff
@@ -1,10 +1,17 @@
-local kt = (import 'kube-thanos.libsonnet');
-
 (import 'observatorium/observatorium-api.libsonnet') {
   observatorium+:: {
-    namespace:: 'observatorium',
+    local namespace = 'observatorium',
```
squat: Why is this local necessary? Vs `namespace:: 'observatorium'`

kakkoyun: Actually it's not. Just a common pattern that we used in this repo, in case we need it in current config. I just followed it in here as well.

kakkoyun: If you prefer simpler version, I can fix it?

squat: Yes please. I’d prefer we not establish any more unneeded jsonnet patterns I think in some of those other uses we actually used the local variable more than once

PR #169, opened 2020-01-31
*: Adapt how metrics whitelist is synced

brancz:
This is continuing the change of having the metrics whitelist be defined not in the telemeter repo but in its downstream usages (the client in cluster-monitoring-operator, and the server in the production manifests which are synced from the cluster-monitoring-operator's manifests).

Related to (and somewhat depends on) https://github.com/openshift/telemeter/pull/306.

@squat @metalmatze @kakkoyun @LiliC


environments/openshift/manifests/telemeter-template.yaml
```diff
@@ -172,6 +172,54 @@ objects:
...
+          - --whitelist={__name__="up"}
+          - --whitelist={__name__="cluster_version"}
+          - --whitelist={__name__="cluster_version_available_updates"}
+          - --whitelist={__name__="cluster_operator_up"}
+          - --whitelist={__name__="cluster_operator_conditions"}
+          - --whitelist={__name__="cluster_version_payload"}
+          - --whitelist={__name__="cluster_installer"}
+          - --whitelist={__name__="cluster_infrastructure_provider"}
+          - --whitelist={__name__="cluster_feature_set"}
+          - --whitelist={__name__="node_uname_info"}
```
LiliC: Hmm something wrong here, you already have this metric on line 138 of this fine.

brancz: We need https://github.com/openshift/telemeter/pull/306 to merge first and then bump the telemeter jsonnet dependency. Then the duplicate whitelist entries should disappear.

brancz:
I'll incorporate this in my refactoring PR.

PR #170, opened 2020-02-04
*: Refactor to not mutate global state

brancz:
This is a WIP PR (it does build! :slightly_smiling_face: ) that demonstrates the new configuration system. There are a few things missing:

* [x] telemeter
* [x] AMS prometheus
* [x] alerts/dashboards

The rest should be functionally equivalent to master.

@bwplotka @metalmatze @kakkoyun @squat @krasi-georgiev


components/observatorium.libsonnet
```diff
@@ -0,0 +1,144 @@
...
+      service.new(
+        'thanos-receive',
+        { 'app.kubernetes.io/component': 'thanos-receive' },
+        [
+          ports.newNamed('grpc', 10901, 10901),
+          ports.newNamed('http', 10902, 10902),
+          ports.newNamed('remote-write', 19291, 19291),
+        ]
+      ) +
+      service.mixin.metadata.withNamespace(obs.config.namespace),
```
metalmatze: It would still be better to move this out of here, to have a simple looping over things to get manifests. Why doesn't it fit into the other component?

brancz: Yeah I've thought about that as well, but I'm unsure where to put it .. this is essentially a "system global object that we only need once" which doesn't really align with the rest of the receive components. If you have a suggestion I'm all ears. I'm also happy to start a "global" section, wdyt?

metalmatze: I would have said that we should have somehow put it in the same obs.receivers object, but then discovered that that is nested 2 levels. Putting it into the top level one would then break the level 2 for-loop. 
Other than leaving it here (which might just want end up happening then) I could think of having another `obs.receiversService` for the sake of having 
```
{
    ['thanos-receive-' + hashring + '-' + name]: obs.receivers[hashring][name]
    for hashring in std.objectFields(obs.receivers)
    for name in std.objectFields(obs.receivers[hashring])
}  + {
    ['thanos-receive-' + name]: obs.receiversService[name]
    for name in std.objectFields(obs.receiversService)
}
```

metalmatze: At the same time, having the 2 level nested component feels a bit like a layer violation. Maybe we can think of a different approach to have the `obs.receiver` still expose all objects in the top level? The nesting of multiple receive hashrings seems more like a implementation detail from a component point of view. :thinking:

brancz: Why do you say two levels? I see only the level of where we loop per hashring, which is pretty much exactly what we want. I wanted to avoid flattening the objects immediately as this way we can treat each hashring and patch each hashring individually and the flattening is really just a thing to output. In reality the objects should be nested I think.

build.sh
```diff
@@ -8,20 +8,24 @@ set -x
...
 
 # Make sure to start with a clean 'manifests' dir
-rm -rf environments/kubernetes/manifests
-mkdir environments/kubernetes/manifests
+rm -rf environments/dev/manifests
+mkdir environments/dev/manifests
 
-jsonnet -J vendor -m environments/kubernetes/manifests environments/kubernetes/main.jsonnet | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml' -- {}
-find environments/kubernetes/manifests -type f ! -name '*.yaml' -delete
+jsonnet -J vendor -m environments/dev/manifests environments/dev/main.jsonnet | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml' -- {}
```
squat: Since this is literally the same contents as build_dev.sh, it might be better to do `./build_dev.sh` so we don't have to keep two files in sync

build.sh
```diff
@@ -7,21 +7,20 @@ set -x
 # only exit with zero if all commands of the pipeline exit successfully
 set -o pipefail
 
+./build_dev.sh
```
squat: :+1:

PR #179, opened 2020-02-12
environments/base: Bump to thanos v0.10.1

brancz:
In preparation for a prod rollout, let's test this version here already.

@squat @metalmatze @kakkoyun @bwplotka @krasi-georgiev

### DIFF
--- a/environments/base/default-config.libsonnet
+++ b/environments/base/default-config.libsonnet
@@ -3,7 +3,7 @@
 
   name: 'observatorium-xyz',
   namespace: 'observatorium',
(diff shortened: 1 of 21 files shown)


bwplotka:
Alse let's be careful as maybe some flag name changed etc.. (I don't 100% remember, but I think nothing was changed)

brancz:
Image can be configured individually but by default all thanos components use the same. Yes the OpenShift template has template variables as overrides.

bwplotka:
Amazing! :+1:

PR #180, opened 2020-02-13
use authorize-url from Secret - attempt #2

maorfr:
This PR intends to take the authorization URL out of the template and into the telemeter-server Secret.

replaces #149

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -2301,7 +2301,7 @@ objects:
           - --tls-crt=/etc/pki/service/tls.crt
           - --internal-tls-key=/etc/pki/service/tls.key
           - --internal-tls-crt=/etc/pki/service/tls.crt
-          - --authorize=${AUTHORIZE_URL}
+          - --authorize=$(AUTHORIZE_URL)
           - --oidc-issuer=$(OIDC_ISSUER)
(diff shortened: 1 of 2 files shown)


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -1454,6 +1454,11 @@ objects:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
+          - name: AUTHORIZE_URL
```
brancz: this change first needs to go into the telemeter-server jsonnet code

maorfr: i thought i did them in https://github.com/openshift/telemeter/pull/288

metalmatze: Yes and they were pulled in with Telemeter as dependency. So this LGTM now.

maorfr:
@metalmatze @brancz anything i can do to push this forward?

metalmatze:
We would need to fix the CI.
Can you run `./build.sh` and commit those changes?

environments/openshift/manifests/observatorium-template.yaml
brancz: Looks like we missed something there: https://github.com/openshift/telemeter/pull/322

brancz:
Closing in favor of #229

PR #185, opened 2020-02-13
Added Makefile for better discoverability of commands, reuse and binaries versioning. Also added version `-in-container`.

bwplotka:
See: make help


Makefile
```diff
@@ -0,0 +1,121 @@
...
+format: ## Auto-formats jsonnet files.
+	@which jsonnetfmt 2>/dev/null || ( \
+		echo "Cannot find jsonnetfmt command, please install from https://github.com/google/jsonnet/releases. If your C++ does not support GLIBCXX_3.4.20, please use xxx-in-container target like format-in-container." && exit 1 \
+	)
+
+	@echo ">> formatting jsonnet"
+	@find . -type f -not -path './vendor/*' \( -name '*.libsonnet' -o -name '*.jsonnet' \) | xargs -n 1 -- $(JSONNET_FMT)
+
+.PHONY: gen
+gen: ## Generates manifests for all environments from jsonnet.
```
kakkoyun: Just for sake of consistency between repos, should we rename this as `generate`?

Makefile
```diff
@@ -0,0 +1,121 @@
...
+gen: deps
+	@ENV="dev" $(MAKE) gen-env
+	@ENV="base" $(MAKE) gen-env
+	@ENV="openshift" $(MAKE) gen-env
+
+.PHONY: gen-env
```
kakkoyun: ditto. maybe `generate-env`?

Makefile
```diff
@@ -0,0 +1,121 @@
...
+gen-in-container: ## Generates manifests for all environments from jsonnet in docker container.
+	@echo ">> Compiling and generating thanos-mixin"
+	$(JSONNET_CONTAINER_CMD) $(MAKE) $(MFLAGS) JSONNET_BUNDLER='/go/bin/jb' deps
+	$(JSONNET_CONTAINER_CMD) $(MAKE) $(MFLAGS) \
+		JSONNET='/go/bin/jsonnet' \
+		JSONNET_BUNDLER='/go/bin/jb' \
+		GOJSONTOYAML='/go/bin/gojsontoyaml' \
+		gen
+
+$(GO_JSONNET):
```
kakkoyun: So for these tool dependencies, could we use the new module approach rather than using a script? https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

The biggest advantage is we can track all our dependencies with `go.mod` and `go.sum` files.
We have already established the same approach for API gateway and telemeter, see [Makefile] (https://github.com/observatorium/observatorium/blob/master/Makefile#L121) and [tools.go](https://github.com/observatorium/observatorium/blob/master/tools.go) for observatorium/observatorium.

Makefile
bwplotka: :+1:

Makefile
bwplotka: Hmmm. What do you think merging this approach and tackle it in separate PR? Also would be nice to move Thanos for this as well.

It's not that easy here as the configuration is *not* a go module... so ... ;p It's weird (:

kakkoyun: Yeah, that's soo true. 🤦‍♂ We don't need it for this repo. Yep, let's merge this.
For Thanos, I'm happy to contribute.

bwplotka: :heart: Remember! Don't promise ;p

bwplotka:
@metalmatze ?

build_deploy.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
-# This is a placeholder file needed by the CI/CD templates
-# You can safely ignore this file :)
-exit 0
```
metalmatze: We need this file for AppSRE.
Eventually for exactly that reason I'd say we move to our internal AppSRE version of this repo to not have this stuff confuse people in the public. :expressionless:

Makefile
```diff
@@ -0,0 +1,121 @@
...
+# Version from 12.12.2019
+GOJSONTOYAML_VERSION    ?= bf2969bbd742d117a9524b859fb417fefb67565d
+GOJSONTOYAML            ?= $(GOBIN)/gojsontoyaml-$(GOJSONTOYAML_VERSION)
+
+# Version from 9.02.2020. We use https://github.com/google/go-jsonnet for ease of building.
+GO_JSONNET_VERSION         ?= 31b9ace0f65a5a5717ed9521b57d8d9e18f24070
+GO_JSONNET                 ?= $(GOBIN)/jsonnet-$(GO_JSONNET_VERSION)
+# v0.2.0
+JSONNET_BUNDLER_VERSION ?= 184841238bb5df1b886ccaecd3e179bc55c17687
+JSONNET_BUNDLER         ?= $(GOBIN)/jb-$(JSONNET_BUNDLER_VERSION)
```
metalmatze: Why not pin the actual release instead of the sha?

Makefile
```diff
@@ -0,0 +1,121 @@
...
+GOPROXY           ?= https://proxy.golang.org
+GIT               ?= $(shell which git)
+
+# Version from 12.12.2019
+GOJSONTOYAML_VERSION    ?= bf2969bbd742d117a9524b859fb417fefb67565d
+GOJSONTOYAML            ?= $(GOBIN)/gojsontoyaml-$(GOJSONTOYAML_VERSION)
+
+# Version from 9.02.2020. We use https://github.com/google/go-jsonnet for ease of building.
+GO_JSONNET_VERSION         ?= 31b9ace0f65a5a5717ed9521b57d8d9e18f24070
+GO_JSONNET                 ?= $(GOBIN)/jsonnet-$(GO_JSONNET_VERSION)
```
metalmatze: I actually use the C++ version most of the time as that one has `jsonnetfmt` which the Go version still has not... :roll_eyes: :expressionless:

Makefile
```diff
@@ -0,0 +1,121 @@
+DEFAULT_GOAL := help
+
+# Ensure everything works even if GOPATH is not set, which is often the case.
+# The `go env GOPATH` will work for all cases for Go 1.8+.
+GOPATH            ?= $(shell go env GOPATH)
+
+TMP_GOPATH        ?= /tmp/thanos-go
```
metalmatze: Seems to be a copy-paste leftover.

build_deploy.sh
bwplotka: I mean... no comment here. (:

Makefile
bwplotka: Just how the script that is proven for Thanos takes commit.

Makefile
bwplotka: Ok, but we use Go for a reason. 

I think we can install one from release page easily if you prefer this, exactly as CI was doing.

bwplotka:
> Basically is jsonnet-bundler v0.2 installed? If not, fail.

I disagree. I spent way too much time downloading a million different tools from different places. It might be easy for you when you maintain jsonnet-bundler on daily basic. But please put yourself on our position.

Also if you don't automate they always end up being in different versions than on CI so it creates a burden. It has to be automated. Or let's ask from other side: Why it cannot be automated? 

> Personally, I'm not a fan of looping or overcomplicating Makefile. They should be simple so that at a glance it also helps as a guide.

At the same time, I am not a fan of looping or overcomplicating inconsistent build, build_dev empty deploy bash scripts. I doubt it was more readable than this, but happy to hear more opinions. **Even worse, what we had previously** is that CI was using some handwritten commands, different to our script, so chances to waste time were huge.

> As for the generation itself, I think that that could be a lot simpler too by not having those targes PHONY but actually check if the files are updated enough.
> Snipped from another repo doing a similar thing

I saw that example and because I am super noob in jsonnet I have no idea what files I should hook into. ): But I can imagine I would need to have at least 5 rules like this (for each `main-*.jsonnet`? What if it imports something and that changed? I doubt Makefile would detect this). Overall lots or repetition, increased complexity, chances of making errors or Makefile actually missing a change. ): So why we cannot have 3 lines code that does everything as we want?

Makefile
bwplotka: No, needed for install function.

Makefile
kakkoyun: Or we can only use c++ version of `jsonnetfmt` and keep `go-jsonnet` one for other uses.

kakkoyun: You could find this useful:
https://github.com/observatorium/observatorium/blob/e009b4adfef6b4134073310f8e84bb3aa4215cde/Makefile#L127-L132

brancz:
I agree with half of each of your comments.

> As for the generation itself, I think that that could be a lot simpler too by not having those targes PHONY but actually check if the files are updated enough.

I think this is true, and mostly why I like makefiles. If everything is a phony target then it's equivalent to a bash script but harder to maintain in my opinion.

> Also if you don't automate those bins always end up being in different versions than on CI or someone will require new one and it's lots of manual steps, so it creates a burden. It has to be automated. Or let's ask from the other side: Why it cannot be automated?

I do think this should be automated and pinned precisely. Maybe to keep the main makefile easier to understand we can keep this in a separate Makefile that just takes care of that .. maybe called `Makefile.tools` ?

rollandf:
Having the fmt and other checks done in the CI in the Makefile is important.
How can we progress on this PR?

brancz:
This is quite outdated. @rollandf if you want to give this another shot I'd be happy with that! :)

bwplotka:
I also wrote something that will help us to install versioned Go tools:
https://github.com/bwplotka/gobin will write more about it this weekend.

kakkoyun:
This needs to be revisited after changes in #271. Or maybe we should close this one and start over.

metalmatze:
I would say, let's take it in small steps from what we currently have.

kakkoyun:
@bwplotka Shall we close this one? It's a bit stale and probably we'd consider a better implementation using https://github.com/bwplotka/bingo

bwplotka:
Let's start over. We have bingo and things to apply... ((:

PR #187, opened 2020-02-18
Update the metrics list in Telemeter

metalmatze:
We want to update the list of metrics Telemeter accepts from our upstream:
https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster_monitoring_operator_04-config.yaml

/cc @kakkoyun @brancz @squat @s-urbaniak @lilic 

The command I copied this file with:
```
cat ~/src/github.com/openshift/cluster-monitoring-operator/manifests/0000_50_cluster_monitoring_operator_04-config.yaml | gojsontoyaml -yamltojson | jq -r '.data."metrics.yaml"' | gojsontoyaml -yamltojson | jq .matches > environments/openshift/metrics.json```

### CODE REVIEW

File: environments/openshift/metrics.json [hunk 1]
```diff
@@ -1,6 +1,7 @@
 [
   "{__name__=~\"cluster:usage:.*\"}",
-  "{__name__=\"up\"}",
```
brancz: this is something new, and we will need to fix our sync strategy for this.. we have to have the whitelist be the union of all whitelists we had previously, meaning this should still be in there, in addition to the new ones

environments/openshift/metrics.json
```diff
@@ -9,7 +10,6 @@
   "{__name__=\"cluster_installer\"}",
   "{__name__=\"cluster_infrastructure_provider\"}",
   "{__name__=\"cluster_feature_set\"}",
-  "{__name__=\"node_uname_info\"}",
```
brancz: same as above

environments/openshift/metrics.json
metalmatze: Yeah. I was wondering about that.

metalmatze:
> Could you please put this command in a script file if it's something we need to do regularly? or in a doc?

Eventually, this should live within the Telemeter repo again IMHO. I'd rather try to get there quickly, than documenting this workaround. If that's ok.

metalmatze:
Did I already say that jq is amazing?

`jq -s 'add | unique' a.json b.json` merges to arrays and the unique filter creates a sorted array with unique elements.

I've split this PR into 3 commits. First one sorts the original metrics.json file. The second one actually merges the update into that file. The third one then creates the script I used to generate the new file :)

PTAL

s-urbaniak:
lgtm :+1:

PR #189, opened 2020-02-18
Refactor API Gateway to comply with new conventions

### LINKED ISSUE #188: Observatorium API manifest not generated
The manifest for the Observatorium API is not generated for base environment

kakkoyun:
This PR,
* removes additional OpenShift template and reuses existing one to deploy API GW (88b09076e3147393f2c1ba48e11f47a5d62476ae)
  * reflect changes from https://github.com/observatorium/observatorium/pull/12
* adds additional Thanos Querier for API Gateway UI (a504fc353d82b6fca8d0a4401242c598e0a53af1)
* modifies end-to-end tests to use API Gateway (4069cad01f807bf8ece47c66e902f685e69c3a32)

Fixes #188 

cc @brancz @squat @bwplotka


jsonnetfile.json
```diff
@@ -44,11 +44,11 @@
       "name": "observatorium",
       "source": {
         "git": {
-          "remote": "https://github.com/observatorium/observatorium",
```
nmagnezi: Shouldn't this be the other way around?

jsonnetfile.json
```diff
@@ -44,11 +44,11 @@
...
+          "remote": "https://github.com/kakkoyun/observatorium",
           "subdir": "jsonnet/lib"
         }
       },
-      "version": "master"
+      "version": "jsonnet_refactor"
```
nmagnezi: ditto

jsonnetfile.lock.json
```diff
@@ -59,12 +59,12 @@
       "name": "observatorium",
       "source": {
         "git": {
-          "remote": "https://github.com/observatorium/observatorium",
```
nmagnezi: same here?

nmagnezi:
Related [issue](https://github.com/observatorium/configuration/issues/188)

jsonnetfile.json
kakkoyun: It's still in a PR, that hasn't merged. I will update it before we merge this.

rollandf:
Should we update the endpoints in configuration/tests/manifests/observatorium-up.yaml ?

kakkoyun:
@rollandf Thanks for pointing that out, I'll fix it as well.

brancz:
Can we add some comments in there somewhere why we're adding the second querier? :) I think we've spoken about it offline but would be good if we documented it as a code comment or something.

kakkoyun:
@brancz  👍 I will. Thanks for reminding.

metalmatze:
I guess I kind of missed the discussion around putting all components into the `observatorium-template.yaml`. We might want to discuss that, and if we want to have a gigantic one for a bunch of smaller ones before merging this PR, as I already started spreading out things before the end of the last year.

components/up.libsonnet
```diff
@@ -0,0 +1,79 @@
...
+    name: error 'must provide name',
+    version: error 'must provide version',
+    image: error 'must provide image',
+    backoffLimit: error 'must provide backoffLimit',
+    writeEndpoint: error 'must provide writeEndpoint',
+    readEndpoint: error 'must provide readEndpoint',
+
+    commonLabels:: {
+      'app.kubernetes.io/name': 'observatorium-up',
+      'app.kubernetes.io/instance': up.config.name,
```
squat: This is not quite right I think, we end up with both `instance` and `name` being `observatorium-up`.  Instance should not be the name of the application

metalmatze: Hmm. That's what we do in a couple of places actually... :thinking: 
https://github.com/thanos-io/kube-thanos/blob/master/jsonnet/kube-thanos/kube-thanos-receive.libsonnet#L17

squat: but the difference is that the `name` there ends up being "default" for the default hashringm here it is exactly the same as the app name

brancz: Isn’t this just coincidental that we happen to name our instance of this observatorium-up component “observatorium-up” ? This seems ok to me.

squat: Well it could be coincidental, but then the question is: what does it mean? I don't know what the 'observatorium-up' instance refers to. If we had multiple ups and sorted by instance label, would you know what this one was?

brancz: If I understand correctly you’re suggesting we should choose a more descriptive name for the instance than “observatorium-up” to reflect its purpose rather than _what_ it is? I’m all for that. Suggestions what to name it?

squat: Kind of, the instance label exists to tell us what instance this app is. "Observatorium-up'" is not an instance, it's an app name. "default" is an instance, "testing"could be an instance, "prod" could even be an instance. I guess I'm also not really sure why this is here. Is it only for e2e tests? Then "e2e" can be an instance

brancz: e2e-test sounds good to me as that is actually it’s purpose here right now. We would have a different name for when we introduce it into the base setup as a Blackbox probe.

kakkoyun: The label is updated now. 
These definitions will be moved to `observatorium/up` projects eventually. I've put them as they solely needed in here right now.

rollandf:
Fixmie seems stuck....

PR #192, opened 2020-02-19
Observatorium Operator

nmagnezi:
Please review and provide feedback so we can incorporate your thoughts.

Signed-off-by: Nir Magnezi <nmagnezi@redhat.com>
Co-authored-by: Fred Rolland <frolland@redhat.com>


Makefile
```diff
@@ -0,0 +1,60 @@
+# Image URL to use all building/pushing image targets
+IMG ?= quay.io/nmagnezi/observatorium-operator:latest
```
brancz: created a new quay repo for this https://quay.io/repository/observatorium/observatorium-operator?namespace=observatorium

Makefile
```diff
@@ -0,0 +1,60 @@
...
+# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
```
brancz: why do we want this?

Makefile
```diff
@@ -0,0 +1,60 @@
...
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifeq (,$(shell go env GOBIN))
+GOBIN=$(shell go env GOPATH)/bin
+else
+GOBIN=$(shell go env GOBIN)
+endif
+
+# Install CRDs into a cluster
+install: manifests
+	kustomize build config/crd | kubectl apply -f -
```
brancz: is there something specific why we need to use kustomize for this?

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// Objest Storage Configuration
+	ObjectStorageConfig ObjectStorageConfig `json:"objectStorageConfig"`
+	// Hashrings describes a list of Hashrings
+	Hashrings []*Hashring `json:"hashrings"`
+	// Thanos CompactSpec
+	Compact CompactSpec `json:"compact"`
+	// Thanos Receive Controller Spec
+	ReceiveController ReceiveControllerSpec `json:"thanosReceiveController"`
+	// Thanos ThanosPersistentSpec
+	Receivers ThanosPersistentSpec `json:"receivers"`
```
brancz: experience has shown that having separate types that embed the "generic" type have worked best

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ObservatoriumSpec defines the desired state of Observatorium
+type ObservatoriumSpec struct {
+	// Thanos Image name
+	Image *string `json:"thanosImage"`
```
brancz: why a pointer to a string?

deploy/operator.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: observatorium-operator
```
brancz: could we as a follow up generate these with jsonnet as well?

deploy/operator.yaml
```diff
@@ -0,0 +1,29 @@
...
+  labels:
+    control-plane: observatorium-operator
```
brancz: let's add the kubernetes app labels here like we do everywhere else

deploy/operator.yaml
```diff
@@ -0,0 +1,29 @@
...
+        control-plane: observatorium-operator
+    spec:
+      containers:
+      - name: observatorium-operator
+        image: quay.io/nmagnezi/observatorium-operator:latest
+        args:
+          - "--log-level=debug"
+        resources:
+          limits:
+            cpu: 1
```
brancz: have we observed resource requirements this high?

deploy/cluster_role.yaml
```diff
@@ -0,0 +1,100 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/name: observatorium-operator
+    app.kubernetes.io/version: v0.1
+  name: observatorium-operator
```
brancz: if we have some time, this would be amazing if we could implement in locutus that locutus generates the RBAC roles it needs as it knows the steps it might do upfront

Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Build the manager binary
+FROM golang:1.13 as builder
+WORKDIR /workspace
+# Copy the jsonnet source
+COPY environments/operator/ environments/operator/
+COPY components/ components/
+COPY jsonnetfile.json jsonnetfile.json
+# Build
+RUN GO111MODULE="on" go get github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
+RUN jb install
```
brancz: cpaas builds are offline only, so for the downstream fork we will need to check in the `vendor/` directory from jsonnet as well, which reminds me that we should try to split the go and jsonnet vendor directories

.gitignore
```diff
@@ -2,3 +2,4 @@
 /kubectl
 vendor/
 tmp/
+.idea
```
metalmatze: You shouldn't add this to every project. Please use a global gitignore on your machine like:
https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

Makefile
rollandf: Fixed.
Still with previous image in the operator.yaml until we have an imaged pushed.

Makefile
rollandf: Removed

Makefile
rollandf: No need, we got it from kube-builder. Removed.

api/v1alpha1/observatorium_types.go
rollandf: Done

api/v1alpha1/observatorium_types.go
rollandf: Changed to string only.

deploy/operator.yaml
rollandf: Done

deploy/operator.yaml
rollandf: Done

.gitignore
rollandf: Done

Dockerfile
rollandf: Should this be done in this PR or a follow-up one?

deploy/operator.yaml
rollandf: Should this be done in this PR or a follow-up one?

Dockerfile
brancz: not in this repo at all I think, let's take care of this downstream

deploy/operator.yaml
brancz: happy to have this be a follow up

Makefile
```diff
@@ -0,0 +1,58 @@
...
+	set -e ;\
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4 ;\
+	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
+	}
+CONTROLLER_GEN=$(GOBIN)/controller-gen
+else
+CONTROLLER_GEN=$(shell which controller-gen)
```
squat: this is probably not a great idea since the generators vary quite a bit between different versions. If we cannot be sure of the installed version then we cannot guarantee the build will work

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+WORKDIR /
+COPY --from=builder /go/bin/locutus .
+COPY --from=builder /workspace/environments/operator /environments/operator
+COPY --from=builder /workspace/components/ /components/
+COPY --from=builder /workspace/vendor/ /vendor/
+RUN chgrp -R 0 /vendor && chmod -R g=u /vendor
+RUN chgrp -R 0 /environments && chmod -R g=u /environments
+RUN chgrp -R 0 /components && chmod -R g=u /components
+ENTRYPOINT ["/locutus", "--renderer=jsonnet", "--renderer.jsonnet.entrypoint=environments/operator/main.jsonnet", "--trigger=resource", "--trigger.resource.config=environments/operator/config.yaml"]
```
squat: nit: we should have a newline here

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Copy the jsonnet source
+COPY environments/operator/ environments/operator/
+COPY components/ components/
+COPY jsonnetfile.json jsonnetfile.json
+# Build
+RUN GO111MODULE="on" go get github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
+RUN jb install
+RUN GO111MODULE="on" go get github.com/brancz/locutus
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
squat: Should we really use this base image for the upstream container?

api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,35 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "obs-api.observatorium.io", Version: "v1alpha1"}
```
squat: I don't think `obs-api` is a very useful group name. All `observatorium.io` API groups will probably be `obs ` and they are all already `API` by definition. `obs-api` doesn't actually add any information IMO. Also, `obs` is not really a meaningful/legible string for most people, me included :)

maybe simply "core" would be better

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// Hashrings describes a list of Hashrings
+	Hashrings []*Hashring `json:"hashrings"`
+	// Thanos CompactSpec
+	Compact CompactSpec `json:"compact"`
+	// Thanos Receive Controller Spec
+	ReceiveController ReceiveControllerSpec `json:"thanosReceiveController"`
+	// Thanos ThanosPersistentSpec
+	Receivers ReceiversSpec `json:"receivers"`
+	// Thanos QuerierCache
+	QueryCache QuerierCacheSpec `json:"queryCache"`
```
squat: We should try to be consistent here with naming.

Above, we use `CompactSpec` and `ReceiveSpec` but here we use `QuerierCacheSpec`. Shouldn't it be `QueryCacheSpec`?

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// Thanos StoreSpec
+	Store StoreSpec `json:"store"`
+	// Thanos RulerSpec
+	Rule RuleSpec `json:"rule"`
+	// API Gateway
+	ApiGateway ApiGatewaySpec `json:"apiGateway"`
```
squat: Following golang naming conventions, this should be `APIGateway` and `APIGatewaySpec`

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,143 @@
...
+type ObjectStorageConfig struct {
+	// Object Store Config Secret Name for Thanos
+	Name string `json:"name"`
+	// Object Store Config key for Thanos
+	Key string `json:"key"`
+}
+
+type ReceiveControllerSpec struct {
+	Image string `json:"image"`
+	// Tag describes the tag of Thanos receive controller to use.
```
squat: There is no field called `Tag`, maybe you meant `Version`?

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// RetentionResolutionRaw
+	RetentionResolution1h string `json:"retentionResolution1h"`
+}
+
+type VolumeClaimTemplate struct {
+	VolumeClaimTemplateSpec v1.PersistentVolumeClaimSpec `json:"spec"`
+}
+
+type QuerierCacheSpec struct {
+	// Thanos Querier Cache  name
```
squat: Is this really the name? Or is this the image?

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type ObjectStorageConfig struct {
+	// Object Store Config Secret Name for Thanos
+	Name string `json:"name"`
+	// Object Store Config key for Thanos
+	Key string `json:"key"`
+}
+
+type ReceiveControllerSpec struct {
```
squat: Let's do a follow up to this PR to improve the doc comments for this file. It will help with generated documentation later.

deploy/operator.yaml
```diff
@@ -0,0 +1,32 @@
...
+      control-plane: observatorium-operator
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        control-plane: observatorium-operator
+    spec:
+      containers:
+      - name: observatorium-operator
+        image: quay.io/nmagnezi/observatorium-operator:latest
```
squat: Do we need to update this?

environments/operator/config.yaml
```diff
@@ -0,0 +1,5 @@
+mainResource: obs
+resources:
+- name: obs
+  kind: Observatorium
+  apiVersion: obs-api.observatorium.io/v1alpha1
```
squat: nit: add a newline here

api/v1alpha1/observatorium_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// RetentionResolutionRaw
+	RetentionResolutionRaw string `json:"retentionResolutionRaw"`
+	// RetentionResolutionRaw
+	RetentionResolution5m string `json:"retentionResolution5m"`
+	// RetentionResolutionRaw
+	RetentionResolution1h string `json:"retentionResolution1h"`
+}
+
+type VolumeClaimTemplate struct {
+	VolumeClaimTemplateSpec v1.PersistentVolumeClaimSpec `json:"spec"`
```
squat: Nit: this field seems to be a bit overqualified IMO, `X.VolumeClaimTemplate.VolumeClaimTemplateSpec.Y` is quite repetitive. Consider calling it simply `Spec`

Dockerfile
rollandf: Any reason why not use this image? It is open for use for everybody.
https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image

Dockerfile
rollandf: Done

Makefile
rollandf: This was generated by kubebuilder. How would you like to handle this?

api/v1alpha1/groupversion_info.go
rollandf: @brancz Any thoughts on what should be the group name?
Isn't "core" too generic?

api/v1alpha1/observatorium_types.go
rollandf: Done

api/v1alpha1/observatorium_types.go
rollandf: Done

api/v1alpha1/observatorium_types.go
rollandf: Sure

api/v1alpha1/observatorium_types.go
rollandf: Done

api/v1alpha1/observatorium_types.go
rollandf: Done

api/v1alpha1/observatorium_types.go
rollandf: Done

deploy/operator.yaml
rollandf: Yes, we will once we get approval on the rest of the PR, and when the operator will be pushed to the final location.
It is needed now for QE.

environments/operator/config.yaml
rollandf: Done

api/v1alpha1/groupversion_info.go
squat: IMO it is more meaningful than `obs-api`. An API group named `api` is the most generic API group i can think of. I'd prefer to put some thought into this now rather than have to change API groups later down the road

Dockerfile
squat: mainly because we don't use it for anything else in the thanos or prometheus ecosystem

Makefile
squat: I'd prefer to install this dependency the same way we install go dependencies in all of our other projects, e.g. up, observatorium, thanos-receive-controller:
https://github.com/observatorium/observatorium/blob/master/tools.go
https://github.com/observatorium/observatorium/blob/master/Makefile#L118

nmagnezi:
@brancz @metalmatze @squat 

ci/circleci: test-operator fails because we have no permissions to push images to your quay repo. Please advise.

```
  Type     Reason     Age                    From               Message
  ----     ------     ----                   ----               -------
  Normal   Scheduled  6m27s                  default-scheduler  Successfully assigned default/observatorium-operator-547ff9cd9b-vcxx7 to minikube
  Normal   Pulling    4m49s (x4 over 6m26s)  kubelet, minikube  Pulling image "quay.io/observatorium/observatorium-operator:latest"
  Warning  Failed     4m47s (x4 over 6m23s)  kubelet, minikube  Failed to pull image "quay.io/observatorium/observatorium-operator:latest": rpc error: code = Unknown desc = Error response from daemon: manifest for quay.io/observatorium/observatorium-operator:latest not found: manifest unknown: manifest unknown
  Warning  Failed     4m47s (x4 over 6m23s)  kubelet, minikube  Error: ErrImagePull
  Normal   BackOff    4m21s (x7 over 6m22s)  kubelet, minikube  Back-off pulling image "quay.io/observatorium/observatorium-operator:latest"
  Warning  Failed     81s (x20 over 6m22s)   kubelet, minikube  Error: ImagePullBackOff
```

deploy/operator.yaml
nmagnezi: Done

api/v1alpha1/groupversion_info.go
nmagnezi: Done

Dockerfile
nmagnezi: Done

Makefile
nmagnezi: Done

rollandf:
Ping

Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+# Image URL to use all building/pushing image targets
+IMG ?= quay.io/observatorium/observatorium-operator:latest
+BIN_DIR ?= $(shell pwd)/tmp/bin
+
+CONTROLLER_GEN ?= $(BIN_DIR)/controller-gen
+
+# Install CRDs into a cluster
+install: manifests
+	kubectl apply -f deploy/crds
```
squat: Do we actually use this anywhere in the stack? I don't think it's in the tests, no?

Makefile
```diff
@@ -0,0 +1,46 @@
...
+vet:
+	go vet ./...
+
+# Generate code
+generate: $(CONTROLLER_GEN)
+	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
+
+# Build the docker image
+docker-build:
+	docker build . -t ${IMG}
```
squat: We need to add this to the circle ci config so that images are built in an isolated env rather than by hand on a Deb's machine. Also, we avoid using `latest`. We should use a unique tag per commit/date eg https://github.com/observatorium/up/blob/master/.drone.yml#L52

example/core_v1alpha1_observatorium.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: core.observatorium.io/v1alpha1
+kind: Observatorium
+metadata:
+  name: observatorium-xyz
```
squat: We should make sure this object gets gets the common labels that all other resources have

squat:
We can do fix my new comments in followups but at least the container building and image tagging is quite important

Makefile
rollandf: Removed

Makefile
rollandf: Done, following https://github.com/observatorium/observatorium/blob/master/.circleci/config.yml

example/core_v1alpha1_observatorium.yaml
rollandf: Done

.circleci/config.yml
```diff
@@ -36,3 +45,24 @@ workflows:
...
+      - build
+      - tests
+      - test-operator
+      - container-push:
+          requires:
+            - build
+          filters:
+            branches:
+              only:
+                - master
```
squat: nit: let's put a newline here

Dockerfile
```diff
@@ -21,4 +21,30 @@ COPY --from=builder /workspace/vendor/ /vendor/
...
+ARG BUILD_DATE
+ARG VERSION
+ARG VCS_REF
+ARG DOCKERFILE_PATH
+ARG VCS_BRANCH
+
+LABEL vendor="Observatorium" \
+    name="observatorium/configuration" \
+    description="Observatorium Operator" \
+    io.k8s.display-name="observatorium/configuration" \
```
squat: all of these should rather be `observatorium/observatoriumoperator` no?

.circleci/config.yml
rollandf: Done

Dockerfile
rollandf: Done

nmagnezi: Done

example/core_v1alpha1_observatorium.yaml
```diff
@@ -0,0 +1,63 @@
...
+  app.kubernetes.io/component: observatorium-cr
+  app.kubernetes.io/instance: observatorium-xyz
+  app.kubernetes.io/name: observatorium-cr
+  app.kubernetes.io/part-of: observatorium
+spec:
+  thanosImage: "quay.io/thanos/thanos:"
```
squat: hm this is kind of weird, no? The images really should not end in a `:`

example/core_v1alpha1_observatorium.yaml
```diff
@@ -0,0 +1,63 @@
...
+    name: thanos-objectstorage
+  hashrings:
+    # - hashring: hashring0
+    #  tenants:
+    #    - foo
+    #    - bar
+    - hashring: default
+      tenants: []
+  queryCache:
+    image: "quay.io/cortexproject/cortex:"
```
squat: same for the rest of these

example/core_v1alpha1_observatorium.yaml
rollandf: See here:
https://github.com/observatorium/configuration/blob/master/environments/base/default-config.libsonnet#L7

squat: sure, but this is still incorrect. why isn't this `thanosImage: "quay.io/thanos/thanos:master-2020-02-13-adfef4b5`

rollandf: Done

example/core_v1alpha1_observatorium.yaml
rollandf: Done

deploy/cluster_role.yaml
nmagnezi: Hi @brancz, how do we plan to proceed with this?

squat: I think this is a nice-to-have that we can do later. It doesn't change the API surface of the operator or the external behavior, just the amount of manifests that we need to generate. This PR is sufficiently larger so let's get it merged :muscle:

PR #194, opened 2020-02-24
*: Introduce store sharding

brancz:
I left everything at 1 shard for the base and dev environment but 3 shards for the prod env.

@bwplotka @squat @metalmatze @kakkoyun @krasi-georgiev

### DIFF
--- a/components/observatorium-configure.libsonnet
+++ b/components/observatorium-configure.libsonnet
@@ -21,7 +21,10 @@
   },
 
   store+:: {
-    config+:: obs.config.store,
+    ['shard' + i]+: {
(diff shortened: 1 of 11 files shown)


metalmatze:
If I read that correctly then on production we're going to shard to 3 nodes by modulo the ULID on those 3 nodes?

brancz:
Correct, the ULID itself is already well distributed but hashmod also uses a cryptographic hashfunction so either way it'll be well distributed.

brancz:
Haha, yes staging first, if that looks roughly like what we expect then we can promote to prod.

PR #195, opened 2020-02-24
*: Label store shard objects appropriately

brancz:
This way each store shard's service will actually only select its shard.

@metalmatze @squat @bwplotka @krasi-georgiev @kakkoyun


brancz:
>  (which should be low footprint, but still)?

Correct.

PR #199, opened 2020-02-26
components/jaeger: Reduce retention to 3h

brancz:
By default the retention is 72h, which at consistent load explodes with
the 50G disk.

https://www.jaegertracing.io/docs/1.17/cli/#jaeger-all-in-one-badger

@squat @metalmatze @bwplotka @krasi-georgiev @aditya-konarde


metalmatze:
We don't really nerd more than 3 hours of traces as of right now.

brancz:
So far we've almost exclusively only used it for ad hoc immediate analysis. Happy to extend the disk and retention, if this shows to solve our problem.

Issue #200, opened 2020-03-03
build_dev fails with RUNTIME ERROR: Field does not exist: service


rollandf:
With commit 5c400432ad2fadd0e7e3f0f4f5ba2503a5c3316d when running ./build_dev.sh, we get the following error:

```
$ ./build_dev.sh 
+ set -o pipefail
+ rm -rf environments/dev/manifests
+ mkdir environments/dev/manifests
+ jsonnet -J vendor -m environments/dev/manifests environments/dev/main.jsonnet
+ xargs '-I{}' sh -c 'cat {} | gojsontoyaml > {}.yaml' -- '{}'
RUNTIME ERROR: Field does not exist: service
	components/observatorium.libsonnet:207:45-62	thunk from <object <anonymous>>
	components/observatorium.libsonnet:206:58-65	
	<std>:688:15-22	thunk <val> from <function <format_codes_arr>>
	<std>:695:27-30	thunk from <thunk <s> from <function <format_codes_arr>>>
	<std>:565:22-25	thunk from <function <format_code>>
	<std>:565:9-26	function <format_code>
	<std>:695:15-60	thunk <s> from <function <format_codes_arr>>
	<std>:700:24-25	thunk from <thunk <s_padded> from <function <format_codes_arr>>>
	<std>:475:30-33	thunk from <thunk from <function <pad_left>>>
	<std>:475:19-34	thunk from <function <pad_left>>
	...
	<std>:750:7-46	function <anonymous>
	<std>:227:7-23	function <anonymous>
		
	vendor/kube-thanos/kube-thanos-query.libsonnet:61:24-29	thunk from <thunk from <thunk <c> from <object <anonymous>>>>
	<std>:227:21-22	thunk from <function <anonymous>>
	<std>:749:17-21	thunk from <function <anonymous>>
	<std>:749:8-22	function <anonymous>
	<std>:227:7-23	function <anonymous>
		
	During manifestation	

```

metalmatze:
Can reproduce this locally, looking into it right now.

PR #201, opened 2020-03-03
Set store and replica-label flag on API Gateway to empty

metalmatze:
As this Querier is just for the UI assets we don't need these store and replica-labels setup.

/cc @kakkoyun @squat @rollandf


metalmatze:
Closes #200

metalmatze:
Resolved the conflicts and tests are green.

PR #209, opened 2020-03-16
Test operator with image build on PR

rollandf:
The current test is pulling the latest image from quay, so it is not testing the changes introduced in PRs.

With this change, an image is build based on the PR and loaded in the cluster using `kind load docker-image`.

### DIFF
--- a/tests/e2e.sh
+++ b/tests/e2e.sh
@@ -45,6 +45,8 @@ wait_for_cr() {
 }
 
 deploy_operator() {
+    docker build -t quay.io/observatorium/observatorium-operator:latest .
+    ./kind load docker-image quay.io/observatorium/observatorium-operator:latest
     $KUBECTL apply -f https://raw.githubusercontent.com/coreos/kube-prometheus/master/manifests/setup/prometheus-operator-0servicemonitorCustomResourceDefinition.yaml
     $KUBECTL apply -f https://raw.githubusercontent.com/coreos/kube-prometheus/master/manifests/setup/prometheus-operator-0prometheusruleCustomResourceDefinition.yaml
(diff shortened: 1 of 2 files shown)


brancz:
Wouldn't it be simpler to build the image and use `kind load-image` instead?

rollandf:
I tried the `kind load docker-image`, but I had issues with it.
I will try to deep dive more into it to see why.

rollandf:
So it works for me only if I specify `"imagePullPolicy": "IfNotPresent"`.
Is it better this way?

brancz:
`"imagePullPolicy": "IfNotPresent"` is what we should have by default anyways. I think we should change the manifest to use that rather and not patch it for the test.

rollandf:
I updated the operator manifest with `"imagePullPolicy": "IfNotPresent"`

tests/e2e.sh
```diff
@@ -55,6 +57,7 @@ deploy_operator() {
     $KUBECTL apply -f environments/dev/manifests/minio-service.yaml
     $KUBECTL apply -f deploy/crds
     $KUBECTL apply -f deploy/
+    $KUBECTL patch deployment observatorium-operator --patch '{"spec": {"template": {"spec": {"containers": [{"name": "observatorium-operator","image": "obs-operator:test"}]}}}}'
```
brancz: if we directly build and load `quay.io/observatorium/observatorium-operator:latest` then we don't need to patch anything here

rollandf: We should not use `latest` according to this: https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster.
But for some reasons it worked...

PR #213, opened 2020-03-19
Upgrade Observatorium API GW

kakkoyun:
Upgrade Observatorium API GW to its latest version

cc @nmagnezi


brancz:
lgtm on green

kakkoyun:
I need to check why the operator fails.

nmagnezi:
You should also update the image tag in the CR example/core_v1alpha1_observatorium.yaml

rollandf:
It would be great to have the CR example generated from https://github.com/observatorium/configuration/blob/master/environments/base/default-config.libsonnet

PR #215, opened 2020-03-19
Generate CRD example

kakkoyun:
It may not be the optimum way to generate but better than nothing I believe.

cc @nmagnezi @rollandf


example/main.jsonnet
```diff
@@ -0,0 +1,64 @@
...
+        hashrings:: null,
+      },
+      apiGateway: {
+        image: obs.config.apiGateway.image,
+        version: obs.config.apiGateway.version,
+      },
+    },
+  },
+
+  manifests+:: {
```
brancz: I don’t think we need to apply this pattern for this, just declare the crd directly I would say.

example/main.jsonnet
```diff
@@ -0,0 +1,52 @@
+local obs = (import '../environments/base/observatorium.jsonnet');
+
+{
+  'observatorium-crd': {
```
brancz: I don’t think this is even necessary just wrote the cr directly

kakkoyun: I hope this time, I understand you correctly :)

build_dev.sh
```diff
@@ -14,5 +14,14 @@ mkdir environments/dev/manifests
...
+rm -rf tests/manifests
+mkdir tests/manifests
+
 jsonnet -J vendor -m tests/manifests tests/main.jsonnet | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml' -- {}
 find tests/manifests -type f ! -name '*.yaml' -delete
+
+rm -rf example/manifests
+mkdir example/manifests
+
+jsonnet -J vendor example/main.jsonnet | gojsontoyaml >example/manifests/observatorium-crd.yaml
```
brancz: I think this should be in a follow up PR, as I don't want to change more than necessary in this PR, but `-crd.yaml` is misleading, as this is a custom resource instance not a custom resource definition. I think in a follow up we should just rename this to `observatorium.yaml`.

brancz:
Not sure why but there still seems to be a CI failure.

kakkoyun:
CI is a bit flaky. I don't know the exact reason. I'm on a bad strake with CI runs recently, CIs just hate me.
I'll re-run it if it fails.

PR #217, opened 2020-03-22
Create minio resources in a seperate namespace

### LINKED ISSUE #216: CI Tests create minio namespace, whereas minio resources reside on Observatorium
This means that both Observatorium CR and minio reside on the same namespace.

nmagnezi:
Fixes #216

### DIFF
--- a/components/minio.libsonnet
+++ b/components/minio.libsonnet
@@ -100,18 +100,18 @@
     kind: 'Secret',
     metadata: {
       name: 'thanos-objectstorage',
-      namespace: minio.config.namespace,
+      namespace: minio.config.bucketSecretNamespace,
     },
     stringData: {
       'thanos.yaml': |||
         type: s3
         config:
           bucket: thanos
-          endpoint: minio:9000
+          endpoint: %s.%s.svc.cluster.local:9000
           insecure: true
           access_key: minio
           secret_key: minio123
-      |||,
+      ||| % [minio.service.metadata.name, minio.config.namespace],
     },
     type: 'Opaque',
   },
--- a/tests/e2e.sh
+++ b/tests/e2e.sh
@@ -62,7 +62,7 @@ deploy_operator() {
 }
 
 run_test() {
-    $KUBECTL wait --for=condition=available --timeout=10m -n observatorium deploy/minio || ($KUBECTL get pods --all-namespaces && exit 1)
+    $KUBECTL wait --for=condition=available --timeout=10m -n minio deploy/minio || ($KUBECTL get pods --all-namespaces && exit 1)
     $KUBECTL wait --for=condition=available --timeout=10m -n observatorium deploy/observatorium-xyz-thanos-query || ($KUBECTL get pods --all-namespaces && exit 1)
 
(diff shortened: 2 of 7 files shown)


components/minio.libsonnet
```diff
@@ -2,15 +2,15 @@
   local minio = self,
 
   config:: {
-    namespace: error 'must provide namespace',
+    s3namespace: error 'must provide namespace',
```
brancz: this is a nested config already, there is no need for a separate name

brancz:
CI still failing, but generally happy with having it in a separate namespace

components/minio.libsonnet
nmagnezi: Done

environments/base/default-config.libsonnet
```diff
@@ -3,6 +3,7 @@
 
   name: 'observatorium-xyz',
   namespace: 'observatorium',
+  s3namespace: 'minio',
```
brancz: this should be under a `minio` section, and then it can be `namespace` again

components/minio.libsonnet
```diff
@@ -107,11 +107,11 @@
         type: s3
         config:
           bucket: thanos
-          endpoint: minio:9000
+          endpoint: %s.%s.svc.cluster.local:9000
           insecure: true
           access_key: minio
           secret_key: minio123
-      |||,
+      ||| % [ minio.config.s3namespace, minio.service.metadata.name ],
```
rollandf: I think the namespace is after the svc name

environments/base/default-config.libsonnet
nmagnezi: Done

components/minio.libsonnet
nmagnezi: Done

environments/base/default-config.libsonnet
```diff
@@ -10,6 +10,11 @@
     key: 'thanos.yaml',
   },
 
+  minio: {
```
brancz: let's only have this in the dev config, nothing but the dev config actually uses minio in any way

components/minio.libsonnet
brancz: we can rename this now that it doesn't clash

nmagnezi: Done

environments/base/default-config.libsonnet
nmagnezi: Done

tests/e2e.sh
```diff
@@ -62,7 +62,7 @@ deploy_operator() {
 }
 
 run_test() {
-    $KUBECTL wait --for=condition=available --timeout=10m -n observatorium deploy/minio || ($KUBECTL get pods --all-namespaces && exit 1)
+    $KUBECTL wait --for=condition=available --timeout=10m -n minio deploy/minio || ($KUBECTL get pods --all-namespaces && exit 1)
```
brancz: since we specify the namespaces, and part of the manifests in here have a different one, maybe we should skip the `-n` flag?

squat: ack, it's overconstrained right now

PR #219, opened 2020-03-24
Update UP Read URL

rollandf:
Based on:
https://github.com/observatorium/observatorium/blob/master/test/integration.sh#L43


rollandf:
@kakkoyun , can you please review when possible?

kakkoyun:
And I guess we should fix the issue in `up`, we should have caught this in the first place.

PR #221, opened 2020-03-24
Enable offline deduplication

kakkoyun:
This PR depends on the latest version of Thanos and kube-thanos.
Depends: https://github.com/observatorium/configuration/pull/222
xref: https://github.com/thanos-io/kube-thanos/pull/105

About flags:
```
--delete-delay=48h

Time before a block marked for deletion is deleted from bucket.

If delete-delay is non zero, blocks will be marked for deletion 
and compactor component will delete blocks marked for deletion from the bucket.

If delete-delay is 0, blocks will be deleted straight away.
Use this if you want to get rid of or move the block immediately.

Note that deleting blocks immediately can cause query failures, 
if store gateway still has the block loaded, or compactor is ignoring the deletion 
because it's compacting the block at the same time.
```
```
--ignore-deletion-marks-delay=24h

Duration after which the blocks marked for deletion will be filtered out while fetching blocks.
The idea of ignore-deletion-marks-delay is to ignore blocks that are marked for deletion with some delay. 

This ensures store can still serve blocks that are meant to be deleted but do not have a replacement yet. 
If delete-delay duration is provided to compactor or bucket verify component, 
it will upload deletion-mark.json file to mark after what duration the block should be deleted rather than deleting the block straight away.

If delete-delay is non-zero for compactor or bucket verify component, 
ignore-deletion-marks-delay should be set to (delete-delay)/2 so that blocks marked for deletion are filtered out while fetching blocks before being deleted from bucket. 
Default is 24h, half of the default value for --delete-delay on compactor.
```
# Changes
* Add deduplication replica label flags daca967a6e36765ab9d0628bbb3f283034f2e461
* Add delete delay flags 57f99e9a2cbf2b6a27a07b8a2960fa0671e30f7b

### DIFF
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -17,6 +17,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     t.compact +
     t.compact.withRetention +
     t.compact.withDownsamplingDisabled +
+    t.compact.withDeleteDelay +
     t.compact.withDeduplication + {
       config+:: {
         local cfg = self,
@@ -25,6 +26,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         replicas: 1,
         commonLabels+:: obs.config.commonLabels,
         deduplicationReplicaLabels: obs.config.deduplicationReplicaLabels,
+        deleteDelay: '48h',
       },
     },
 
@@ -79,7 +81,8 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
 
   store:: {
     ['shard' + i]:
-      t.store {
+      t.store +
+      t.store.withIgnoreDeletionMarksDelay {
         config+:: {
           local cfg = self,
           name: obs.config.name + '-' + cfg.commonLabels['app.kubernetes.io/name'] + '-shard-' + i,
@@ -88,6 +91,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
             'store.observatorium.io/shard': 'shard-' + i,
           },
           replicas: 1,
+          ignoreDeletionMarksDelay: '24h',
         },
       } + {
         statefulSet+: {
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -9,18 +9,22 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       'app.kubernetes.io/part-of': 'observatorium',
       'app.kubernetes.io/instance': obs.config.name,
     },
+    replicaLabels:: ['prometheus_replica', 'rule_replica', 'replica'],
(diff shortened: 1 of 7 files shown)


kakkoyun:
cc @squat @metalmatze @krasi-georgiev

components/observatorium.libsonnet
```diff
@@ -9,18 +9,23 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       'app.kubernetes.io/part-of': 'observatorium',
       'app.kubernetes.io/instance': obs.config.name,
     },
+    replicaLabels:: ['prometheus_replica', 'rule_replica', 'replica'],
```
bwplotka: This is vertical compaction, so.... `prometheus_replica` should not be touched IMO.

bwplotka: Same with rule replica

components/observatorium.libsonnet
```diff
@@ -9,18 +9,23 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+    t.compact.withDeleteDelay +
+    t.compact.withDeduplication + {
       config+:: {
         local cfg = self,
         name: obs.config.name + '-' + cfg.commonLabels['app.kubernetes.io/name'],
         namespace: obs.config.namespace,
         replicas: 1,
         commonLabels+:: obs.config.commonLabels,
+        deduplicationReplicaLabels: obs.config.replicaLabels,
+        deleteDelay: '4h',
```
bwplotka: Why 4h? No harm in default value, no?

components/observatorium.libsonnet
bwplotka: BTW this looks like really offline dedup will help to reduce data even more 2x... right? @brancz @squat

bwplotka: as 2 replicas of Prom are pushing

components/observatorium.libsonnet
kakkoyun: All right, makes sense. I will do the changes.

kakkoyun:
@bwplotka I've created another PR for the upgrade https://github.com/observatorium/configuration/pull/222.

Updated `deduplication.replicate-labels` and changed the `delete-delay` flag values with default ones without deleting them to make them visible.

components/observatorium.libsonnet
brancz: > BTW this tells me that really proper offline dedup will help to reduce data even more 2x... right?

Not yet but eventually. It's specific to our case though, because we don't write directly with remote write, but scrape 1 of the prometheus replicas federation endpoint and convert that to remote write, so we really just have 1 (almost always).

Still

> This is vertical compaction, so.... prometheus_replica should not be touched IMO.

this is not an opinion, this is critical, especially because we _have_ pushed data via both remote write replicas before.

PR #223, opened 2020-03-25
Add OAuth proxy for thanos compactor

kakkoyun:
OAuth proxy is needed to use Thanos compactor UI

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -498,9 +498,73 @@ objects:
       app.kubernetes.io/instance: observatorium
       app.kubernetes.io/name: cortex-query-frontend
       app.kubernetes.io/part-of: observatorium
+- spec:
+    template:
+      spec:
+        containers:
+        - args:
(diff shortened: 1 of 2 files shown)


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -514,6 +578,9 @@ objects:
     - name: http
       port: 10902
       targetPort: http
+    - name: https
+      port: 8080
```
squat: hmm this is funny, are we always using 8080 for https in this repo? it doesn't really matter of course since it's all isolated in the container and used for mapping but by convention we normally use 8080 for http and something like 8443 for https

kakkoyun: We use `9091` for the rest of it, but I believe it's also weird. I'll use `8443`.

squat: ack makes sense, we use 9090 by convention for http in the prometheus-world, so 9091 is often https or any proxy. 8443 sounds :100:

Issue #232, opened 2020-04-01
Go module name


clyang82:
When I use the observatorium api, it throws
```
go: finding github.com/observatorium/configuration latest
go: github.com/open-cluster-management/multicluster-monitoring-operator/cmd/manager imports
	github.com/observatorium/configuration/api/v1alpha1: github.com/observatorium/configuration@v0.0.0-20200331123052-734d6c5722c4: parsing go.mod:
	module declares its path as: github.com/configuration
	        but was required as: github.com/observatorium/configuration
```
I think we should rename the module name to `github.com/observatorium/configuration` in go.mod. Does it make sense? @nmagnezi

nmagnezi:
Hi @clyang82, thanks for reporting the issue!

I'm looking into the module name issue, but can you please explain what you are trying to do? I need to better understand your use case.
 
If you wish to import the observatorium-api, the git repo is [github.com/observatorium/observatorium](https://github.com/observatorium/observatorium)

clyang82:
Hi @nmagnezi , thanks for your time to look at this issue. 

I would like to reuse the types defined in  https://github.com/observatorium/configuration/blob/master/api/v1alpha1/observatorium_types.go

I follow this document to add 3rd party resource - https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#adding-3rd-party-resources-to-your-operator
so the above error occurs.

PR #233, opened 2020-04-01
Fix go module path to include observatorium/configuration

nmagnezi:
This patch will also add a Makefile entry to better handle go vendoring

Fixes: #232


clyang82:
Thanks @nmagnezi

nmagnezi:
@squat @brancz @kakkoyun could you guys please have a look so @clyang82 will not be blocked for this? Thanks 😃

PR #244, opened 2020-04-07
Use the THANOS_COMPACTOR_REPLICAS variable

rrati:


metalmatze:
The actual bug is fixed in https://github.com/thanos-io/kube-thanos/pull/109. Once we merged that one we can simply update kube-thanos in this project and it will be correctly generated from there on.
Thanks for the report @rrati!

PR #248, opened 2020-04-16
thanos: Bump to v0.12.0.

bwplotka:
I hope receiver version change is ok... it was bit old for a reason?

### DIFF
--- a/environments/openshift/manifests/observatorium-template.yaml
+++ b/environments/openshift/manifests/observatorium-template.yaml
@@ -2628,7 +2628,7 @@ parameters:
 - name: THANOS_IMAGE
   value: quay.io/thanos/thanos
 - name: THANOS_IMAGE_TAG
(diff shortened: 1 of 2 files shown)


environments/openshift/manifests/observatorium-template.yaml
```diff
@@ -2642,7 +2642,7 @@ parameters:
 - name: THANOS_RECEIVE_CONTROLLER_IMAGE
   value: quay.io/observatorium/thanos-receive-controller
 - name: THANOS_RECEIVE_CONTROLLER_IMAGE_TAG
-  value: master-2019-10-18-d55fee2
```
squat, line 2645: This is for the receive controller, not receive itself but this tag doesn't exist for the controller

bwplotka: I need to read things fully :facepalm:

PR #249, opened 2020-04-16
Remove telemetric scripts

rollandf:
Not needed as Openshift environment was removed


rollandf:
@krasi-georgiev @kakkoyun

kakkoyun:
@krasi-georgiev Could you merge this when we move them?

krasi-georgiev:
yes just checking why are there 2 script for this. After checking with @brancz he suggested that only the script that adds new whitelisted metrics without removing old ones should be used.

krasi-georgiev:
Here is the new script. The PR also includes the new generated whitelisted values.

https://gitlab.cee.redhat.com/observatorium/configuration/merge_requests/36/diffs

Please review and lets merge:

cc @brancz @lilic @metalmatze

PR #250, opened 2020-04-16
Upgrade Observatorium API GW

kakkoyun:


environments/base/default-config.libsonnet
```diff
@@ -111,7 +111,7 @@
 
   apiGateway: {
     local gwConfig = self,
-    version: 'master-2020-03-19-62d4b8e',
+    version: 'HEAD-2020-04-16-v0.1.0',
```
brancz, line 114: just curious, any reason why we changed the naming scheme here?

kakkoyun: This is what happened with the same script we generate all the versions with a tag. I've fixed it to generate `v0.1.0`  for tagged releases. I just didn't fix the tag retrospectively.

brancz: interesting. no strong opinion, feel free to continue

PR #251, opened 2020-04-17
Allow write for the up tool for deployments

krasi-georgiev:

### DIFF
--- a/components/up.libsonnet
+++ b/components/up.libsonnet
@@ -60,11 +60,14 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     local c =
       container.new('observatorium-up', up.config.image) +
       container.withArgs([
-        '--endpoint-read=' + up.config.readEndpoint,
         '--duration=0',
         '--queries-file=/etc/up/queries.yaml',
         '--log.level=debug',
-      ] + if std.objectHas(up.config, 'writeEndpoint') then [
+      ] +
+      if std.objectHas(up.config, 'readEndpoint') then [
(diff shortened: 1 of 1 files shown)


components/up.libsonnet
```diff
@@ -64,9 +64,13 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         '--duration=0',
         '--queries-file=/etc/up/queries.yaml',
         '--log.level=debug',
-      ]) + container.withPorts([
+      ] + if std.objectHas(up.config, 'writeEndpoint') then [
```
kakkoyun, line 67: If we are going down to this path, let's make the read endpoint optional as well.

krasi-georgiev: added

components/up.libsonnet
```diff
@@ -60,13 +60,20 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     local c =
       container.new('observatorium-up', up.config.image) +
       container.withArgs([
-        '--endpoint-read=' + up.config.readEndpoint,
-        '--duration=0',
-        '--queries-file=/etc/up/queries.yaml',
-        '--log.level=debug',
-      ]) + container.withPorts([
+                           '--duration=0',
```
kakkoyun, line 63: nit: I don't think this indentation changes necessary. Better to organize the code so that formatted wouldn't change it in such a way.

krasi-georgiev: good catch, thanks, I didn't even realize it looks to messed up.


I think it is much better now

PR #252, opened 2020-04-19
Add Locutus as Go dependency

rollandf:
For downstream builds


rollandf:
This is not related to the Jsonnet vendoring. It is a Go package that is in use in this repo.

kakkoyun:
@rollandf I've said that because in that PR we would have changed directories could've touched go dependencies. Nevermind.

rollandf:
Thanks @kakkoyun

Issue #239, opened 2020-04-07
Add "Vendor" directories


rollandf:
For offline builds, there is a need for "vendor" directory.

1. Add vendor for Go libraries.
2. Support different vendor directory location for Jsonnet libraries

rollandf:
Not needed for Go, we use Cachito.
Still needed for Jsonnet.

kakkoyun:
This one is also getting a little bit frustrating while developing.

rollandf:
Fixed by #257

PR #253, opened 2020-04-19
Jsonnet vendoring

rollandf:


rollandf:
Fixes #239

rollandf:
@brancz PTAL

brancz:
As Kemal noted this repo is intended to be used as a dependency of downstream usages, this maens that `jsonnetfile.lock.json` should not be in the root of the directory but rather per environment. We can easily achieve this with local directory dependencies.

rollandf:
> As Kemal noted this repo is intended to be used as a dependency of downstream usages, this maens that `jsonnetfile.lock.json` should not be in the root of the directory but rather per environment. We can easily achieve this with local directory dependencies.

Following up this: 
so there will be a sonnetfile.json under environments/base and another one under environments/dev?
Then we could have the vendor dir under each?
for the operator we could copy the vendor dir from base in the docker file.

Will this work?

PR #257, opened 2020-04-23
Jsonnet vendoring

rollandf:
As suggested by @brancz , move the vendoring to each environment.

Is this breaking downstream project?


metalmatze:
We don't really need a vendor folder per environment. They build on each others based on the same dependencies. 
How about simply having `vendor/` and `vendor-jsonnet/` or something here? :thinking:

brancz:
Base environment should not have a lock file nor should it have the vendor checked in, only the operator environment should. The dev environment should have a local directory dependency pointing to the base environment.

rollandf:
> We don't really need a vendor folder per environment. They build on each others based on the same dependencies.
> How about simply having `vendor/` and `vendor-jsonnet/` or something here? thinking

@metalmatze I have tried this way, but Go mode vendor will scan also  `vendor-jsonnet/`
You can try and see:
jb install --jsonnetpkg-home=jsonnet-vendor
go mod vendor

brancz:
Maybe we can put the go and operator code in a subdirectory?

rollandf:
> Maybe we can put the go and operator code in a subdirectory?

OK, what do you think the structure will be? Where the `jsonnetfile.json` should be?
Don't Go modules implies to have the go.mod in the root directory?

brancz:
> Don't Go modules implies to have the go.mod in the root directory?

Is this really the case? If so I wasn't aware of it.

> Where the jsonnetfile.json should be?

I'm thinking also a separate directory that we just call `jsonnet`.

rollandf:
> A module is a collection of Go packages stored in a file tree with a go.mod file at its root. 
from https://blog.golang.org/using-go-modules

brancz:
The way I read this, `github.com/observatorium/configuration/operator` would be the "root" here of the mod, so `github.com/observatorium/configuration/operator/go.mod` should work just fine.

rollandf:
OK, I will try this way

rollandf:
@brancz PTAL
Do you want the jsonnetfile.json/jsonnetfile.lock.json under operator/jsonnet to be link to the one in root?
Still need to handle some changes in Makefile and circleci.

operator/jsonnet/jsonnetfile.json
```diff
@@ -0,0 +1,89 @@
+{
```
brancz, line 1: this jsonnetfile should only need to point at the base of this repo as a directory type package, like here: https://github.com/coreos/kube-prometheus/blob/master/jsonnetfile.json

the required dependencies are then transitively resolved

rollandf: Here the structure is different, do I tried:

`{
  "version": 1,
  "dependencies": [
    {
      "source": {
        "local": {
          "directory": "../.."
        }
      },
      "version": ""
    }
  ],
  "legacyImports": true
}`

then jb install in `configuration/operator`, then it kind of messed up the directory.
It changed the `jsonnet` dir to a link:
`jsonnet -> /home/frolland/git/go/src/observatorium/configuration`

brancz: That sounds like a bug in jb

brancz: make sure to give the package a name still

rollandf: Where should I give a name?

brancz: `name` is a sibling of `version`

rollandf: Added a name

brancz: sorry, it's a sibling next to the dependency version, I should have been more clear:

```
{
  "version": 1,
  "dependencies": [
    {
      "source": {
        "local": {
          "directory": "../.."
        }
      },
      "version": "",
       "name": "configuration",
    }
  ],
  "legacyImports": true
}
```

rollandf: I still have the issue with the link. If I have the jsonnetfile.json as above, under operator/jsonnet and then I run jb install, the dir turns to a link to the root, and the jsonnet dir content disappear, along with the jsonnetfile.json

rollandf: Solved with jb built from master

brancz:
This still has `WIP` is this ready for review?

rollandf:
I still need to fix the Makefile.
Comments are welcome. :-)

rollandf:
Ready for review.

Do I need to remove the `jsonnetfile.lock.json` in the root directory and add it to `.gitignore`?

brancz:
> Do I need to remove the jsonnetfile.lock.json in the root directory and add it to .gitignore?

yes, we should only have the one from the operator environment as that's what produces the checked in vendoring

rollandf:
@brancz  PTAL

rollandf:
@brancz thanks for the guidance!

PR #260, opened 2020-04-30
Update Locutus version

rollandf:


rollandf:
@metalmatze @kakkoyun @squat @brancz PTAL

PR #263, opened 2020-04-30
Align structs name to tags

### LINKED ISSUE #261: Struct tags should not have different values than their field name
There are a few field names that differ from their struct tag representation. This is a bad practice as it often causes confusion and is also inconsistent. They should be aligned. An example:

https://github.com/observatorium/configuration/blob/c835853933cae03341a19fef1ce8a960b54a2707/api/v1alpha1/observatorium_types.go#L26

@nmagnezi @rollandf

rollandf:
Fixes #261

### DIFF
--- a/api/v1alpha1/observatorium_types.go
+++ b/api/v1alpha1/observatorium_types.go
@@ -23,9 +23,9 @@ import (
 // ObservatoriumSpec defines the desired state of Observatorium
 type ObservatoriumSpec struct {
 	// Thanos Image name
-	Image string `json:"thanosImage"`
+	ThanosImage string `json:"thanosImage"`
 	// Thanos Image version
-	Version string `json:"thanosVersion"`
+	ThanosVersion string `json:"thanosVersion"`
 	// Objest Storage Configuration
 	ObjectStorageConfig ObjectStorageConfig `json:"objectStorageConfig"`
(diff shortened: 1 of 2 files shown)


rollandf:
@brancz @metalmatze @kakkoyun @squat PTAL

api/v1alpha1/observatorium_types.go
```diff
@@ -23,17 +23,17 @@ import (
 // ObservatoriumSpec defines the desired state of Observatorium
 type ObservatoriumSpec struct {
 	// Thanos Image name
-	Image string `json:"thanosImage"`
+	ThanosImage string `json:"thanosImage"`
 	// Thanos Image version
-	Version string `json:"thanosVersion"`
+	ThanosVersion string `json:"thanosVersion"`
```
brancz, line 28: this should be under each component, not one version for all, versions should be individually selectable

brancz, line 28: this should be under each component, not one version for all, versions should be individually selectable

rollandf: This was fixed in another PR

brancz: sorry didn't see that yet, commented there

rollandf:
I will consolidate in the PR #264

Issue #235, opened 2020-04-03
Cannot cleanup the observatorium pod/svc


clyang82:
Cannot cleanup the observatorium pod/svc/... when removing the observatorium CR. we should add the `ownerReferences` to the created pod/svc...

@nmagnezi

brancz:
Agreed

nmagnezi:
Thanks @clyang82, I'll look into it.

clyang82:
@nmagnezi any progress?

clyang82:
Thanks @brancz  @nmagnezi @rollandf

PR #266, opened 2020-05-04
Add ownerReferences for resources created by CR

nmagnezi:
Fixes: #235
Co-authored-by: Fred Rolland <frolland@redhat.com>

### DIFF
--- a/environments/operator/obs-operator.jsonnet
+++ b/environments/operator/obs-operator.jsonnet
@@ -1,3 +1,19 @@
-(import '../../components/observatorium.libsonnet') + {
-  config+:: (import 'operator-config.libsonnet'),
-} + (import '../../components/observatorium-configure.libsonnet')
+local config = import 'operator-config.libsonnet';
+local manifests = ((import '../../components/observatorium.libsonnet') + {
+                     config+:: config,
+                   } + (import '../../components/observatorium-configure.libsonnet')).manifests;
+
+{
+  manifests: std.mapWithKey(function(k, v) v {
+    metadata+: {
+      ownerReferences: [{
+        apiVersion: config.apiVersion,
+        blockOwnerdeletion: true,
+        controller: true,
+        kind: config.kind,
+        name: config.name,
+        uid: config.uid,
+      }],
+    },
+  }, manifests),
+}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,7 @@ jobs:
           ./tests/e2e.sh kind
           ./tests/e2e.sh deploy-operator
           ./tests/e2e.sh test
+          ./tests/e2e.sh delete-cr
 
   container-push:
     machine:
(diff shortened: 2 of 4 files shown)


rollandf:
@metalmatze @brancz @kakkoyun @squat PTAL

environments/operator/obs-operator.jsonnet
```diff
@@ -1,3 +1,18 @@
...
+
+local addOwner(k, v) = v { metadata+: { ownerReferences: [{
+  apiVersion: config.apiVersion,
+  blockOwnerdeletion: true,
+  controller: true,
+  kind: config.kind,
+  name: config.name,
+  uid: config.uid,
+}] } };
+local ownermanifests = std.mapWithKey(addOwner, manifests);
```
brancz, line 14: as you're not using the key, you can just use `std.map`

environments/operator/obs-operator.jsonnet
```diff
@@ -1,3 +1,18 @@
...
+
+{
+  manifests: ownermanifests,
```
brancz, line 17: I would inline the `std.map` call here, no need for the local variables above except the imports

environments/operator/obs-operator.jsonnet
```diff
@@ -1,3 +1,18 @@
-(import '../../components/observatorium.libsonnet') + {
-  config+:: (import 'operator-config.libsonnet'),
-} + (import '../../components/observatorium-configure.libsonnet')
+local config = import 'operator-config.libsonnet';
+local manifests = ((import '../../components/observatorium.libsonnet') + {
+                     config+:: config,
+                   } + (import '../../components/observatorium-configure.libsonnet')).manifests;
+
+local addOwner(k, v) = v { metadata+: { ownerReferences: [{
```
brancz, line 6: we tend to put fields of nested objects on new lines, so something like:

```jsonnet
... v {
  metadata+: {
    ownerReferences: [{
      ...
    }],
  },
}
```

environments/operator/obs-operator.jsonnet
rollandf: std.map works for array.
This is an object here:
`RUNTIME ERROR: std.map second param must be array / string, got object`

brancz: ah ok yes, makes sense :+1:

environments/operator/obs-operator.jsonnet
rollandf: Done

environments/operator/obs-operator.jsonnet
rollandf: Done

environments/operator/obs-operator.jsonnet
```diff
@@ -1,3 +1,21 @@
...
+      controller: true,
+      kind: config.kind,
+      name: config.name,
+      uid: config.uid,
+    }],
+  },
+};
+
+{
+  manifests: std.mapWithKey(addOwner, manifests),
```
brancz, line 20: I meant to inline the function here

rollandf: Done

PR #267, opened 2020-05-04
add up to the dev env deployments and fix config bugs

krasi-georgiev:


components/up.libsonnet
```diff
@@ -64,9 +63,9 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
           '--duration=0',
           '--queries-file=/etc/up/queries.yaml',
           '--log.level=debug',
-        ],
+        ]
         +if std.objectHas(up.config, 'readEndpoint') then
-          ['--endpoint-read=' + up.config.readEndpoint] else [],
+          ['--endpoint-read=' + up.config.readEndpoint] else []
         +if std.objectHas(up.config, 'writeEndpoint') then
           ['--endpoint-write=' + up.config.writeEndpoint] else []
```
krasi-georgiev, line 70: There is a bug here and I am not sure why the `endpoint-write` doesn't get added [here](https://github.com/observatorium/configuration/pull/267/files#diff-e7c54c7f12b71107af51bc6dd182e74eR28)?

krasi-georgiev:
@brancz if you don't think we should have this in the dev env than I would need to refactor to add some tests to at least make sure that the jsonnet for the up  is valid.

components/up.libsonnet
```diff
@@ -109,6 +104,48 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     },
   },
 
+  withRead:: {
```
kakkoyun, line 107:
```suggestion
  withReadEndpoint:: {
```

components/up.libsonnet
```diff
@@ -109,6 +104,48 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+              } else c
+              for c in super.containers
+            ],
+          },
+        },
+      },
+    },
+  },
+
+  withWrite:: {
```
kakkoyun, line 128: How about following just to be consistent
```suggestion
  withWriteEndpoint:: {
```

environments/dev/main.jsonnet
```diff
@@ -6,5 +6,31 @@ local minio = (import '../../components/minio.libsonnet') + {
   },
 };
 
+
+local up = (import '../../components/up.libsonnet') + {
```
kakkoyun, line 10:
```suggestion
local up = (import '../../components/up.libsonnet');
local up = up + up.withReadEndpoint + up.withWriteEndpoint + {
```

components/up.libsonnet
krasi-georgiev: done

components/up.libsonnet
krasi-georgiev: done

components/up.libsonnet
```diff
@@ -109,6 +86,93 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     },
   },
 
+  withReadEndpoint:: {
+    local u = self,
+    readEndpoint:: error 'must provide read endpoint',
```
kakkoyun, line 91: Following our existing config pattern, I believe this should be in the config.

components/up.libsonnet
```diff
@@ -109,6 +86,93 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+            ],
+          },
+        },
+      },
+    },
+  },
+
+  withWriteEndpoint:: {
+    local u = self,
+    writeEndpoint:: error 'must provide write endpoint',
```
kakkoyun, line 111: ditto.

components/up.libsonnet
```diff
@@ -109,6 +86,93 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+            ],
+          },
+        },
+      },
+    },
+  },
+
+  withQuery:: {
+    local u = self,
+    queryConfig:: error 'must provide query config endpoint',
```
kakkoyun, line 131: ditto.

environments/dev/main.jsonnet
```diff
@@ -6,5 +6,24 @@ local minio = (import '../../components/minio.libsonnet') + {
...
+    version: obs.config.up.version,
+    image: obs.config.up.image,
+  },
+};
+
 obs.manifests +
-minio.manifests
+minio.manifests +
+(up +
+ up.withReadEndpoint {
```
kakkoyun, line 24: 👍

components/up.libsonnet
krasi-georgiev: Yeah I see that, but is it not better to set these only when used? otherwise you can easily forget to remove it when you don't need them. With this approach it is explicit and goes hand in hand with the actual mixin.

PR #268, opened 2020-05-07
Set labels properly on the sample CR

ahadas:
The following keys in the metadata are currently ignored:
app.kubernetes.io/component
app.kubernetes.io/instance
app.kubernetes.io/name
app.kubernetes.io/part-of
this PR moves them to metadata.labels


rollandf:
These are labels.
This file is generated, you should fix in `configuration/example/main.jsonnet`
Then run `build.sh`

ahadas:
> These are labels.
> This file is generated, you should fix in `configuration/example/main.jsonnet`
> Then run `build.sh`

done

rollandf:
:+1:

PR #269, opened 2020-05-07
Support for operator lifecycle manager (OLM)

ahadas:
This adds an additional container that serves the registry for OLM operator to be able to deploy/upgrade the observatorium operator.

The output of `operator-sdk generate csv` is saved in `deploy/olm-catalog`, the default output path of the operator-sdk. It is saved in the source control for two reasons:
1. To enable the deployment of a later version - OLM requires the complete upgrade chain in order to deploy a non-initial version (that is expected to change in Openshift 4.6).
2. To be able to create a variant of this version without having access to the operator-sdk - the csv in this repo can rather be used as a template.


deploy/olm/operator_catalogsource.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: observatorium-operator-catalogsource
+  namespace: openshift-marketplace
+spec:
+  displayName: Openshift Performance Addon Operator
```
rollandf, line 7: Leftovers

ahadas: done

deploy/olm-catalog/observatorium-operator/0.0.1/observatorium-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,114 @@
+apiVersion: operators.coreos.com/v1alpha1
```
squat, line 1: This file (and all of the files in /deploy/olm-catalog) are not actually deployed to the cluster, they are only files used to build the docker container. I think they should _not_ live in /deploy because that will cause confusion around deploying OLM
How are these files generated? If they are the output of a generator binary, then that binary should simply be run in the dockerfile

deploy/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM quay.io/openshift/origin-operator-registry:latest
```
squat, line 1: I don't think that deploy/Dockerfile is the right home for this. IMO this location communicates that it is required for deployments, which this is not

deploy/olm-catalog/observatorium-operator/0.0.1/observatorium-operator.v0.0.1.clusterserviceversion.yaml
ahadas: Right, the files in /deploy/olm-catalog are not deployed to the cluster, only those in /deploy/olm are.
These files were generated using the operator-sdk on a separate project as we were not sure this project follows the structure expected by the operator-sdk (as far as we understand, this operator was not generated by the operator-sdk). We placed those files here only because we saw this convention [elsewhere](https://github.com/openshift-kni/performance-addon-operators/tree/master/deploy/olm-catalog) but this is just a temporary state as the next step would be to try generating these files here - and doing that inside the container is a nice idea.

deploy/Dockerfile
ahadas: Fair enough, this is also something we saw in other projects, please advice where to place it

deploy/olm-catalog/observatorium-operator/0.0.1/observatorium-operator.v0.0.1.clusterserviceversion.yaml
squat: great, let's do that then, let's generate everything in a multi-stage container so we can leave this clean :)

deploy/Dockerfile
squat: hmm this will likely have to go in the root and be renamed to `Dockerfile.olm` or something of the sort

deploy/olm-catalog/observatorium-operator/0.0.1/observatorium-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -108,6 +188,9 @@ spec:
   - observatorium
   - prometheus
   - thanos
+  links:
+  - name: Source Code
+    url: https://github.com/observatorium/configurations
```
danielerez, line 193: remove redundant 's' postfix...

deploy/Dockerfile
ahadas: done

deploy/olm-catalog/observatorium-operator/0.0.1/observatorium-operator.v0.0.1.clusterserviceversion.yaml
ahadas: done

deploy/cluster_role_binding.yaml
```diff
@@ -14,4 +15,21 @@ subjects:
 - kind: ServiceAccount
   name: default
   namespace: default
+---
```
squat, line 18: I think this should be it's own file. This file is called `cluster_role_binding.yaml`, so we should make a new one called `role_binding.yaml`, or rename the file

deploy/olm-catalog/observatorium-operator/0.0.1/core.observatorium.io_observatoria.yaml
```diff
@@ -0,0 +1,803 @@
+---
```
squat, line 1: hmm, do we really need yet another copy of this file in this repo? we have one already at https://github.com/observatorium/configuration/blob/master/deploy/crds/core.observatorium.io_observatoria.yaml

deploy/olm-catalog/observatorium-operator/0.0.1/observatorium-operator.v0.0.1.clusterserviceversion.yaml
squat: Are we still working on removing this from the repo?

ahadas: This is indeed not pretty but apparently we cannot avoid this until image indexing is supported by the OLM catalog. It goes like that:
* Until OpenShift 4.5 where image indexing will be supported, we have to store the history of the versions for upgrades flow. For instance, let's say we are going to release 0.0.3 which will be set to replace 0.0.2, then we'll have to deploy 0.0.1 (that is replaced by 0.0.2) and 0.0.2 as well  - without storing them in the repo, we won't have access to their CSV and CRD resources. So the olm-catalog folder needs to remain, we can move it elsewhere if needed though (that's the default place the operator-sdk uses as well as in KubeVirt and Performance-Addon-Operator).
* As from OpenShift 4.5, when image indexing is supported, we won't need each release to provide all those resources for the upgrade chain.

So we containerized the generation of the CSV, that pulls the operator-sdk, builds the utilities and generates the resources. That container is used by a dedicated target within the Makefile for now, later (with image indexing) we can unify the Dockerfiles Dockerfile.csv and Dockerfile.olm into a multi-stage container.

@squat  Would that make sense?

deploy/cluster_role_binding.yaml
ahadas: We tried, but we didn't manage to make the operator-sdk taking the resources from both files

deploy/olm-catalog/observatorium-operator/0.0.1/core.observatorium.io_observatoria.yaml
ahadas: Unfortunately we need, for upgrades - please see [our comment on a previous thread](https://github.com/observatorium/configuration/pull/269#discussion_r425150683)

deploy/olm-catalog/observatorium-operator/0.0.1/observatorium-operator.v0.0.1.clusterserviceversion.yaml
squat: hmm no, I don't understand that. Why don't we simply build the CSV and bake it into the OLM Docker image (rather than a CSV Docker image) without it ever touching the repo?

ahadas: We can do that now, when we're releasing the first OLM based version, let's call it 0.0.1. We'll face an issue when we'll reach the point we release the next version - the next version, let's say 0.0.2, would be set to replace 0.0.1, right? without placing the CRD and CSV of 0.0.1 in the registry container, users will not be able to deploy 0.0.2 (OLM currently assumes the registry container hold the information that is needed for the whole upgrade chain). So if we don't place it on the repo, how would we have access to the CRD and CSV of 0.0.1, when we'll reach 0.0.2?

danielerez:
@brancz @metalmatze @kakkoyun @squat Can you please take a look?

deploy/cluster_role_binding.yaml
squat: That's too bad. I don't quite understand the limitation :/

operator/hack/csv-generate.sh
```diff
@@ -0,0 +1,84 @@
...
+fi
+if [ -n "$MAINTAINERS_FILE" ]; then
+	MAINTAINERS="-maintainers-from=$MAINTAINERS_FILE"
+fi
+if [ -n "$ANNOTATIONS_FILE" ]; then
+	EXTRA_ANNOTATIONS="-annotations-from=$ANNOTATIONS_FILE"
+fi
+
+clean_package() {
+#	mkdir -p "$CSV_DIR"
```
squat, line 42: Do we need this?

operator/hack/csv-generate.sh
```diff
@@ -0,0 +1,84 @@
...
+	rm -rf "$OUT_DIR"
+	mkdir -p "$OUT_CSV_DIR"
+}
+
+if ! [[ "$CSV_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
squat, line 47: This is not a correct server regex, consider taking one from the semver website

operator/hack/csv-generate.sh
```diff
@@ -0,0 +1,84 @@
...
+	DESCRIPTION="-description-from=$DESCRIPTION_FILE"
+fi
+if [ -n "$MAINTAINERS_FILE" ]; then
+	MAINTAINERS="-maintainers-from=$MAINTAINERS_FILE"
+fi
+if [ -n "$ANNOTATIONS_FILE" ]; then
+	EXTRA_ANNOTATIONS="-annotations-from=$ANNOTATIONS_FILE"
+fi
+
+clean_package() {
```
squat, line 41: This function is only used once; why not inline it?

deploy/csv/maintainers.json
```diff
@@ -0,0 +1,3 @@
+{
+    "Observatorium Project": ""
```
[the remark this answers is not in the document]
danielerez: It should be a contact email for the project. Do we have one that can be used? :)

deploy/csv/annotations.json
```diff
@@ -0,0 +1,5 @@
+{
+    "categories": "OpenShift Optional",
+    "description": "Observatorium Operator provides the ability to install the components that comprise the Observatorium project.",
+    "repository": "https://github.com/observatorium/configuration"
+}
```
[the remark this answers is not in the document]
danielerez: sure, done.

operator/hack/csv-generate.sh
danielerez: right, will use this then: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

operator/hack/csv-generate.sh
danielerez: could be helpful for future reusability, but we can indeed inline it for now.

operator/hack/csv-generate.sh
danielerez: removed :)

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	if *maintainersFile != "" {
+		userData.Maintainers = readKeyValueMapFromFileOrPanic(*maintainersFile)
+	}
+	if *descriptionFile != "" {
+		userData.Description = string(readFileOrPanic(*descriptionFile))
+	}
+
+	// start with a fresh output directory if it already exists
+	os.RemoveAll(*outputDir)
```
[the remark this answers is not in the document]
danielerez: done

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/blang/semver"
+
```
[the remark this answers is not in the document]
danielerez: done

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+	skipRange           = flag.String("skip-range", "", "the CSV version skip range")
+	operatorCSVTemplate = flag.String("operator-csv-template-file", "", "path to csv template example")
+
+	operatorImage = flag.String("operator-image", "", "operator container image")
+
+	inputManifestsDir = flag.String("manifests-directory", "", "The directory containing the extra manifests to be included in the registry bundle")
+
+	outputDir = flag.String("olm-bundle-directory", "", "The directory to output the unified CSV and CRDs to")
+
+	annotationsFile = flag.String("annotations-from", "", "add metadata annotations from given file")
```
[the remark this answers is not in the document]
danielerez: done

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+
+type csvUserData struct {
+	Description      string
+	ExtraAnnotations map[string]string
+	Maintainers      map[string]string
+}
+
+func generateUnifiedCSV(userData csvUserData) {
+
+	operatorCSV := UnmarshalCSV(*operatorCSVTemplate)
```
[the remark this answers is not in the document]
danielerez: just since we extracted them to csvtools

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+type csvUserData struct {
+	Description      string
+	ExtraAnnotations map[string]string
+	Maintainers      map[string]string
+}
+
+func generateUnifiedCSV(userData csvUserData) {
+
```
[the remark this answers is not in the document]
danielerez: done

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+	outputDir = flag.String("olm-bundle-directory", "", "The directory to output the unified CSV and CRDs to")
+
+	annotationsFile = flag.String("annotations-from", "", "add metadata annotations from given file")
+	maintainersFile = flag.String("maintainers-from", "", "add maintainers list from given file")
+	descriptionFile = flag.String("description-from", "", "replace the description with the content of the given file")
+
+	semverVersion *semver.Version
+)
+
+func finalizedCsvFilename() string {
```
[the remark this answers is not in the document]
danielerez: done

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+	csvVersion          = flag.String("csv-version", "", "the unified CSV version")
+	replacesCsvVersion  = flag.String("replaces-csv-version", "", "the unified CSV version this new CSV will replace")
+	skipRange           = flag.String("skip-range", "", "the CSV version skip range")
+	operatorCSVTemplate = flag.String("operator-csv-template-file", "", "path to csv template example")
+
+	operatorImage = flag.String("operator-image", "", "operator container image")
+
+	inputManifestsDir = flag.String("manifests-directory", "", "The directory containing the extra manifests to be included in the registry bundle")
+
+	outputDir = flag.String("olm-bundle-directory", "", "The directory to output the unified CSV and CRDs to")
```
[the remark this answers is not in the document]
danielerez: done

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func generateUnifiedCSV(userData csvUserData) {
+
+	operatorCSV := UnmarshalCSV(*operatorCSVTemplate)
+	strategySpec := operatorCSV.Spec.InstallStrategy.StrategySpec
+
+	// this forces us to update this logic if another deployment is introduced.
+	if len(strategySpec.DeploymentSpecs) != 1 {
+		panic(fmt.Errorf("expected 1 deployment, found %d", len(strategySpec.DeploymentSpecs)))
```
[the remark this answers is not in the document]
danielerez: We basically took the following as a reference: https://github.com/openshift-kni/performance-addon-operators/blob/master/tools/csv-generator/csv-generator.go
But sure, will remove 'panic' here.

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	// Set correct csv versions and name
+	v := version.OperatorVersion{Version: *semverVersion}
+	operatorCSV.Spec.Version = v
+	operatorCSV.Name = "observatorium-operator.v" + *csvVersion
+	if *replacesCsvVersion != "" {
+		operatorCSV.Spec.Replaces = "observatorium-operator.v" + *replacesCsvVersion
+	} else {
+		operatorCSV.Spec.Replaces = ""
```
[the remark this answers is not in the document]
danielerez: right, it might have a wrong value from the csv taken as a template.

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,228 @@
...
+	var err error
+	// Set correct csv versions and name
+	semverVersion, err = semver.New(*csvVersion)
+	if err != nil {
+		panic(err)
+	}
+
+	userData := csvUserData{
+		Description: `
+Observatorium Operator provides the ability to install the components that comprise the Observatorium project.`,
```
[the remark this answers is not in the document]
danielerez: ok, it can always be taken from userData

danielerez:
@squat @brancz @metalmatze @kakkoyun Can you please take another look?

deploy/csv/maintainers.json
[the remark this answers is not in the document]
squat: team-observability-platform@redhat.com

operator/csv-generator/csv-generator.go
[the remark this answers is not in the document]
squat: can we unexport them? they can't be imported from other packages anyways and helps keep this code consistent and clean

operator/csv-generator/csvtools.go
```diff
@@ -0,0 +1,88 @@
...
+	yaml "github.com/ghodss/yaml"
+	csvv1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// UnmarshalCSV decodes a YAML file, by path, and returns a CSV
+func UnmarshalCSV(filePath string) *csvv1.ClusterServiceVersion {
+	bytes, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		panic(err)
```
squat, line 18: there are lots of valid reasons why reading a file might fail, which are not system failure warranting a panic. I think this func's signature should include returning errors. Generally, let's follow our best practices and avoid panicking (without handling panics) when the cause is not certainly a system error

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,224 @@
...
+		log.Fatal("--operator-image is required")
+	} else if *outputDir == "" {
+		log.Fatal("--olm-bundle-directory is required")
+	}
+
+	var err error
+	// Set correct csv versions and name
+	semverVersion, err = semver.New(*csvVersion)
+	if err != nil {
+		panic(err)
```
squat, line 193: this should not panic. If someone happens to pass an invalid tag to the program, this is user error and should be handled appropriately

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,224 @@
...
+)
+
+func finalizedCSVFilename() string {
+	return "observatorium-operator.v" + *csvVersion + ".clusterserviceversion.yaml"
+}
+
+func copyFile(src string, dst string) {
+	srcFile, err := os.Open(src)
+	if err != nil {
+		panic(err)
```
squat, line 45: again, all of these can be valid os errors and should not cause panics, but should instead be correctly handled

danielerez: removed all panics :)

operator/csv-generator/csv-generator.go
danielerez: removed all panics :)

operator/csv-generator/csvtools.go
danielerez: removed all panics :)

deploy/olm-catalog/observatorium-operator/0.0.1/core.observatorium.io_observatoria.yaml
ahadas: We can avoid this if we say that we don't care about upgrades from previous versions (but then why would you want OLM, right?) or wait for image indexing to be introduced which is AFAIU planned for Openshift 4.6. If we want to ship this before image indexing is introduced and support upgrades, I don't really see how we can easily (without checking out older branches) avoid adding that to the repo, that's what other release operators we checked do in order to have access to older versions' CRDs during the creation of the release container...

ahadas: @squat but if you have a suggestion for how to do it, we'd love to hear - it's not like we like this solution...

squat: The first solution I imagine is to use multistage dockerfiles for all of the OLM containers and for the 0.2.0 container to have a stage that copies a file from the 0.1.0 container, ie the manifest from the previous release. Then on the next release, eg 0.3.0, we need to swap out the FROM to point to 0.2.0. This way we always keep the manifests in the containers as build artifacts and they have the benefit of being inherently tagged and generated without needed to be checked into version control.

ahadas: Ah I see, you propose to retrieve the artifacts for older versions from the image of the most recent release container. That could work but I see some possible issues with that:
1. Assuming we want to base the latest release container on the latest version of openshift/origin-operator-registry, it means that we'll need to download two images that are no tiny rather than just one.
2. Assuming we want to use the latest (or a recent) version of the operator-sdk, we would need to provide a way to access it during the generation of downstream release.

I wouldn't suggest to do that unless there is a good reason not to store few more textual files in the repo.

squat: Hmm 1 doesn't sound like a problem. Docker will download two images during building, which is a one-time cost in CI, but not at run time, so this can be ignored IMO, this is what build containers are designed for. 
2 also doesn't sound like a limitation to me, but I understand if this is complexity that you don't want to undertake

rollandf: I think that having the older versions easily available and human readable is valuable.
Someone could find more easily in which version a specific field was added to the CRD for example.

Also, for example, see that all the versions are available:

https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/prometheus

https://github.com/kubevirt/hyperconverged-cluster-operator/tree/master/deploy/olm-catalog/kubevirt-hyperconverged

deploy/csv/annotations.json
```diff
@@ -0,0 +1,5 @@
+{
```
squat, line 1: We did away entirely with the `/deploy` directory since it was confusing; the entire repository contains different deployment configurations and the `/deploy` directory simply held one such configuration for the operator. This directory doesn't exist any more and I think that the OLM stuff should either be nested in the `/operator` directory or in a new top level directory called `/olm`

deploy/csv/annotations.json
```diff
@@ -0,0 +1,5 @@
...
+    "categories": "OpenShift Optional",
+    "description": "Observatorium Operator provides the ability to install the components that comprise the Observatorium project.",
+    "repository": "https://github.com/observatorium/configuration"
```
squat, line 4: we will need to update these URLs since the repo is now named https://github.com/observatorium/deployments

operator/csv-generator/csv-generator.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	// Set api maturity
+	operatorCSV.Spec.Maturity = "alpha"
+
+	// Set links
+	operatorCSV.Spec.Links = []csvv1.AppLink{
+		{
+			Name: "Source Code",
+			URL:  "https://github.com/observatorium/configuration",
```
squat, line 96: same here

doc/operator/deploy-operator.md
```diff
@@ -48,12 +48,20 @@ kubectl apply -f https://raw.githubusercontent.com/observatorium/deployments/mas
...
+In case you need to force a new image download (e.g. development environment), please refer to the [development section](#Development).
 ```shell script
 kubectl apply -f https://raw.githubusercontent.com/observatorium/deployments/master/operator/manifests/crds/core.observatorium.io_observatoria.yaml
 kubectl apply -f https://raw.githubusercontent.com/observatorium/deployments/master/operator/manifests/operator.yaml
 ```
 
+#### Deploy via Operator Lifecycle Manager (OLM) - OpenShift Only
+Alternatively, the operator can be deployed and managed by OLM.
+```shell script
+oc apply -f https://raw.githubusercontent.com/observatorium/configuration/master/deploy/olm/operator_operatorgroup.yaml
```
squat, line 60: same for these

metalmatze:
The Operator was moved out of this repository to https://github.com/observatorium/operator. Probably best to open a new PR (continued) over there.

PR #271, opened 2020-05-12
Upate Observatorium API to get latest improvements

metalmatze:
This update aims to simplify Observatorium APIs configuration surface.

Breaking changes:

 * Observatorium API Thanos Querier serving only UI assets is obsolete
 * `--metrics.ui.endpoint` was removed
* `--metrics.read.endpoint` takes the root path and appends `/api/v1/{query/query_range}`
* `--metrics.write.endpoint` takes the root path and appends `/api/v1/receive`
* Updates from Observatorium API configuration (Port for internal server, healthcheck improvements)

As I was touching all manifests and its jsonnet I took some minutes to add some proper Makefile targets that will only rebuild, if dependent files change.

/cc @squat @brancz @kakkoyun @krasi-georgiev

As we cover the Operator in e2e tests and I actually tried it on a local kind cluster, I don't expect any breaking changes from this for the Operator itself. /cc @rollandf @nmagnezi


rollandf:
Nice additions to the Makefile!

build_deploy.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
-# This is a placeholder file needed by the CI/CD templates
-# You can safely ignore this file :)
-exit 0
```
krasi-georgiev, line 4: are you sure this should be removed? last time I talked to @aditya-konarde he mentioned that this is needed by the CI.

squat: grrr this is true

Issue #272, opened 2020-05-13
The Operator should support TLS for observatorium-api


nmagnezi:
The operator should allow passing a secret in CR, to be used in observatorium-api pod args.

nmagnezi:
I do not have permissions to assign a myself, but I'm working on that.

squat:
closed in https://github.com/observatorium/deployments/pull/329

Issue #57, opened 2019-09-09
Setup Jaeger backend


jpkrohling:
Setup a simple Jaeger backend, made up of:

- Jaeger All in One with Badger (cc @burmanm). Once we figure out how HA can be achieved in our setup, we can split the deployments: collector in read-write mode, query in read-only mode.
- Jaeger Service, exposing the UI port (16686) and the collector's gRPC (14250).

@aditya-konarde: how can we proceed with adding the Jaeger Agent as a sidecar to the Thanos pods? Should we start with the all-in-one+service first, and the sidecar in a follow-up issue? Or is there a deployment where I could already add the sidecar as part of the initial PR?

cc @objectiser

jpkrohling:
For those wondering: operators are a no-go in the current setup, but the idea is to move to it in the future.

jpkrohling:
> However we still need to add the agent sidecars to Thanos

I can work on that in the coming week(s), if you can wait. Otherwise, here's the relevant commit I've shown during the presentation: https://github.com/jpkrohling/configuration/commit/c0f0faa1d855b6880365f71953a50a78c667e66c

Apart from adding a [sidecar](https://github.com/jaegertracing/jaeger-operator/blob/bad90a64177d062f7be973d73de00d404b92b95f/pkg/inject/sidecar.go#L153-L184) to the affected components, the `traceConfig` object would have to be changed, to remove the `endpoint` property, as the default client configuration will try to connect to the Agent on `localhost:6831/udp`.

kakkoyun:
Completed by #105

PR #273, opened 2020-05-14
Adds Operator support  for TLS enabled observatorium-api 

nmagnezi:
Issue: #272 

Depends on:
* [observatorium/issues/54](https://github.com/observatorium/observatorium/issues/54)
* [observatorium/issues/57](https://github.com/observatorium/observatorium/issues/57)
* [observatorium/issues/60](https://github.com/observatorium/observatorium/issues/60)


components/observatorium-api-tls-secret.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+  // WARNING: Never use this example secret in production!
+  secret: {
+    apiVersion: 'v1',
+    kind: 'Secret',
+    metadata: {
+      name: 'observatorium-api-tls-secret',
+      namespace: api_tls_secret.config.namespace,
+    },
+    data: {
+      'ca.pem': 'LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJkVENDQVJxZ0F3SUJBZ0lVRUNEUzQ3Z3ZoUEFWSm5vNTJlOFluemhONlZRd0NnWUlLb1pJemowRUF3SXcKR0RFV01CUUdBMVVFQXhNTmIySnpaWEoyWVhSdmNtbDFiVEFlRncweU1EQTBNVFl4TVRRek1EQmFGdzB5TlRBMApNVFV4TVRRek1EQmFNQmd4RmpBVUJnTlZCQU1URFc5aWMyVnlkbUYwYjNKcGRXMHdXVEFUQmdjcWhrak9QUUlCCkJnZ3Foa2pPUFFNQkJ3TkNBQVQwWDJxR2Y4Z0FKTlRJeWFDc290QXVoY3BweTBIa1NuY0JmdG1aaWpqWUNsZW0KN0JPd2I1a1RScnhaeVJxUU9Kdll0RmV0R1E3T2RjdUJOaVpXb05vem8wSXdRREFPQmdOVkhROEJBZjhFQkFNQwpBUVl3RHdZRFZSMFRBUUgvQkFVd0F3RUIvekFkQmdOVkhRNEVGZ1FVMHUzTm83YnNFekxtTzNYd0h2OUN3S21EClVNY3dDZ1lJS29aSXpqMEVBd0lEU1FBd1JnSWhBUFFsbS8wQWdoSkJEYmxtZHdkRE81V2xEbHZiUFovaE1yc2IKMFh3SHF0aHFBaUVBcXAzaTgvYmJyMnVFZzlYS1RMbE16Sit2OFdHSFIyc0wwYVNqUk9zeG9tUT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=',
```
brancz, line 16: we should pass these in as parameters in the `config`, and you can use `stringData` so this doesn't have to be base64 encoded

brancz, line 16: we should pass these in as parameters in the config, and you can use stringData so this doesn't have to be base64 encoded

nmagnezi: this is now passed as param to a generated manifest.
I still use base64, but it works well and the mount shows decoded files

nmagnezi:
@brancz I'm waiting for [PR-298](https://github.com/observatorium/configuration/pull/289) to get merged so I can use the latest observatorium-api image and remove the readiness probs workarounds.

Please have a look and provide feedback when you can, thank you 😃

operator/api/v1alpha1/observatorium_types.go
```diff
@@ -158,6 +158,8 @@ type APISpec struct {
 	Image string `json:"image,omitempty"`
 	// Version describes the version of API to use.
 	Version string `json:"version,omitempty"`
+	// TLS Secret Configuration
+	Secret string `json:"secret,omitempty"`
```
squat, line 162: This should probably be named `TLSSecretName` or something more descriptive. Also, I would rather propose making this a struct called `TLS` with a sub field called `SecretName`, `CAField`, `CertificateField` etc otherwise it is super hard to communicate to the user how to structure their secret correctly. It's much easier to ask the user to tell us how to structure it IMO

components/observatorium.libsonnet
```diff
@@ -207,7 +207,57 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+                },
+              },
+            ],
+          },
+        },
+      },
+    },
+  },
+
+  api:: (import 'observatorium/observatorium-api.libsonnet') + obs.apiWithTLS {
```
squat, line 260: Shouldn't this be the basic component without TLS? TLS should be added with the mixin in an environment, eg base

components/up-tls-secret.libsonnet
```diff
@@ -0,0 +1,25 @@
+{
```
squat, line 1: Why is this a separate file and not part of the observatorium up component?

environments/dev/main.jsonnet
```diff
@@ -87,6 +87,17 @@ local obs = (import '../base/observatorium.jsonnet') + {
   },
 };
 
+local tls_secret = (import '../../components/observatorium-api-tls-secret.libsonnet') + {
```
squat, line 90: Rather than being a separate file that need to be imported, the TLS secret should be automatically appended to the observatorium manifests when you add the withTLS mixin to the API. There should be no need to explicitly add another set of manifests or input another libsonnet file

components/observatorium.libsonnet
nmagnezi: Done

environments/dev/main.jsonnet
nmagnezi: done

operator/api/v1alpha1/observatorium_types.go
nmagnezi: Done

components/up-tls-secret.libsonnet
nmagnezi: Done

nmagnezi:
Hi @metalmatze, @squat 
The [PR api-deployment manifest](https://github.com/observatorium/deployments/blob/f41a7e4d2000008ca5d0aacf1bbed96077b53129/environments/dev/manifests/api-deployment.yaml#L45-L52) looks on-par with observatorium [MTLS manifest example](https://github.com/observatorium/observatorium/blob/4420cccb4899129205b1782cf99b94aef9dd7537/examples/manifests/deployment-with-mtls.yaml#L42-L49) and [integration tests configuration](https://github.com/observatorium/observatorium/blob/4420cccb4899129205b1782cf99b94aef9dd7537/test/integration.sh#L35-L40), I still get the following:
1. Pod not marked as ready
2. Logs indicate an issue. It looks like a response to liveness/readiness probes since no metrics were sent at this point.
```
2020/06/21 06:43:38 http: TLS handshake error from 127.0.0.1:43254: remote error: tls: bad certificate
2020/06/21 06:43:40 http: TLS handshake error from 127.0.0.1:43274: remote error: tls: bad certificate
2020/06/21 06:43:43 http: TLS handshake error from 127.0.0.1:43306: remote error: tls: bad certificate
2020/06/21 06:43:48 http: TLS handshake error from 127.0.0.1:43356: remote error: tls: bad certificate
```

Since observatorium integration tests are not running under k8s, is it possible that [observatorium/issues/60]( https://github.com/observatorium/observatorium/issues/60) persists?

environments/base/default-config.libsonnet
```diff
@@ -121,10 +121,36 @@
...
+
+  withMTLS: {
+    local apiWithMTLS = self,
+    caPem: 'ca.pem',
+    volumeMounts: {
+      name: 'observatorium-api-tls-client-ca',
+      mountPath: '/mnt/clientca',
+      readOnly: true,
+    },
+    clientCAFile: apiWithMTLS.volumeMounts.mountPath + apiWithMTLS.caPem + 'Aaa',
```
squat, line 151: What is 'Aaa'?

environments/dev/main.jsonnet
```diff
@@ -34,6 +34,35 @@ local dex = (import '../../components/dex.libsonnet') + {
...
+  },
+  tls_configmap: {
+    apiVersion: 'v1',
+    kind: 'ConfigMap',
+    metadata: {
+      name: 'observatorium-api-tls-client-ca',
+      namespace: obs.config.namespace,
+    },
+    data: {
+      'ca.pem': importstr '../../tmp/certs/up_client/ca.pem',
```
squat, line 63: The reason that mTLS is failing is that you have generated two different sets of client certs each with a different CA. The client certs for the up binary were generated using this certificate, so using this CA as the client-ca means up will work. *But* the client certs for the healthchecks were generated using the observatorium/ca.pem cert, so when the API checks the client certificates against the up_client/ca.pem cert, the validation will fail

environments/dev/main.jsonnet
```diff
@@ -34,6 +34,35 @@ local dex = (import '../../components/dex.libsonnet') + {
...
 local obs = (import '../base/observatorium.jsonnet') + {
+  tls_secret: {
+    apiVersion: 'v1',
+    kind: 'Secret',
+    metadata: {
+      name: 'observatorium-api-tls-certs',
+      namespace: obs.config.namespace,
+    },
+    data: {
+      'server.key': importstr '../../tmp/base64_certs/observatorium/server.key',
```
squat, line 45: Let's keep this simple and do the base64 encoding inside of jsonnet, that way we don't need the extra steps in the makefile

example/main.jsonnet
```diff
@@ -37,118 +37,149 @@ local obs = (import '../environments/base/observatorium.jsonnet');
 
 {
   local cr = self,
-  name:: 'observatorium-cr',
+  cr_no_tls_name:: 'observatorium-cr',
```
squat, line 40: Let's be consistent and use camelCase for everything

environments/dev/main.jsonnet
nmagnezi: I have tested this as well, and Done per your request. Yet I still get the`TLS handshake error`

environments/base/default-config.libsonnet
nmagnezi: Done

example/main.jsonnet
nmagnezi: Done

environments/dev/main.jsonnet
nmagnezi: Done

brancz:
Looks like this needs a rebase and CI is failing. Once green we can have another look.

nmagnezi:
> Looks like this needs a rebase and CI is failing. Once green we can have another look.

@brancz rebase done. it will fail CI for the same reason I mentioned [here](https://github.com/observatorium/deployments/pull/273#discussion_r443194527)

squat:
This should not fail for the same reason since the healthchecks no longer use mTLS. e.g. these flags no longer exist: https://github.com/observatorium/deployments/pull/273/files#diff-600fb07f5f3b3600a9a1111605f807ceR48
I think we need to fix the rebase actually

squat:
let me know if you need help :)

operator/api/v1alpha1/observatorium_types.go
```diff
@@ -178,11 +178,35 @@ type APITenant struct {
...
+	Name           string `json:"name"`
+	PrivateKeyFile string `json:"privateKeyFile"`
+	ReloadInterval string `json:"reloadInterval"`
+}
+
+type TLS struct {
+	TLSSecret TLSSecret `json:"secret"`
+}
+
+type MTLSConfigMap struct {
```
squat, line 192: We don't have any Observatorium-wide mTLS settings (anymore), so we should remove this struct

operator/api/v1alpha1/observatorium_types.go
```diff
@@ -178,11 +178,35 @@ type APITenant struct {
 	OIDC TenantOIDC `json:"oidc"`
 }
 
+type TLSSecret struct {
+	CertFile       string `json:"certFile"`
+	Name           string `json:"name"`
+	PrivateKeyFile string `json:"privateKeyFile"`
+	ReloadInterval string `json:"reloadInterval"`
```
squat, line 185: We will need an optional field to load the CA that signed the certificate. This is optional because the healthchecker needs to know that the certificate is valid and if the CA is not in the system certs then it needs to be specified. Note, this is different from an mTLS CA used to check clients of the API; this is for clients to validate the server's cert

components/up-job.libsonnet
```diff
@@ -38,6 +38,9 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
           '--latency=10s',
           '--initial-query-delay=5s',
           '--threshold=0.90',
+          '--tls-ca-file=' + up.config.withTLS.clientCAFile,
```
squat, line 41: This should really be called `serverCAFile` because it's the CA that signed the _server's_ certificate that matters here, not the client's

environments/dev/main.jsonnet
```diff
@@ -34,14 +34,41 @@ local dex = (import '../../components/dex.libsonnet') + {
 };
 
 local obs = (import '../base/observatorium.jsonnet') + {
+  tls_secret: {
```
squat, line 37: Let's keep everything here camelCase for consistency

environments/base/default-config.libsonnet
```diff
@@ -133,6 +133,32 @@
...
+      mountPath: '/mnt/certs',
+      readOnly: true,
+    },
+    certFile: apiWithTLS.volumeMounts.mountPath + apiWithTLS.serverPem,
+    privateKeyFile: apiWithTLS.volumeMounts.mountPath + apiWithTLS.serverKey,
+    reloadInterval: '1m',
+  },
+
+  withMTLS: {
+    local apiWithMTLS = self,
```
squat, line 152: We should remove this because the API doesn't support any process-wide mTLS

components/observatorium-configure.libsonnet
```diff
@@ -43,6 +43,14 @@
     config+:: obs.config.api,
   },
 
+  withTLS+:: {
+    config+:: obs.config.withTLS,
+  },
+
+  withMTLS+:: {
```
squat, line 50: Same here: there is no longer any API-wide mTLS settings

environments/dev/manifests/api-deployment.yaml
```diff
@@ -42,6 +42,14 @@ spec:
         - --log.level=warn
         - --rbac.config=/etc/observatorium/rbac.yaml
         - --tenants.config=/etc/observatorium/tenants.yaml
+        - --web.healthchecks.url=https://127.0.0.1:8080
+        - --tls.server.cert-file=/mnt/certs/server.pem
+        - --tls.server.key-file=/mnt/certs/server.key
+        - --tls.healthchecks.cert-file=/mnt/certs/client.pem
```
squat, line 48: Several of these flags no longer exist.

environments/dev/manifests/api-deployment.yaml
```diff
@@ -42,6 +42,14 @@ spec:
...
+        - --tls.healthchecks.key-file=/mnt/certs/client.key
```
squat, line 49: This one too

environments/dev/manifests/api-deployment.yaml
```diff
@@ -42,6 +42,14 @@ spec:
...
+        - --tls.healthchecks.server-ca-file=/mnt/certs/ca.pem
+        - --tls.reload-interval=1m
+        - --tls.server.client-ca-file=/mnt/clientca/ca.pem
```
squat, line 52: And this one

squat:
ping @nmagnezi will you be able to take a look at the issues/comments I added a few weeks ago?

Makefile
```diff
@@ -7,11 +7,27 @@ VCS_BRANCH := $(strip $(shell git rev-parse --abbrev-ref HEAD))
...
 DOCKER_REPO ?= quay.io/observatorium/observatorium-operator
 
-BIN_DIR ?= $(shell pwd)/tmp/bin
+TMP_DIR := $(shell pwd)/tmp
+BIN_DIR ?= $(TMP_DIR)/bin
+CERT_DIR ?= $(TMP_DIR)/certs
 
 CONTROLLER_GEN ?= $(BIN_DIR)/controller-gen
 JB ?= $(BIN_DIR)/jb
+ GENERATE_TLS_CERT ?= $(BIN_DIR)/generate-tls-cert
```
squat, line 16: nit: extra space here

Makefile
```diff
@@ -7,11 +7,27 @@ VCS_BRANCH := $(strip $(shell git rev-parse --abbrev-ref HEAD))
...
 
+$(BIN_DIR):
+	mkdir -p $(BIN_DIR)
+
+$(CERT_DIR):
+	mkdir -p $(CERT_DIR)
+
+# Generate TLS certificates for local development.
+generate-cert: $(GENERATE_TLS_CERT) $(CERT_DIR)
```
squat, line 25: We should make the dependency on the cert dir target order-only
```suggestion
generate-cert: $(GENERATE_TLS_CERT) | $(CERT_DIR)
```

squat, line 25: This should be a phony target

squat:
Closed in https://github.com/observatorium/deployments/pull/329

PR #275, opened 2020-05-14
receive: Increase local retention to 15 days

brancz:
@metalmatze


brancz:
in case uploads are not happening for any reason 6h of buffer is a little small to react

squat:
it might not be enough while we don't have alerts for failing uploads. we should keep data on disk for at least the time that uploads aren't succeeding + the time we need to triage/fix the uploads

PR #276, opened 2020-05-14
Remove withXXX helper functions from up component

metalmatze:
This should be easier to configure and the --endpoint-read and --endpoint-write are added as flags, simply by not having `''` in the config.

/cc @kakkoyun @brancz

### DIFF
--- a/components/up.libsonnet
+++ b/components/up.libsonnet
@@ -8,6 +8,14 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     namespace: error 'must provide namespace',
     version: error 'must provide version',
     image: error 'must provide image',
+    // queryConfig: {},
+    readEndpoint: '',
+    writeEndpoint: '',
+    resources: {
+      requests: {},
+      limits: {},
+    },
+    serviceMonitor: false,
 
     commonLabels:: {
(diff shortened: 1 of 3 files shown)


components/up.libsonnet
```diff
@@ -176,31 +115,30 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         name: up.config.name,
         namespace: up.config.namespace,
       },
-    },
-  },
-
-  withServiceMonitor:: {
```
kakkoyun, line 182: I think we should keep `withServiceMonitor`. It is different than other changes, it adds an additional resource.
Also we have used the same pattern for every other component, so I personally prefer this pattern for external resources.

metalmatze: I agree, we simply didn't want to match things. 
The funny thing is, we figured out a way to elegantly solve this too though. Maybe we can still improve it further down the road.

Essentially, if `serviceMonitor == true` then we create the object and add it to the overall object, else this serviceMonitor object is simply `null`.
Then if you look at the manifest generation, we check for object being `null` and then simply ignore them :)

WDYT?

kakkoyun: Fair enough 👍

PR #280, opened 2020-05-20
Enable Thanos Store with caching bucket

kakkoyun:
depends: https://github.com/observatorium/configuration/pull/279

### DIFF
--- a/environments/base/manifests/thanos-store-shard0-statefulSet.yaml
+++ b/environments/base/manifests/thanos-store-shard0-statefulSet.yaml
@@ -6,7 +6,7 @@ metadata:
     app.kubernetes.io/instance: observatorium-xyz
     app.kubernetes.io/name: thanos-store
     app.kubernetes.io/part-of: observatorium
-    app.kubernetes.io/version: v0.12.0-rc.1
+    app.kubernetes.io/version: v0.13.0-rc.0
     store.observatorium.io/shard: shard-0
   name: observatorium-xyz-thanos-store-shard-0
   namespace: observatorium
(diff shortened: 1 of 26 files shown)


environments/base/manifests/thanos-store-shard0-statefulSet.yaml
```diff
@@ -51,6 +51,28 @@ spec:
...
+          "config":
+            "addresses":
+            - "dnssrv+_client._tcp.observatorium-xyz-thanos-store-memcached.observatorium.svc"
+            "dns_provider_update_interval": "10s"
+            "max_async_buffer_size": 100000
+            "max_async_concurrency": 100
+            "max_get_multi_batch_size": 1000
+            "max_get_multi_concurrency": 900
+            "max_idle_connections": 1000
+            "max_item_size": "1MiB"
```
brancz, line 68: Can we ask Marco about their settings here? I can't believe that 1MiB is a very good size for this.

kakkoyun: Yes, I can do that. FYI, these are the defaults from upstream Thanos.

brancz: Yeah we should improve those defaults as well.

kakkoyun: @brancz After some offline conversation and checking the PR again, values seem alright. Caching bucket caches the chunk pieces using `Bucket.GetRange` so the cached pieces are a lot smaller than `max_item_size` limit. And the limit is specified by `chunk_subrange_size: 16000`. So it's good to go. We can do fine-tuning at downstream when needed.

brancz: Got it, thanks for clarifying. All good from my side!

PR #281, opened 2020-05-25
Add loki manifests to observatorium

periklis:
This PR adds basic support to deploy Loki as part of the observatorium. It is inspired by Grafana's production manifest but adapted to the observatorium configuration. It uses per default the `boltdb_shipper`  feature on `minio` to minimize operational cost. 

The PR is work in progress. Missing pieces:
- [x] Add support to override loki's config and overrides per environment
- [x] Add configuration options for resources and replicas per component and environment
- [x] Add service monitors for the distributor and ingester components
- [x] Add boltdb_shipper configuration to write index on minio.
- [x] Add `withMemberlist` for ring configuration based on cortex gossip support
- [x] Add `withEtcd` for a default ring configuration based on etcd deployment
- [x] Add `with{Chunk,IndexQuery,IndexWrite,Results}Cache` cache support based on memcached
- [x] Add Loki E2E tests based on `observatorium/up`

Depends on: 
- observatorium/observatorium#47, observatorium/observatorium#70, observatorium/observatorium#72
- #289, #306
- observatorium/up#16, observatorium/up#18


jsonnetfile.json
```diff
@@ -40,17 +40,17 @@
     {
       "source": {
         "git": {
-          "remote": "https://github.com/observatorium/observatorium",
+          "remote": "https://github.com/periklis/observatorium.git",
```
periklis, line 43: This needs to be reverted to former configuration after deps are merged

environments/base/default-config.libsonnet
```diff
@@ -121,8 +128,10 @@
 
   api: {
     local apiConfig = self,
-    version: 'master-2020-05-11-v0.1.1-35-g3c24d9b',
-    image: 'quay.io/observatorium/observatorium:' + apiConfig.version,
+    // version: 'master-2020-05-11-v0.1.1-35-g3c24d9b',
+    // image: 'quay.io/observatorium/observatorium:' + apiConfig.version,
+    version: 'loki-support-2020-05-25-v0.1.1-36-g72e7af6',
+    image: 'quay.io/periklis/observatorium:' + apiConfig.version,
```
periklis, line 134: This needs to be reverted to former configuration after deps are merged

bwplotka:
cc @squat @metalmatze  (:

squat:
This is failing until we can fix CI by:
1. decreasing replica counts so all the pods fit (temporary fix)
2. deploy on the LLC using the new namespace-provisioner

we are working on getting 2 working this week to unblock this PR :)

environments/dev/manifests/loki-config-map.yaml
```diff
@@ -0,0 +1,146 @@
...
+    "frontend":
+      "compress_responses": true
+      "max_outstanding_per_tenant": 200
+    "frontend_worker":
+      "frontend_address": "observatorium-xyz-loki-query-frontend.observatorium.svc.cluster.local:9095"
+      "grpc_client_config":
+        "max_send_msg_size": 104857600
+      "parallelism": 8
+    "ingester":
+      "chunk_block_size": 262144
```
cyriltovena, line 40: I would use snappy, you get better performance.

environments/dev/manifests/loki-config-map.yaml
```diff
@@ -0,0 +1,146 @@
...
+      "ingestion_rate_strategy": "global"
+      "max_global_streams_per_user": 10000
+      "max_query_length": "12000h"
+      "max_query_parallelism": 32
+      "max_streams_per_user": 0
+      "reject_old_samples": true
+      "reject_old_samples_max_age": "168h"
+    "querier":
+      "engine":
+        "max_look_back_period": "2h"
```
cyriltovena, line 74: The default value for this is better 3m. Basically for instant query how far back we're looking for data.

environments/dev/manifests/loki-config-map.yaml
cyriltovena: here is what we find that works well:

```
    chunk_block_size: 262144
    chunk_encoding: snappy
    chunk_idle_period: 2h
    chunk_retain_period: 1m
    chunk_target_size: 1.572864e+06
```

environments/dev/manifests/loki-config-map.yaml
cyriltovena: Actually scratch that you might want to look at 2h for instant query. But that's what this is for. Doesn't really impact all other queries.

environments/dev/manifests/loki-config-map.yaml
```diff
@@ -0,0 +1,146 @@
...
+        "memcached":
+          "batch_size": 100
+          "parallelism": 100
+        "memcached_client":
+          "addresses": "dns+observatorium-xyz-loki-chunk-cache.observatorium.svc.cluster.local:11211"
+          "consistent_hash": false
+          "max_idle_conns": 100
+          "timeout": "100ms"
+          "update_interval": "1m"
+      "max_look_back_period": "2h"
```
cyriltovena, line 16: I would remove this, I think you want to be able to query all data :)

environments/dev/manifests/loki-query-frontend-service.yaml
```diff
@@ -0,0 +1,23 @@
...
+    app.kubernetes.io/version: 1.5.0
+  name: observatorium-xyz-loki-query-frontend
+  namespace: observatorium
+spec:
+  ports:
+  - name: metrics
+    port: 3100
+    targetPort: 3100
+  - name: grcp
+    port: 9095
```
cyriltovena, line 17: I think you need this one to be `clusterIP: None`. Workers connect to frontend manually via endpoints without using  service IP.

periklis: AFAIK, i can configure only per service if the cluster IP should be used. In the present configuration a single service is used for both ports. Setting `clusterIP: None` would disable the service IP for `3100` which is used for ingress traffic from other observatorium components, e.g. our observatorium api

cyriltovena: I think it still fine, the DNS will have 2 IP and the traffic will flow correctly. However having a cluster IP for the GRPC service will cause issue for the frontend workers.

cyriltovena:
```
❯ k describe svc query-frontend -n loki-prod2
Name:              query-frontend
Namespace:         loki-prod2
Labels:            name=query-frontend
Annotations:       kubectl.kubernetes.io/last-applied-configuration:
                     {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"labels":{"name":"query-frontend"},"name":"query-frontend","namespace":"l...
Selector:          name=query-frontend
Type:              ClusterIP
IP:                None
Port:              query-frontend-http-metrics  80/TCP
TargetPort:        80/TCP
Endpoints:         10.52.215.231:80,10.52.42.238:80
Port:              query-frontend-grpc  9095/TCP
TargetPort:        9095/TCP
Endpoints:         10.52.215.231:9095,10.52.42.238:9095
Session Affinity:  None
Events:            <none>
```

Our gateway also hits the 80 port for incoming **read** requests. For writes our gateway hits directly distributors service using httpgrpc.

```
❯ k describe svc distributor -n loki-prod2
Name:              distributor
Namespace:         loki-prod2
Labels:            name=distributor
Annotations:       kubectl.kubernetes.io/last-applied-configuration:
                     {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"labels":{"name":"distributor"},"name":"distributor","namespace":"loki-pr...
Selector:          name=distributor
Type:              ClusterIP
IP:                None
Port:              distributor-http-metrics  80/TCP
TargetPort:        80/TCP
Endpoints:         10.52.13.146:80,10.52.173.28:80,10.52.180.194:80 + 17 more...
Port:              distributor-grpc  9095/TCP
TargetPort:        9095/TCP
Endpoints:         10.52.13.146:9095,10.52.173.28:9095,10.52.180.194:9095 + 17 more...
Session Affinity:  None
Events:            <none>
```

cyriltovena: For writes you should start with http for now, but I'd recommend to look into httpgrpc later on for high volumes since it will multiplex connections.

periklis: For the record our offline discussion summarized:
- Need to split the svc between grpc and http.
- GRPC is doing load balancing itself thus we need a headless svc.
- HTTP can benefit from a separate load-balanced svc via clusterIP.

In addition, tailing requests should go to the querier directly.

components/minio.libsonnet
```diff
@@ -95,7 +99,7 @@
     },
   },
 
-  secret: {
+  secret_thanos: {
```
squat, line 102: nit: we normally use camelcase in jsonnet

components/observatorium.libsonnet
```diff
@@ -49,8 +50,8 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
           local cfg = self,
           name: obs.config.name + '-' + cfg.commonLabels['app.kubernetes.io/name'] + '-' + hashring.hashring,
           namespace: obs.config.namespace,
-          replicas: 3,
-          replicationFactor: 3,
+          replicas: 1,
```
squat, line 53: :+1:  this will bring down our resource utilization for the tests. I wonder if we should make tests adjustments in a new `environments/test` environment. So that `environments/dev` and `environments/base` inherit the higher replica counts by default. we can do that in a follow up though

components/up-job.libsonnet
```diff
@@ -146,7 +150,63 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     ),
   },
 
+  withLogsFile:: {
+    local job = k.batch.v1.job,
+    local container = job.mixin.spec.template.spec.containersType,
+    local u = self,
+    config+:: {
+      bashImage: error 'must provide image for bash',
```
squat, line 158: Is there a reason no to reuse the `observatorium/up` image for the init container? This image hash `/bin/sh`, as it's built on alpine: https://github.com/observatorium/up/blob/master/Dockerfile#L1. I think this would be a nice simplification and reduce the number of images we need to mentally consider

squat: The other option would be to use a configmap with the logs-file as a key. It's shifting the complexity into another place but I think it's the more correct way to load these static files

components/up-job.libsonnet
```diff
@@ -146,7 +150,63 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+            containers: [
+              if c.name == 'observatorium-up' then c {
+                resources: u.config.resources,
+                args+: [
+                  '--logs-file=/var/shared/logs.yaml',
+                ],
+              } + container.withVolumeMounts({
+                name: 'shared',
+                mountPath: '/var/shared',
+                readOnly: false,
```
squat, line 194: this can be `readOnly: true` for this container, no?

components/loki.libsonnet
```diff
@@ -0,0 +1,624 @@
...
+    local gossipRingName = 'gossip-ring',
+    local gossipPort = 7946,
+    local gossipSvc =
+      local service = k.core.v1.service;
+      local ports = service.mixin.spec.portsType;
+      local name = l.config.name + '-' + gossipRingName;
+
+      service.new(
+        name,
+        l.config.podLabelSelector { 'app.kubernetes.io/gossip': 'true' },
```
squat, line 510: we should namespace this label to something other than `app.kubernetes.io` since it's not a well-known or recommended label from k8s: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels. consider `loki.grafana.com/gossip`?

environments/dev/manifests/api-deployment.yaml
```diff
@@ -28,20 +28,20 @@ spec:
...
-        app.kubernetes.io/version: master-2020-05-27-v0.1.1-67-gb2eb292
+        app.kubernetes.io/version: master-2020-06-03-v0.1.1-76-ga89822f
     spec:
       containers:
       - args:
         - --web.listen=0.0.0.0:8080
         - --web.internal.listen=0.0.0.0:8081
-        - --logs.read.endpoint=http://127.0.0.1
-        - --logs.write.endpoint=http://127.0.0.1
+        - --logs.read.endpoint=http://observatorium-xyz-loki-query-frontend-http.observatorium.svc.cluster.local:3100
```
squat, line 37: :raised_hands:

tests/e2e.sh
```diff
@@ -93,15 +95,51 @@ delete_cr() {
...
+
+    $KUBECTL apply -f tests/manifests/observatorium-up-logs.yaml
+
+    sleep 5
+
+    # This should wait for ~2min for the job to finish.
+    $KUBECTL wait --for=condition=complete --timeout=10m -n default job/observatorium-up-logs || (must_gather "$ARTIFACT_DIR" && exit 1)
+}
+
+must_gather() {
```
squat, line 117: very cool

environments/base/manifests/loki-distributor-grpc-service.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/instance: observatorium-xyz
+    app.kubernetes.io/name: loki
+    app.kubernetes.io/part-of: observatorium
+    app.kubernetes.io/version: master-815c475
+  name: observatorium-xyz-loki-distributor-grpc
```
squat, line 9: why don't we combine the various services that select one single set of pods (e.g. `-grpc` and `-http`) into the same service and instead specify multiple port combinations? I think this would reduce some mental overhead

i.e. we have a loki-distributor service, a loki-frontend service, and a loki-querier service rather than 2-3 for each of those components

environments/base/manifests/loki-distributor-grpc-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: observatorium
+spec:
+  clusterIP: None
+  ports:
+  - name: grcp
+    port: 9095
+    targetPort: 9095
+  selector:
+    app.kubernetes.io/component: observatorium-xyz-loki-distributor
```
squat, line 18: It looks like all of the loki manifests are using the name of the resource as the component label, but this is not the pattern we follow. According to the recommended labels from K8s, this should be the _logical_ purpose of this resource in the stack, i.e. `logs-ingestion`, `load-balancer` etc

environments/base/manifests/loki-distributor-grpc-service.yaml
periklis: The reason why we have 2 services is reflected in GRPC being responsible for load-balancing itself. Thus for GRPC communication we need a headless service. However, for incoming HTTP traffic we keep the k8s load-balancing via ClusterIP. The gossip service is a cross-component service that constraints communication only between distributors, ingesters and queriers based on an extended label selector set. IMHO it is not possible to unify all three demand in one service, e.g. enabling ClusterIP for the HTTP port while headless for the GRPC. For more info please check above the discussion with @cyriltovena:

squat: ack sounds good :+1:

components/loki.libsonnet
periklis: I addressed this point. You are absolutely right!

components/up-job.libsonnet
periklis: I addressed this point. You are absolutely right!

jsonnetfile.json
```diff
@@ -40,11 +40,11 @@
     {
       "source": {
         "git": {
-          "remote": "https://github.com/observatorium/observatorium",
+          "remote": "https://github.com/periklis/observatorium",
```
periklis, line 43: Revert this back to upstream master when observatorium/observatorium#72 is merged.

squat: ready!

environments/base/manifests/loki-distributor-grpc-service.yaml
periklis: Good catch!

components/up-job.libsonnet
periklis: First of all moving this away is not a good idea, because the logs file is a construct consumed only by observatorium/up. Furthermore, the need for a special image is due to the fact that the Loki API demands timestamps as unix epoch in nanoseconds (See [/loki/api/v1/push](https://github.com/grafana/loki/blob/master/docs/api.md#post-lokiapiv1push)). The most simple date implementation supporting this is GNU's coreutils' date. Alpine does not provide support for timetamps as unix epoch in nanoseconds.

squat: ok, not work bikeshedding this too much :)

squat: ok, to simplify our lives I would say to simply append some zeros to the `date` output, e.g.: `echo $(date +%s)000000000)` to get a fake nanosecond timestamp, but it's not worth bikeshedding this too much :)

squat:
I'm very satisfied with this :) Let's get this merged and we can follow up in the future in smaller PRs

PR #287, opened 2020-05-27
 Use defaults for images in operator

### LINKED ISSUE #284: Images in the CRD should be optional, with defaults handled by operator
We would like to have defaults for the images in the Observatorium CR. There are currently are 11 images to configure (6x Thanos,api,queryCache, memcached, memcached_exporter, receivecontroller). This could be a hurdle for the user.

1. Make images as optional in the CRD.
2. Operator will use preconfigured images if the user did not specified them in the CR.
3. The user can override one or more images.

rollandf:
Fixes #284

### DIFF
--- a/operator/api/v1alpha1/observatorium_types.go
+++ b/operator/api/v1alpha1/observatorium_types.go
@@ -29,21 +29,21 @@ type ObservatoriumSpec struct {
 	// Thanos CompactSpec
 	Compact CompactSpec `json:"compact"`
 	// Thanos Receive Controller Spec
-	ThanosReceiveController ThanosReceiveControllerSpec `json:"thanosReceiveController"`
+	ThanosReceiveController ThanosReceiveControllerSpec `json:"thanosReceiveController,omitempty"`
 	// Thanos ThanosPersistentSpec
 	Receivers ReceiversSpec `json:"receivers"`
 	// Thanos QueryCache
-	QueryCache QueryCacheSpec `json:"queryCache"`
+	QueryCache QueryCacheSpec `json:"queryCache,omitempty"`
 	// Thanos StoreSpec
 	Store StoreSpec `json:"store"`
 	// Thanos RulerSpec
 	Rule RuleSpec `json:"rule"`
 	// API
-	API APISpec `json:"api"`
+	API APISpec `json:"api,omitempty"`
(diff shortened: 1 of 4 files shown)


rollandf:
Should we have an additional CR example to test this in CI?

operator/api/v1alpha1/observatorium_types.go
```diff
@@ -55,14 +55,14 @@ type ObjectStorageConfig struct {
 
 type ThanosReceiveControllerSpec struct {
 	// Receive Controller image
-	Image string `json:"image"`
+	Image string `json:"image,omitempty"`
```
brancz, line 58: I would argue that _if_ `image` is set, we want to force users to also set version. So they understand that they need to appropriately set `version` according to what's in the `image` they specify.

rollandf:
> Love the direction! I'm trying to think of how this would work with productization, typically the red hat pipelines need it to be flags on the operator binary etc. and they are replaced by the machinery. Do we know how this would work? (I'm not requesting to solve this today, just we should think about it, we might just set the "defaults" via flags eventually)

I was thinking of having a downstream version of `default-config.libsonnet` that will have the downstream images fitting a specific version. Then is the Downstream Docker file just copy it instead of the upstream one.

operator/api/v1alpha1/observatorium_types.go
rollandf: Well, this is not especially related to this patch, but I had always doubts about the `version` field and for is the purpose of it.
I am not sure how to address your comment actually.

brancz: I'd be curious if we could have conditional validation in the openapischema? Not blocking this I think, but we should have an issue to track this.

brancz:
> I was thinking of having a downstream version of default-config.libsonnet that will have the downstream images fitting a specific version. Then is the Downstream Docker file just copy it instead of the upstream one.

At least looking at how things worked in OpenShift, this would not have worked, but let's figure out how to solve that when we're there :)

operator/api/v1alpha1/observatorium_types.go
rollandf: I open issue #288 for follow up

PR #292, opened 2020-06-01
Fix thanos shard statefulset

rollandf:


nmagnezi:
LGTM 👍 , CI failure seem unrelated.

rollandf:
Not needed any more fixed by #293

PR #306, opened 2020-06-19
Streamline CI resource utilization

periklis:
This PR addresses the resource utilization of the components used on CI to run E2E tests. This enables CI runs with a full-blown Loki cluster (See #281).

/cc @squat


periklis:
> this is super helpful :)
> 
> note: we need to ensure that the replica and resource changes down propagate downstream to the real infra

I will open a PR in a moment for this in our production repo.

PR #312, opened 2020-07-02
WIP: add jeager service monitor

krasi-georgiev:

### DIFF
--- a/components/jaeger-collector.libsonnet
+++ b/components/jaeger-collector.libsonnet
@@ -3,8 +3,9 @@ local service = k.core.v1.service;
 local jaegerAgent = import './jaeger-agent.libsonnet';
 
 {
-  jaeger+:: {
-    local j = self,
+  local _self = self,
+
+  config+:: {
     namespace:: error 'must set namespace for jaeger',
     image:: error 'must set image for jaeger',
     name:: error 'must set name for jaeger',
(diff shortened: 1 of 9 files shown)


components/jaeger-collector.libsonnet
```diff
@@ -7,11 +7,16 @@ local jaegerAgent = import './jaeger-agent.libsonnet';
     local j = self,
     namespace:: error 'must set namespace for jaeger',
     image:: error 'must set image for jaeger',
+    name:: error 'must set name for jaeger',
     replicas:: 1,
     pvc:: {
       class: 'standard',
       size: '50Gi',
     },
+    commonLabels:: {
```
brancz, line 16: this should also be used in the deployment

components/jaeger-collector.libsonnet
```diff
@@ -3,110 +3,150 @@ local service = k.core.v1.service;
 local jaegerAgent = import './jaeger-agent.libsonnet';
 
 {
-  jaeger+:: {
-    local j = self,
+  local _self = self,
```
brancz, line 6: in other places we call this what the object is called, which would be `jaeger` in this case.. I find `_self` to be confusing as `self` is relative to the object but `_self` wouldn't be.

krasi-georgiev: why `_self` isn't relative to the object?

I find it confusing the other way around. for example, in many places, I see `obs` and it is very difficult to trace down what does it point to.

brancz: it's always the top level object that's being exported, I think the pattern is discussable, but if we change it we should do it consistently and not do it in one place but not the rest

krasi-georgiev: yes I see what you mean, I am still trying different things so let me think about this a bit again.

brancz: (if I wasn't clear I think we should not change the style for all others in this PR as well, in this PR we should stay consistent with the rest and then in a follow up we can discuss style changes for everything)

krasi-georgiev:
will try using functions

PR #316, opened 2020-08-03
Make loki is optional

clyang82:
Fixes: https://github.com/observatorium/deployments/issues/314

### DIFF
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -224,7 +224,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       namespace: obs.config.namespace,
       replicas: 1,
       commonLabels+:: obs.config.commonLabels,
-      logs: {
+      [if std.length(obs.config.loki) != 0 then 'logs']: {
         readEndpoint: 'http://%s.%s.svc.cluster.local:%d' % [
           obs.loki.manifests['query-frontend-http-service'].metadata.name,
           obs.loki.manifests['query-frontend-http-service'].metadata.namespace,
@@ -314,5 +314,6 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
   } + {
     ['loki-' + name]: obs.loki.manifests[name]
     for name in std.objectFields(obs.loki.manifests)
+    if std.length(obs.config.loki) != 0
   },
 }
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -224,7 +224,20 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       namespace: obs.config.namespace,
       replicas: 1,
       commonLabels+:: obs.config.commonLabels,
-      [if std.length(obs.config.loki) != 0 then 'logs']: {
+      metrics: {
+        readEndpoint: 'http://%s.%s.svc.cluster.local:%d' % [
+          obs.queryCache.service.metadata.name,
+          obs.queryCache.service.metadata.namespace,
+          obs.queryCache.service.spec.ports[0].port,
+        ],
+        writeEndpoint: 'http://%s.%s.svc.cluster.local:%d' % [
+          obs.receiveService.metadata.name,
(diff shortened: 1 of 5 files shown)


brancz:
cc @periklis

operator/jsonnet/operator-config.libsonnet
```diff
@@ -64,10 +64,10 @@ cr.spec {
...
-    replicas: cr.spec.loki.replicas,
-    version: cr.spec.loki.version,
-    objectStorageConfig: lokiObjectStorageConfig,
-  },
+  loki+:: if std.objectHas(cr.spec, 'loki') then {
+    image: if std.objectHas(cr.spec, 'loki') && std.objectHas(cr.spec.loki, 'image') then cr.spec.loki.image else default.loki.image,
+    replicas: if std.objectHas(cr.spec, 'loki') && std.objectHas(cr.spec.loki, 'replicas') then cr.spec.loki.replicas else default.loki.replicas,
+    version: if std.objectHas(cr.spec, 'loki') && std.objectHas(cr.spec.loki, 'version') then cr.spec.loki.version else default.loki.version,
+    objectStorageConfig: if lokiObjectStorageConfig != null then lokiObjectStorageConfig else default.loki.objectStorageConfig,
+  } else {},
```
periklis, line 72: nit. duplicate checks `std.objectHas(cr.spec, 'loki')` inside conditional block.

components/observatorium.libsonnet
```diff
@@ -224,7 +224,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       namespace: obs.config.namespace,
       replicas: 1,
       commonLabels+:: obs.config.commonLabels,
-      logs: {
+      [if std.length(obs.config.loki) != 0 then 'logs']: {
```
periklis, line 227: Consider using the style in [L315-317](https://github.com/observatorium/deployments/pull/316/files#diff-bf6438e7ce24362188f3a78b9183b071R315-R317) for adding optional into a config object.

operator/jsonnet/operator-config.libsonnet
clyang82: Done

components/observatorium.libsonnet
clyang82: Done

clyang82:
@periklis Thanks for your help.

components/observatorium.libsonnet
```diff
@@ -224,6 +224,19 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+          obs.queryCache.service.metadata.namespace,
+          obs.queryCache.service.spec.ports[0].port,
+        ],
+        writeEndpoint: 'http://%s.%s.svc.cluster.local:%d' % [
+          obs.receiveService.metadata.name,
+          obs.receiveService.metadata.namespace,
+          obs.receiveService.spec.ports[2].port,
+        ],
+      },
+    } + if std.length(obs.config.loki) != 0 then {
```
periklis, line 239: nit. prefer the more visual function `std.objectHas(o, f)`

> Returns true if the given object has the field (given as a string), otherwise false. Raises an error if the arguments are not object and string respectively. Returns false if the field is hidden.

clyang82: Thanks @periklis.  For this case, I think there is an empty `loki` object. because
```
  loki+:: if std.objectHas(cr.spec, 'loki') then {
    image: if std.objectHas(cr.spec.loki, 'image') then cr.spec.loki.image else default.loki.image,
    replicas: if std.objectHas(cr.spec.loki, 'replicas') then cr.spec.loki.replicas else default.loki.replicas,
    version: if std.objectHas(cr.spec.loki, 'version') then cr.spec.loki.version else default.loki.version,
    objectStorageConfig: if lokiObjectStorageConfig != null then lokiObjectStorageConfig else default.loki.objectStorageConfig,
  } else {},
```
correct me If I am wrong. Thanks.

periklis: let's move forward with this PR, it's just a nit nothing special.

PR #317, opened 2020-08-03
Patch volumeClaimTemplates

clyang82:
fixes: https://github.com/observatorium/deployments/issues/313


clyang82:
Do not have permission to rerun the failed circleci test.

brancz:
I think we should only do this in the operator jsonnet.

clyang82:
@brancz Thanks for response. Do you mean that we should do this in https://github.com/observatorium/deployments/blob/master/operator/jsonnet/obs-operator.jsonnet? why?

brancz:
The component is literally just a component, whereas you are configuring concrete use of the component, which is exactly why it should be in the file you linked.

clyang82:
@squat PTAL
Thanks.

PR #318, opened 2020-08-06
Update API to reflect loki is optional

clyang82:
1. Update the observatorium api jsonnet vendor
2. Update observatorium operator api to allow the Loki is not present.


clyang82:
Thanks @periklis

PR #320, opened 2020-08-11
thanos *: Upgrade to latest kube-thanos

kakkoyun:

## Changes
* Add ability to change the log level https://github.com/thanos-io/kube-thanos/pull/143
* Only perform anti-affinity against receive pods https://github.com/thanos-io/kube-thanos/pull/134

### DIFF
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -70,9 +70,6 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
               spec+: {
                 containers: [
                   if c.name == 'thanos-receive' then c {
-                    args+: [
-                      '--log.level=' + obs.config.receivers.logLevel,
-                    ],
                     env+: [
                       {
                         name: 'DEBUG',
--- a/environments/base/default-config.libsonnet
+++ b/environments/base/default-config.libsonnet
@@ -44,6 +44,7 @@
         },
       },
     },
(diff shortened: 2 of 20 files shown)


components/observatorium.libsonnet
```diff
@@ -29,6 +29,22 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         deduplicationReplicaLabels: obs.config.deduplicationReplicaLabels,
         deleteDelay: '48h',
       },
+      statefulSet+: {
+        spec+: {
+          template+: {
+            spec+: {
+              containers: [
+                if c.name == 'thanos-compact' then c {
+                  args+: [
```
brancz, line 38: I think a slightly modified strategy might be better here, I think we should first filter any `--log.level=` arg out, and then add it again, so that if any such multiple merges are essentially idempotent.

kakkoyun:
@brancz Thanks, great idea! Just to move fast, let's merge this after the idempotency addressed and I'm gonna create a PR for upstream `kube-thanos` for all the components.

brancz:
If we're not in a hurry then I would still prefer kube-thanos first, but I'll let it be your call.

kakkoyun:
@brancz Here you go sir https://github.com/thanos-io/kube-thanos/pull/143.
I'm gonna update this PR after upstream one merged.

PR #325, opened 2020-08-20
Update jsonnet vendor folder in Makefile to root vendor/

metalmatze:
For some unknown reason we started using the wrong jsonnet vendor folder. This fixes it to use the top-level `vendor/` again.


Makefile
```diff
@@ -66,26 +66,26 @@ environments/base/manifests: environments/base/main.jsonnet $(JSONNET_SRC)
 	-make jsonnetfmt
```
rollandf, line 66: Adding 'vendor-jsonnet' on line 65 will be enough.

This work has been done for downstream build of the operator to separate GO vendor from Jsonnet Vendor

metalmatze: I'm not exactly sure what you mean? This is about the operator then, right? I'm trying to get the other manifests to be generated from the root vendor folder, as that's what we actually should be using.

rollandf: The vendor folder has moved to the operator dir in #257
What do you mean `that's what we actually should be using.`
Does it matter if the Jsonnet vendor dir is not in the root ?

metalmatze: We currently have 2 jsonnet vendor directories in `vendor/` and `operator/jsonnet/vendor`. Only the operator should use the `operator/jsonnet/vendor/` whereas the other manifests should be generated from `vendor/`.
There are also 2 different jsonnetfile.json. That's what I'm trying to clean up.
I think the underlying problem is, that the boundaries (even within the same repository and reuse of code) between the generate manifests and the operator aren't quite clear.

metalmatze:
This is not needed anymore, as we move the operator out of this repository into its separate.

Issue #315, opened 2020-07-30
Feature Request: should have a configurable params for enable/disable downsampling


songleo:
I installed observatorium-operator and found the thanos-compact disable downsampling (`--downsampling.disable`). I didn’t find any config to enable it, we should have this be configurable in the CR. For detail: https://coreos.slack.com/archives/C010A637DGB/p1594884831397400

brancz:
yeah I think making this configurable and have downsampling on by default makes sense

clyang82:
I am working on that.

PR #327, opened 2020-08-31
Configure enable/disable downsampling

clyang82:
fixes: #315

### DIFF
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -28,7 +28,11 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         deduplicationReplicaLabels: obs.config.deduplicationReplicaLabels,
         deleteDelay: '48h',
       },
-    } + (if std.objectHas(obs.config.compact, 'enableDownsampling') && obs.config.compact.enableDownsampling == true then {} else t.compact.withDownsamplingDisabled),
+    } + (
+      if std.objectHas(obs.config.compact, 'enableDownsampling') && obs.config.compact.enableDownsampling == true then
+        {}
+      else t.compact.withDownsamplingDisabled
+    ),
 
   thanosReceiveController:: (import 'thanos-receive-controller/thanos-receive-controller.libsonnet') + {
     config+:: {
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -17,7 +17,6 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
   compact::
     t.compact +
     t.compact.withRetention +
-    t.compact.withDownsamplingDisabled +
     t.compact.withDeleteDelay +
     t.compact.withDeduplication + {
       config+:: {
--- a/components/observatorium.libsonnet
+++ b/components/observatorium.libsonnet
@@ -28,7 +28,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         deduplicationReplicaLabels: obs.config.deduplicationReplicaLabels,
         deleteDelay: '48h',
       },
-    },
+    } + (if std.objectHas(obs.config.compact, 'enableDownsampling') && obs.config.compact.enableDownsampling == true then {} else t.compact.withDownsamplingDisabled),
 
   thanosReceiveController:: (import 'thanos-receive-controller/thanos-receive-controller.libsonnet') + {
     config+:: {
--- a/environments/base/default-config.libsonnet
+++ b/environments/base/default-config.libsonnet
@@ -34,7 +34,7 @@
     retentionResolutionRaw: '14d',
     retentionResolution5m: '1s',
     retentionResolution1h: '1s',
-    enableDownSampling: true,
+    enableDownsampling: true,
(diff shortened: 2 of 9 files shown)


operator/jsonnet/obs-operator.jsonnet
```diff
@@ -8,7 +8,7 @@ local patchObs = obs {
   compact+::
     t.compact.withVolumeClaimTemplate {
       config+:: obs.compact.config,
-    },
+    } + (if std.objectHas(obs.compact.config, 'enableDownSampling') && obs.compact.config.enableDownSampling == false then t.compact.withDownsamplingDisabled else {}),
```
squat, line 11: This configuration defaults to having downsampling enabled, but this is not what the CRD field name implies. Let's flip the logic here so that you need to set the EnableDownsampling field to true for it to actually be enabled.

operator/api/v1alpha1/observatorium_types.go
```diff
@@ -227,6 +227,8 @@ type CompactSpec struct {
 	RetentionResolution5m string `json:"retentionResolution5m"`
 	// RetentionResolutionRaw
 	RetentionResolution1h string `json:"retentionResolution1h"`
+	// EnableDownSampling enables the downsampling
```
squat, line 230: Nit: 
```suggestion
	// EnableDownsampling enables downsampling.
```

squat (commented):
Overall I am fine with this feature :)) let's clean up some small things and merge

operator/manifests/crds/core.observatorium.io_observatoria.yaml
```diff
@@ -179,6 +179,9 @@ spec:
             compact:
               description: Thanos CompactSpec
               properties:
+                enableDownSampling:
```
squat, line 182: nit, if we are treating `downsampling` as one single word, then the casing should be:
```suggestion
                enableDownsampling:
```

operator/manifests/crds/core.observatorium.io_observatoria.yaml
```diff
@@ -179,6 +179,9 @@ spec:
...
+                  description: EnableDownSampling enables downsampling.
```
squat, line 183: same as above:

```suggestion
                  description: EnableDownsampling enables downsampling.
```

operator/manifests/crds/core.observatorium.io_observatoria.yaml
clyang82: sorry for missing the previous comment to use `downsampling` as a single word.

environments/base/default-config.libsonnet
```diff
@@ -34,6 +34,7 @@
     retentionResolutionRaw: '14d',
     retentionResolution5m: '1s',
     retentionResolution1h: '1s',
+    enableDownsampling: true,
```
squat, line 37: note: here we are changing the default downsampling behavior of all environments that inherit from `base`.

cc @observatorium/maintainers

clyang82: Can we have it merged if no further concerns? @squat

kakkoyun: I think we shouldn't change the default behaviour for the `base`.

kakkoyun (changes requested):

clyang82:
@kakkoyun Thanks for your comments. Can `retentionResolution5m` and `retentionResolution1h` work as design without enable downsampling?

environments/base/default-config.libsonnet
squat: We should preferable leave the default to disabled and then if you want it, you can explicitly enable it in the custom resource when you run the operator. WDYT, @clyang82?

kakkoyun:
@clyang82 Yes, they would. When downsampling is disabled it won't create any new downsampled blocks, but the retention is controlled by those flags. https://thanos.io/tip/components/compact.md/#downsampling-resolution-and-retention

environments/base/default-config.libsonnet
clyang82: Yes. Agree with that. Let me update accordingly.

components/observatorium.libsonnet
```diff
@@ -29,7 +28,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         deduplicationReplicaLabels: obs.config.deduplicationReplicaLabels,
         deleteDelay: '48h',
       },
-    },
+    } + (if std.objectHas(obs.config.compact, 'enableDownsampling') && obs.config.compact.enableDownsampling == true then {} else t.compact.withDownsamplingDisabled),
```
metalmatze, line 31: I think a new line would be nice here. At least something like:
```suggestion
} + (
  if std.objectHas(obs.config.compact, 'enableDownsampling') && obs.config.compact.enableDownsampling == true then
    {}
  else t.compact.withDownsamplingDisabled
)
```

metalmatze (changes requested):
Let's add a new line on a large one, otherwise looking good.

components/observatorium.libsonnet
clyang82: done

PR #332, opened 2020-09-07
Improve loki boltdb shipper local and s3 storage setup

periklis:
### Description
This PR addresses two reliability issues when operating Loki with `boltdb-shipper` on S3 compatible storages beyond minio:
1. The ingester requires to be operated as a StatefulSet to ensure zero index file loss on replica restarts.
2. The S3 storage access is split to the required AWS environment variables instead of a simple URL endpoint.

For boltdb-shipper operations, consider [boltdb-shipper docs](https://grafana.com/docs/loki/latest/operations/storage/boltdb-shipper/#ingesters).


environments/base/default-config.libsonnet
```diff
@@ -1,3 +1,4 @@
+local l = (import '../../components/loki.libsonnet');
```
squat, line 1: Is this actually used?

operator/api/v1alpha1/observatorium_types.go
```diff
@@ -54,7 +54,7 @@ type ObjectStorageConfig struct {
 	Thanos *ObjectStorageConfigSpec `json:"thanos"`
```
squat, line 54: Can we also have the corresponding name change here?

operator/api/v1alpha1/observatorium_types.go
```diff
@@ -64,6 +64,26 @@ type ObjectStorageConfigSpec struct {
 	Key string `json:"key"`
 }
 
+type LokiObjectStorageConfigSpec struct {
+	// Object Store Config Secret Name
+	Name string `json:"name"`
```
squat, line 69: Could we use `SecretName`? I think this helps communicate the expectation of the field, vs maybe suggesting we are naming this struct

environments/base/default-config.libsonnet
periklis: Nope!

operator/api/v1alpha1/observatorium_types.go
periklis: Will do, but I will leave the thanos config as is.

periklis:
@squat Could you please take a look again.

squat (approved):
lgtm :))

PR #335, opened 2020-09-08
Remove the Operator after moving to separate repository

metalmatze:
The Operator moved to a separate repository in https://github.com/observatorium/operator

This PR cleans up this repository, which going forward, should only include the generated deployments YAML.

/cc @observatorium/maintainers


rollandf:
The new repository https://github.com/observatorium/operator is not keeping the jsonnet vendored files in the repo.
It will fail in downstream, as builds are done offline.

rollandf:
FYI @clyang82

clyang82:
@metalmatze Can we cut a branch before deleting to destroy the operator in deployment repo. ACM observability depends on it. We do not make a plan to switch the dependency from `deployments` to `operator`. We need to raise it in our weekly meeting before moving forward. As @rollandf mentioned that if it breaks the downstream, we need to consider it.

metalmatze:
Sure, we can create a branch but that branch won't be expected to be maintained. We asked several weeks ago about making this move and haven't heard about any blockers.
Please let me know when you discussed it at the meeting. I'll create an `old-operator` branch from the current master branch.

clyang82:
Thanks @metalmatze I just want to ensure that we have build the DS image successfully then we can do the switch accordingly. I will give you update soon. Thanks.

metalmatze:
The PR to update the Operator repo to the latest state of this repository has been merged: https://github.com/observatorium/operator/pull/2

We also commit the full vendor directory again, so offline builds should work. Is there anything else blocking merging this PR now? I'd really want to avoid having 2 same operators for too long.

kakkoyun (commented):
Nice! However, tests fail with `error: the path "tests/manifests/observatorium-xyz-tls-configmap.yaml" does not exist`. Otherwise looks good to me.

metalmatze:
Whoops. Yes, I already looked into them and hopefully fixed them. It seems I forgot to push them before lunch, which I wanted to :man_shrugging:

squat:
This is looking pretty good. It looks like the `build` test is still doing something related to the operator and thus failing

kakkoyun (approved):
💯

PR #336, opened 2020-09-10
Bump loki queriers to statefulsets to use pvc for sync'ing indices

periklis:
### Description
This PR addresses a performance optimization when running Loki queriers with boltdb indices. In detail it bumps queriers to statefulsets to use PVC for storing boltdb index files after syncing them from the object storage. This ensures that queriers don't require to share the node disk. Furthermore it aligns with grafana/loki/pull/2547.


environments/base/manifests/loki-querier-deployment.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
```
squat, line 2: In general, we always try to have the filename match the k8s resource type, so if the loki querier is a statefulset, then the file should be called `loki-querier-statefulset.yaml`

components/loki.libsonnet
```diff
@@ -691,12 +694,13 @@ local k = (import 'ksonnet/ksonnet.beta.4/k.libsonnet');
...
-    [normalizedName(name) + '-deployment']:
-      if name == 'ingester' then
-        newStatefulSet(name, loki.components[name])
-      else
-        newDeployment(name, loki.components[name])
+    [normalizedName(name) + '-deployment']: newDeployment(name, loki.components[name])
+    for name in std.objectFields(loki.components)
+    if !isStatefulSet(name)
+  } + {
+    [normalizedName(name) + '-statefulset']: newDeployment(name, loki.components[name])
```
squat, line 701: this should have `newStatefulSet` I think

periklis: :facepalm: naming is hard, because of renaming ;D

periklis:
@squat I believe we are fine now to merge? WDYT?

squat (approved):

PR #343, opened 2020-09-17
Create SA for memcached

clyang82:
Pod analysis shows most observability pods use the default service account with cluster-admin access. A separate service account should be used by the pods so they can each be bound to only the privileges they need. None of them should be using the default service account or the wild card privileges [map[apiGroups:[*] resources:[*] verbs:[*]] map[nonResourceURLs:[*] verbs:[*]]] that the default service account currently uses.

Thanks @metalmatze and @squat for  the comments to provide the service account in observatorium project directly instead of making changes in thanos side.

### DIFF
--- a/components/memcached.libsonnet
+++ b/components/memcached.libsonnet
@@ -37,6 +37,14 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
     },
   },
 
+  serviceAccount:
+    local sa = k.core.v1.serviceAccount;
+
+    sa.new() +
+    sa.mixin.metadata.withName(mc.config.name) +
+    sa.mixin.metadata.withNamespace(mc.config.namespace) +
+    sa.mixin.metadata.withLabels(mc.config.commonLabels),
+
   service:
     local service = k.core.v1.service;
     local ports = service.mixin.spec.portsType;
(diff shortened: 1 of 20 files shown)


metalmatze:
Generally, I'm happy to accept this :)

I'm wondering though, where do you see that the default service account has cluster-admin? I'm surprised to learn this. Usually the default ServiceAccount has very very limited privileges if not none at all. That's why using the default ServiceAccount is usually totally fine.

clyang82:
@metalmatze Sometimes cluster-admin role was added because someone wanted to make their CI/CD tool (e.g. Jenkins) manage Kubernetes resources in the cluster, and it was easier to apply cluster-admin to a default service account than to set all the individual RBAC privileges correctly. so the the pod who use default service account has the cluster-admin role. 

I will append more when I got from our security experts. Thanks.

metalmatze:
Hm, that seems interesting and I'm not sure that fixing this via changing all ServiceAccounts in other projects is the correct solution. I don't mind merging this, I was just curious what really happened, cause I've never seen that before.

squat:
Yes agreed with @metalmatze here. Having thanos use a non default account doesn't fix the fact that there is a default service account with cluster-admin

clyang82:
Hi @squat and @metalmatze Sorry for misleading about `cluster-admin` permission. The original ticket mentioned that the default account has the cluster-admin permission, it means observatorium operator bundles some privileges to the default service account so that the other components who are using the default service account will have such privileges. here is clusterrolebinding in operator side - https://github.com/observatorium/operator/blob/0aed9f935ac6cc097300b54a7936c5cb8f4ba12c/manifests/cluster_role_binding.yaml#L14

I am going to fix the operator side to use `observatorium` service account.

But I would like to have this PR merged as well so that we can guarantee our components use "least privilege" with our owned service account. Do you agree with that? Thanks.

PR #345, opened 2020-09-28
[loki] Remove obsolete table manager component

periklis:
This PR addresses removes Loki's table managercomponent. The current setup is based on boltdb-shipper and table-manager is not used there. Furthermore the component will be replaced by the soon to be released compactor.


brancz:
Should we even have it optional? Kind of the point of the project is to be opinionated, and the opinion is to never use NoSQL. I'd be even for removing this entirely.

bwplotka:
As I said in previous MR, I would remove it totally.

bwplotka (approved):
LGTM!

PR #112, opened 2019-11-06
Add network_attachment_definition for multus changes

aneeshkp:
whitelist metrics and update  temeter version  for following metrics   cluster:network_attachment_definition_instances:max cluster:network_attachment_definition_enabled_instance_up:max
Continuation from
1. https://github.com/openshift/telemeter/pull/262 
2. https://github.com/openshift/cluster-monitoring-operator/pull/537

PR #347, opened 2020-10-22
Add TLS to dex

jpkrohling:
Adds TLS support for dex by annotating the service with OpenShift-specific annotations, so that a TLS secret is provisioned. This is then mounted in the deployment and used by the dex configuration.
This is required by the OpenTelemetry Collector, which requires TLS when auth is enabled.

﻿Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


jpkrohling:
@squat, could you please review this one? If it looks good, I'll manually apply the resulting manifests to the dev cluster.

squat:
This is going to break our e2e tests in this repo, which try to deploy dex. Can we generate certs for testing? Else we'll need to move the TLS stuff downstream somewhere.

jpkrohling:
I should be able to adapt the e2e tests.

jpkrohling:
Current status: the up job seems to be failing with:

```
caller=level.go:63 ts=2020-10-22T15:25:27.481618014Z name=up level=error caller=main.go:102 component=writer msg="failed to make request" err="non-200 status: 401 Unauthorized"
caller=level.go:63 ts=2020-10-22T15:25:27.481973081Z name=up level=error caller=main.go:130 component=reader msg="failed to query" err="query response parse failed: invalid character 'e' looking for beginning of value"
```

I'll take a look tomorrow, but if anyone has an idea on what I'm missing, I'd appreciate a hint :-)

squat:
Yes, it's failing because the issuer URL in the tenant configuration needs to be updated: https://github.com/observatorium/deployments/blob/master/environments/dev/main.jsonnet#L87
Otherwise, Observatorium can't instantiate the OIDC provider to verify tokens in requests.

This is why the Observatorium API is logging:
```
level=warn name=observatorium ts=2020-10-22T15:18:43.828940736Z caller=main.go:317 msg="failed to instantiate OIDC provider for tenant \"test\": Get \"https://dex.dex.svc.cluster.local:5556/dex/.well-known/openid-configuration\": http: server gave HTTP response to HTTPS client"
```

environments/dev/main.jsonnet
```diff
@@ -84,7 +84,7 @@ local obs = (import '../base/observatorium.jsonnet') + {
             oidc: {
               clientID: dex.config.config.staticClients[0].id,
               clientSecret: dex.config.config.staticClients[0].secret,
-              issuerURL: 'http://%s.%s.svc.cluster.local:%d/dex' % [
+              issuerURL: 'https://%s.%s.svc.cluster.local:%d/dex' % [
```
jpkrohling, line 87: @squat the line you references was changed in the PR

jpkrohling:
> server gave HTTP response to HTTPS client

This seems to indicate that the server isn't configured with TLS?

jpkrohling:
I got past the previous problem, only to get to a new roadblock that I'm not quite sure how to overcome.

The state of the PR as of the first commit will lead to a scenario where the Observatorium API will complain that it can't get the OIDC server configuration for the tenant "test" because the TLS cert can't be verified (unknown CA).

In a real cluster, DEX will get its cert provisioned by OpenShift, so, we just need to mount the service-ca into the API deployment and point the tenants.yaml to the service.ca.crt . 

And this is where I'm not sure how to proceed: for dev, the manifests should be OK (yet to be tested locally with crc), but for tests, the API deployment will get created and the config map will not be populated, so, it will fail. The container is in "Error" state and the test will fail before the container starts again.

squat:
Thanks a lot for the effort so far @jpkrohling. One thing indogn quite understand however is  why the test scenario is any different from the deployment scenario. In both cases we need to mount the CA into the API's container and point the tenant's configuration at the CA. The Observatorium API provides two fields in the tenants.yaml file for this purpose:
https://github.com/observatorium/observatorium/blob/master/main.go#L125-L126
That is, the CA can be given literally or as a file path. Is there a reason we do this in the test?

jpkrohling:
I just squashed the commits and updated the PR. I had a bug in the deploy script that caused the `dex-tls` to be filled with the test data, which would cause consumers to fail with TLS verification. Right now, applying the dev manifests makes everything work and dex has TLS in front of it.

>  why the test scenario is any different from the deployment scenario

OpenShift can provision the server cert for dex automatically by annotating the dex service:

https://github.com/observatorium/deployments/blob/366e007e11a7b88bc4b2864b9a0cb6a652c974d9/components/dex.libsonnet#L163

As the dex deployment references this secret, the deployment won't be created until the cert is provisioned, ie, the secret is ready. In the e2e tests, `kind` won't automatically create it, so, we need to explicitly create this secret before the `dex()` function is called, otherwise, it times out.

Later on, we need to have access to the CA that was used to issue the cert. For OpenShift, we can annotate a configmap that will be populated with that:

https://github.com/observatorium/deployments/blob/366e007e11a7b88bc4b2864b9a0cb6a652c974d9/environments/dev/main.jsonnet#L144

This is done via webhooks on OpenShift, so, at the time this configmap is created, it has the `service-ca.crt` entry already with a valid cert. We then mount this in the places where we need (otelcol, observatorium-api) and reference this in the OIDC configuration. 

For the tests, we also need to create this secret before the consumers are created, otherwise they will load an empty configmap.

I think we might need a step to run before the `deploy()` function to create the `dex-tls` secret, and one before the `test()` (but after `deploy()` to adjust the `service-ca-tls` configmap.

jpkrohling:
As discussed in Slack, the build failure here is because I manually changed the Observatorium's jsonnet lib, which made it generate new YAML files, causing the diff in the build.

The PR observatorium/observatorium#110 was opened based on my changes to the lib. Once that one is merged, I'll update this PR.

jpkrohling:
Current status: waiting on https://github.com/observatorium/observatorium/pull/111 and https://github.com/observatorium/observatorium/issues/112

components/up-job.libsonnet
```diff
@@ -127,7 +127,8 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
                 '-c',
                 |||
                   curl --request POST \
-                      --silent \
+                      --verbose \
+                      --insecure \
```
jpkrohling, line 131: I think I might need to take care of this.

jpkrohling (commented):
Other than the `--insecure` in the curl command, this is ready to be reviewed.

components/up-job.libsonnet
jpkrohling: Done!

components/up-job.libsonnet
```diff
@@ -128,6 +129,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
                 |||
                   curl --request POST \
                       --silent \
+                      %s \
```
jpkrohling, line 132: When `u.config.oidc` isn't set, it will generate a command like:

```console
$ curl --request POST \
--silent \
 \
--url https://dex.dex.svc.cluster.local:5556/dex/token \
--header 'content-type: application/x-www-form-urlencoded' \
--data grant_type=password \
--data username=admin@example.com \
--data password=password \
--data client_id=test \
--data client_secret=ZXhhbXBsZS1hcHAtc2VjcmV0 \
--data scope="openid email" | sed 's/^{.*"id_token":[^"]*"\([^"]*\)".*}/\1/'
```

I confirmed that this works (at least with my curl).

jpkrohling:
This is now ready for review. @squat could you please take a look, or assign someone to it?

jpkrohling:
ping @squat

jpkrohling:
ping @squat

metalmatze:
There are some conflicts here that need to be resolved. We also moved some of these components away from ksonnet.

jpkrohling:
PR rebased.

jpkrohling:
ping @metalmatze, @squat

tests/e2e.sh
```diff
@@ -14,6 +14,19 @@ kind() {
...
+    $KUBECTL create ns dex || true
+    $KUBECTL apply -f tests/manifests/observatorium-xyz-tls-dex.yaml
+
+    # in a real OpenShift cluster, this configmap is automatically populated with the service-ca
+    # upon creation. For the e2e tests, we need to manually populate it.
+    $KUBECTL create ns observatorium || true
+    cp environments/dev/manifests/test-ca-tls.yaml environments/dev/manifests/test-ca-tls.yaml.orig
+    cp tests/manifests/test-ca-tls.yaml environments/dev/manifests/test-ca-tls.yaml
+}
+
```
metalmatze, line 29: I don't think we need this entire new function here. It should be possible to integrate with the deploy function, simliar to what we have with in the operator repository: https://github.com/observatorium/operator/blob/master/tests/e2e.sh#L17-L34

metalmatze (changes requested):
Generally looking good :+1: 
One last comment on the e2e.sh, it might be worth trying to keep it close to the operator e2e and vise-versa. 
At last there's a conflict we need to solve.

components/dex.libsonnet
```diff
@@ -70,6 +76,24 @@
...
             },
+            {
+              name: 'tls',
+              secret: {
+                // this secret is created automatically by OpenShift when
+                // the service has the appropriate annotation
+                secretName: dex.config.tlsSecret,
+                items: [
+                  {
+                    key: 'tls.crt',
```
squat, line 87: we typically like to make the keys configurable in jsonnet (e.g. [0]) because not every certificate sercret creator uses the same key names. 

[0] https://github.com/observatorium/observatorium/blob/master/jsonnet/lib/observatorium-api.libsonnet#L204-L216

components/dex.libsonnet
```diff
@@ -109,6 +133,9 @@
     apiVersion: 'v1',
     kind: 'Service',
     metadata: {
+      annotations: {
+        'service.beta.openshift.io/serving-cert-secret-name': dex.config.tlsSecret,
```
squat, line 137: should this go in the upstream config? Or rather in openshift-specific down-stream configuration?

cc @metalmatze

tests/e2e.sh
```diff
@@ -14,6 +14,19 @@ kind() {
     ./kind create cluster
 }
 
+prepare-e2e() {
+    # in a real OpenShift cluster, this secret is automatically created and populated
+    # based on the annotation in the dex service. For tests, we need to manually create this.
+    $KUBECTL create ns dex || true
+    $KUBECTL apply -f tests/manifests/observatorium-xyz-tls-dex.yaml
```
squat, line 21: I think this should instead go in the `dex()` function.

tests/e2e.sh
```diff
@@ -14,6 +14,19 @@ kind() {
     ./kind create cluster
 }
 
+prepare-e2e() {
+    # in a real OpenShift cluster, this secret is automatically created and populated
+    # based on the annotation in the dex service. For tests, we need to manually create this.
+    $KUBECTL create ns dex || true
```
squat, line 20: this already happens in the `dex()` function; can we deduplicate this?

environments/dev/manifests/test-ca-tls.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    service.beta.openshift.io/inject-cabundle: "true"
+  name: test-ca-tls
```
squat, line 6: I find this a bit odd. Why are we distributing this as part of the dev manifests?

tests/e2e.sh
jpkrohling: +1, I remember facing a mismatch in the steps in some earlier iterations, but looks like it doesn't exist anymore. I removed this and it works with a single `deploy` function.

components/dex.libsonnet
jpkrohling: This has no side-effect for non-openshift deployments. We have a similar case in the Jaeger Operator where we add Istio and Linkerd annotations to some deployments, even though we might not have any in the cluster (and never both at the same time). The extra complexity isn't worth it, IMO.

jpkrohling: Just noticed you mentioned a "downstream configuration". Is it in this repo? It might indeed be easy enough to define this annotation there.

tests/e2e.sh
```diff
@@ -14,6 +14,19 @@ kind() {
...
+    $KUBECTL apply -f tests/manifests/observatorium-xyz-tls-dex.yaml
+
+    # in a real OpenShift cluster, this configmap is automatically populated with the service-ca
```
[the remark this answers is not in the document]
jpkrohling: Comment removed.

components/dex.libsonnet
squat: exactly, it's not in this repo, but it is also jsonnet and it inherits this config. We have all of our openshift stuff there, so let's be consistent and keep this repo as agnostic as possible to openshift.

components/dex.libsonnet
jpkrohling: Done

environments/dev/manifests/test-ca-tls.yaml
jpkrohling: You mean, a config map that contains the service CA that signed dex' cert, or a config map with `test` in the name?

squat: a configmap with test in the name; this is indicates to me that it's not supposed to be deployed to production, yet out production infrastructure will inherit this.

jpkrohling: The configmap name is derived from the client_id for the auth, which is currently set to `test`.

jpkrohling:
PR is ready for another review round. The only open point is about the name for the secret containing the CA that signed the tenant's token endpoint (tenant `test`).

components/dex.libsonnet
```diff
@@ -6,13 +6,20 @@
     image: error 'must provide image',
     version: error 'must provide version',
     namespace: error 'must provide namespace',
+    tlsSecret: '%s-tls' % [dex.config.name],
+    tlsCertMapEntry: 'tls.crt',
```
squat, line 10: Nit: can we call these: `tlsCertKey` and `tlsKeyKey` respectively? These fields do not belong to a ConfigMap, and thus having `Map` in the name is confusing. Additionally, using `Entry` sounds awkward to me since we don't use this terminology anywhere. They double `Key` is clunky but it is at least consistent with our other repos: https://github.com/observatorium/observatorium/blob/master/jsonnet/lib/observatorium-api.libsonnet#L222

squat (commented):
Looks good! Just one nit and a rebase. Then we can merge :rocket:

components/dex.libsonnet
jpkrohling: Alright, I used "Entry" because `KeyKey` sounded really wrong, but I agree with your assessment.

jpkrohling:
There was a change in master that removed the `up` component, in favor of externalizing it in its own component. This PR now depends on the changes to the up job being merged: observatorium/up#30.

jpkrohling:
This is ready for review again.

components/dex.libsonnet
```diff
@@ -7,13 +7,20 @@ local defaults = {
   image: error 'must provide image',
   version: error 'must provide version',
   namespace: error 'must provide namespace',
+  tlsSecret: '%s-tls' % [defaults.name],
+  tlsCertMapEntry: 'tls.crt',
```
squat, line 11: Can we still change the names of these fields?

jpkrohling: Ouch, sorry! Yes, absolutely. I focused so much on the rebase that I forgot this one :-)

jpkrohling: PR updated.

squat (approved):
awesome, thanks!!!!!!11111

PR #351, opened 2020-10-26
Support nodeSelector

clyang82:
Support nodeSelector for `memcached` and `loki`

/assign @squat @periklis


metalmatze:
For now, I want to see where https://github.com/observatorium/operator/pull/21 is going. If there are other reasons to have this in the future, then I'm happy to discuss it in the future again.

PR #354, opened 2020-10-29
Bump loki 1.6.1 to 2.0.0

periklis:
This is a minor PR with major impact as it is addressing the bump of Loki from 1.6.1 to 2.0.0. Major points addressed by upstream regarding boltdb-shipper:
- https://github.com/grafana/loki/pull/2770
- https://github.com/grafana/loki/pull/2601

PR #355, opened 2020-10-29
Add loki compactor component

periklis:
This PR addresses the missing loki compactor component for a boltdb-shipper only setup. The compactor takes the following duties formerly executed by queriers:
- Deduplicate index entries
- Optimize index file sizes for faster uploads from/downloads to S3 shared storage.

Further important characteristics introduced by the compactor component:
- Makes the index write cache obsolete, as this duty is part of the compactor.
- Makes use of PVCs as a local performance optimization when compacting downloaded indices from S3.

Depends on: #354 (Loki 2.0.0)

### DIFF
--- a/components/loki.libsonnet
+++ b/components/loki.libsonnet
@@ -57,7 +57,7 @@ local k = (import 'ksonnet/ksonnet.beta.4/k.libsonnet');
     std.member(['distributor', 'ingester', 'querier'], component),
 
   local isStatefulSet(component) =
-    std.member(['ingester', 'querier'], component),
+    std.member(['compactor', 'ingester', 'querier'], component),
 
   local newLokiContainer(name, component, config) =
     local deployment = k.apps.v1.deployment;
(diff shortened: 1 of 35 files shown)


components/loki.libsonnet
```diff
@@ -208,6 +208,13 @@ local k = (import 'ksonnet/ksonnet.beta.4/k.libsonnet');
     service.mixin.metadata.withLabels(commonLabels),
 
   components:: {
+    compactor: {
+      withReadinessProbe: true,
+      resources: {
+        requests: { cpu: '100m', memory: '100Mi' },
+        limits: { cpu: '200m', memory: '200Mi' },
```
blockloop, line 215: Perhaps this is my Thanos experience coming out, but don't you think that 200Mi memory limit for the compactor is way too small?

blockloop: I'm not familiar with how the Loki compactor works, but if it's _anything_ like the Thanos compactor I'm used to it's going to need a lot more memory.

periklis: In general the requests/limits here are **not** sane defaults, because we are using only a free-tier CircleCI instance. They are set to such low values for E2E testing only.

components/loki.libsonnet
```diff
@@ -388,6 +402,9 @@ local k = (import 'ksonnet/ksonnet.beta.4/k.libsonnet');
     config+:: {
       replicas: error 'must provide replicas per component',
     },
+
+    assert l.config.replicas.compactor == 1 : 'Only one replica for compactor component allowed',
```
kakkoyun, line 406: ❤️ Loved it, we should have it for Thanos compactor conf as well.

kakkoyun (approved):
LGTM 🚀

PR #359, opened 2020-11-02
Start using functions of kube-thanos

metalmatze:
Refactoring to use functions that kube-thanos exposes going forward.

### DIFF
--- a/environments/base/manifests/thanos-receive-default-service.yaml
+++ b/environments/base/manifests/thanos-receive-default-service.yaml
@@ -6,7 +6,7 @@ metadata:
     app.kubernetes.io/instance: observatorium-xyz
     app.kubernetes.io/name: thanos-receive
     app.kubernetes.io/part-of: observatorium
-    app.kubernetes.io/version: master-2020-08-12-70f89d83
+    app.kubernetes.io/version: v0.16.0
     controller.receive.thanos.io/hashring: default
   name: observatorium-xyz-thanos-receive-default
   namespace: observatorium
--- a/environments/base/manifests/thanos-receive-default-service.yaml
+++ b/environments/base/manifests/thanos-receive-default-service.yaml
@@ -15,13 +15,13 @@ spec:
   ports:
   - name: grpc
     port: 10901
-    targetPort: grpc
+    targetPort: 10901
   - name: http
     port: 10902
-    targetPort: http
+    targetPort: 10902
   - name: remote-write
     port: 19291
-    targetPort: remote-write
+    targetPort: 19291
   selector:
     app.kubernetes.io/component: database-write-hashring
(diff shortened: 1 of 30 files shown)


environments/base/manifests/thanos-receive-default-service.yaml
```diff
@@ -15,13 +15,13 @@ spec:
   ports:
   - name: grpc
     port: 10901
-    targetPort: 10901
+    targetPort: grpc
```
kakkoyun, line 18: Let's be consistent with target ports and use the actual port number not the name or vice versa.

environments/base/manifests/thanos-store-shard0-service.yaml
```diff
@@ -15,10 +15,10 @@ spec:
   ports:
   - name: grpc
     port: 10901
-    targetPort: 10901
+    targetPort: grpc
```
kakkoyun, line 18: ditto

environments/dev/manifests/thanos-receive-default-service.yaml
```diff
@@ -15,13 +15,13 @@ spec:
   ports:
   - name: grpc
     port: 10901
-    targetPort: 10901
+    targetPort: grpc
```
kakkoyun, line 18: ditto

environments/dev/manifests/thanos-store-shard0-service.yaml
```diff
@@ -15,10 +15,10 @@ spec:
   ports:
   - name: grpc
     port: 10901
-    targetPort: 10901
+    targetPort: grpc
```
kakkoyun, line 18: ditto

kakkoyun (commented):
Looks great.
My only comment concerning ports, we should be consistent and use only name or port number for target ports.

environments/base/manifests/thanos-receive-default-service.yaml
metalmatze: That's actually a kube-thanos change and I made this after looking at the Kubernetes docs on Services.
I guess we want the port _number_ for both port and targetPort then?

kakkoyun: Yes, I believe it's simpler and straightforward. However, if it's the _recommended_ way of doing it then let's use it everywhere.

metalmatze: It's not very clear and I feel like there's no definite answer. However, now I see that the docs say:

> Note: A Service can map any incoming port to a targetPort. By default and for convenience, the targetPort is set to the same value as the port field.

I'll move to `port == targetPort` then :man_shrugging: :smile:

kakkoyun (approved):
🥇 LGTM

metalmatze:
nice

PR #364, opened 2020-11-10
Remove unreferenced files within repo

kakkoyun:

To keep things clean, this PR removes unused files from this repo. The ones that are referenced we'll be moved to downstream.

cc @metalmatze @periklis


periklis:
@kakkoyun ` components/loki-caches.libsonnet` is not unused. It is actively used in our production configuration. I intend to move most of the production configuration here and leave only the SRE parts in gitlab.

kakkoyun:
@periklis We discussed it in the team and on the contrary, we decided to move the production-related configuration to GitLab (it is what I refer as downstream). I will move that file. This repo and the Gitlab repo will be orthogonal to each other. We're going to maintain this one for operator work and community edition. The goal is to reduce the number of steps needed to ship a change to live environments. We intend to use this repo for shared components or libraries.

squat (approved):

brancz:
Kind of the point of this repo is to share a production-ready configuration. Do we think we could entertain the idea of a more production-ready setup within this repo, even if red hat doesn't end up using it?

metalmatze:
Right. I generally agree and we often said we want this. This is merely an effort to reduce complexity.
We should discuss again on how to still make most of what actually gets run public upstream.

kakkoyun:
And we're going to continue to maintain this repo and share production-ready configuration.

PR #371, opened 2020-11-23
[loki] Add withDataReplication to enable replication config

periklis:
This PR exposes the `replication_factor` configuration option via `withReplicationFactor`. This way it enables overrides via OpenShift template parameters. Later cannot be replaced inside config map data blocks. Thus, the present implementation provides an override via container parameters.


kakkoyun (approved):
LGTM 🥇

bwplotka (approved):
:+1:

PR #376, opened 2020-12-04
Refactor Loki generators to use new functional pattern

kakkoyun:


kakkoyun:
@metalmatze @periklis This ready for review. I tried to keep changes to generated files minimal. Also, I tried to stick to the existing patterns. I'm open to suggestions tell me what you think.

components/loki.libsonnet
```diff
@@ -1,28 +1,62 @@
...
+  version: error 'must provide version',
+  image: error 'must provide image',
+  replicas: error 'must provide replicas',
+  objectStorageConfig: error 'must provide object storage config',
+  queryConcurrency: 32,
+  queryParallelism: 32,  // Defaults to queryConcurrency because single query-frontend replica
+  ports: {
+    gossip: 7946,
+  },
+  replicationFactor: 0,
```
periklis, line 18: The default should be always 1.

components/loki.libsonnet
```diff
@@ -1,28 +1,62 @@
...
+  namespace: error 'must provide namespace',
+  version: error 'must provide version',
+  image: error 'must provide image',
+  replicas: error 'must provide replicas',
+  objectStorageConfig: error 'must provide object storage config',
+  queryConcurrency: 32,
+  queryParallelism: 32,  // Defaults to queryConcurrency because single query-frontend replica
+  ports: {
+    gossip: 7946,
+  },
```
periklis, line 17: What is the reason to expose this port per config?

components/observatorium.libsonnet
```diff
@@ -118,9 +138,8 @@ local api = (import 'observatorium/observatorium-api.libsonnet');
     } + {
       ['thanos-' + name]: obs.thanos.manifests[name]
       for name in std.objectFields(obs.thanos.manifests)
-    } + if std.length(obs.config.loki) != 0 then {
+    } + {
       ['loki-' + name]: obs.loki.manifests[name]
       for name in std.objectFields(obs.loki.manifests)
-    }
-    else {},
+    },
```
periklis, line 144: Is the `if std.length(obs.config.loki) != 0` still ensured? IIRC, this check was added by ACM folks to deploy observatorium w/o loki.

periklis (dismissed):
Looks in general good to me.

components/observatorium.libsonnet
kakkoyun: We now encourage an alternative way to build up the configuration in the downstream repo. Rather than inheritance, we want them to compose theirs on setup utilizing functions. So this should be fined.

components/loki.libsonnet
kakkoyun: I don't actually get the question, can you maybe rephrase? This is not exposed. This just a top-level configuration that we utilize whenever we need to access the ports.

periklis: My intention was to say, that folks using the new pattern should be able to override this port.

periklis (commented):
Awesome work! :rocket:

components/loki.libsonnet
kakkoyun: 👍 Yes, it's totally possible and cleaner.

components/loki.libsonnet
```diff
@@ -1,29 +1,329 @@
-{
-  local loki = self,
+// These are the defaults for this components configuration.
+// When calling the function to generate the component's manifest,
+// you can pass an object structured like the default to overwrite default values.
+local defaults = {
+  local defaults = self,
+
+  name: error 'must provide name',
```
metalmatze, line 7: It should be fine to set a default name here.

metalmatze (commented):
Overall going in a good direction. I think there can be still a clearer separation between the individual components of Loki and thus splitting the defaults object quite drastically. Nothing that needs to be done in this PR though.

PR #381, opened 2021-03-04
Extend configurability and update kube-thanos

brancz:
This exposes some more configurability for the thanos component. As you can see (by `make generate` not causing anything to change), this does not change any of the output, it's just a refactoring exposing these configuration options. It also updates and changes kube-thanos "version" from "master" to "main".

As a side note, this "thanos" component is now so generic, it should probably be in kube-thanos.

@kakkoyun


kakkoyun (approved):
The changes look good, they're easier on the eyes. 🚀 
Tests are failing but nothing has changed functionally, assuming it's just a CI glitch. Rerunning.

PR #384, opened 2021-03-04
components/thanos: Correctly render store ServiceAccount

brancz:
Follow up for #383. I missed to actually render the store ServiceAccount.

@kakkoyun


kakkoyun (approved):
🥇 Oops I've missed that as well

PR #386, opened 2021-03-05
Allow piping through all configuration and configuring all ServiceMonitors

brancz:
As in the title.

@kakkoyun

### DIFF
--- a/components/thanos.libsonnet
+++ b/components/thanos.libsonnet
@@ -22,6 +22,7 @@ local defaults = {
   stores+: {
     shards: 1,
     storage: '50Gi',
+    serviceMonitor: false,
   },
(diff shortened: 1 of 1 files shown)


brancz:
I think a nice follow up would be to have each component have the `.manifests` field, which can then be recursively built, so we don't need to leak the configuration into rendering like we do here.

kakkoyun (approved):
💯

kakkoyun:
> I think a nice follow up would be to have each component have the `.manifests` field, which can then be recursively built, so we don't need to leak the configuration into rendering like we do here.

I love the idea. We do it for major building blocks like Thanos, Loki. It'd be best to have it for each component.

PR #387, opened 2021-03-06
components/memcached: Use non-root and own ServiceAccount

brancz:
Tested on a cluster, and works beautifully :) 

@kakkoyun


kakkoyun (approved):
🚀  LGTM

PR #388, opened 2021-03-10
Simplify the logic

clyang82:
we already have logic if thanos.config.receive.serviceMonitor is true then create thanos.receivers.serviceMonitor. so remove the logic here. another reason is operator depends on it - https://github.com/observatorium/operator/pull/49

/assign @squat 


brancz:
Can you explain why this is needed for https://github.com/observatorium/operator/pull/49 ? If I'm not mistaken this was needed for me to make this work for the default configuration (exactly why CI is failing).

clyang82:
This PR is not valid anymore. Thanks.

PR #389, opened 2021-03-12
expose fields

clyang82:
expose fields in conf, including `volumeClaimTemplate`, etc. in the past, we allow to configure `storageClassName` in `volumeClaimTemplate`. that is real case in production environment.

partial fix for https://github.com/observatorium/operator/issues/50

/assign @squat @kakkoyun 

### DIFF
--- a/components/thanos.libsonnet
+++ b/components/thanos.libsonnet
@@ -21,8 +21,18 @@ local defaults = {
   }],
   stores+: {
     shards: 1,
-    storage: '50Gi',
+    volumeClaimTemplate: {
+      spec: {
+        accessModes: ['ReadWriteOnce'],
+        resources: {
+          requests: {
+            storage: '50Gi',
+          },
+        },
+      },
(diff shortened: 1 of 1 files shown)


brancz:
I think this generally makes sense, but we should enforce that `accessModes: ['ReadWriteOnce']` is always the case, as if this is not the case then there can be various weird situations that the storage was not designed for.

clyang82:
@brancz Thanks for your comment. Do you mean that we should enforce `accessModes: ['ReadWriteOnce']` in operator side?

clyang82:
Agree with your point. I can have another PR in kube-thanos side to enforce it. But I think this PR does not depend on that PR. Can we merge this one firstly? Thanks.

brancz:
Yep, makes sense!

brancz (approved):

clyang82:
Thanks @brancz

PR #391, opened 2021-03-15
Set requests for up test

clyang82:
If a Container specifies its own CPU limit, but does not specify a CPU request, Kubernetes automatically assigns a CPU request that matches the limit. Set requests for up smaller than limit to ensure the tests can be passed in https://github.com/observatorium/operator/pull/52.

### DIFF
--- a/tests/main.jsonnet
+++ b/tests/main.jsonnet
@@ -70,6 +70,11 @@ local metricsConfig = {
       memory: '128Mi',
       cpu: '500m',
     },
+    requests: {
+      memory: '128Mi',
+      cpu: '50m',
+    },
+
   },
   endpointType: 'metrics',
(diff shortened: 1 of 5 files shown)


clyang82:
/assign @squat @brancz 

Please help to review this PR. Thanks.

squat:
This is also true for memory requests/limits, right? Should we explicitly decrease the memory request also?

clyang82:
we have sufficient memory in test cluster. I think we can continue to use `128Mi`.
```
  Resource           Requests      Limits
  --------           --------      ------
  cpu                1500m (75%)   1700m (85%)
  memory             1074Mi (14%)  1774Mi (23%)
```

kakkoyun (approved):
💯  LGTM

squat:
Ah i see, the problem was that there wasn't enough CPU to schedule during tests

PR #402, opened 2021-05-02
Renamed environments to examples; Cleaned not-up-to-date docs.

bwplotka:


onprem (approved):
Let;s change the link of Get Started button to the getting started guide (https://github.com/observatorium/observatorium/blob/main/website/layouts/index.html#L10)

We need to change the default URL of menu here as well (https://github.com/observatorium/observatorium/blob/main/website/config/_default/menus.toml#L15). Also I think now we don't have anything under "Architecture" menu, so we can remove the entry from the same file as well.

bwplotka:
Thanks! let's merge and iterate.

RiRa12621:
I'm missing the multitenancy model in the architectural diagrams, if that is something you want to keep in the back of your head for the next revision.

PR #411, opened 2021-05-21
Update loki.libsonnet to make compute resources configurable for Loki

4n4nd:
This will make compute resources configurable for Loki components.
The resource section will be duplicated over all the Loki components, but that is still better than no control over the resources.

Also updated the default compute resource limits to:
```yaml
resources: {
    requests: { cpu: '100m', memory: '100Mi' },
    limits: { cpu: '2000m', memory: '2000Mi' },
  }
```

Related https://github.com/observatorium/operator/issues/71


periklis (changes requested):
The change set provided breaks clearly intended functionality when running Loki on production environments. Each component is quite different on the work it has to do and thus requires separate request/limits. Please consider the different settings used in various production environments:
1. RHOBS: https://github.com/rhobs/configuration/blob/main/services/observatorium-logs-template.jsonnet#L29
2. Grafana: https://github.com/grafana/loki/tree/main/production/ksonnet/loki

4n4nd:
@periklis this is ready for review now :+1:

configuration/components/observatorium.libsonnet
```diff
@@ -102,6 +102,58 @@ local api = (import 'observatorium-api/observatorium-api.libsonnet');
...
+        requests: {
+          memory: '100Mi',
+          cpu: '100m',
+        },
+        limits: {
+          memory: '200Mi',
+          cpu: '200m',
+        },
+      },
+    },
```
periklis, line 156: The above change looks like duplicating defaults which are already included in [loki.libsonnet](https://github.com/observatorium/observatorium/blob/main/configuration/components/loki.libsonnet#L32). AFAICS there is a small bug introduced by @kakkoyun's refactoring [here](https://github.com/observatorium/observatorium/blob/main/configuration/components/loki.libsonnet#L20). 

My suggestion is to remove [loki.libsonnet#L20-21]() and fix the [loki.libsonnet#L321](https://github.com/observatorium/observatorium/blob/main/configuration/components/loki.libsonnet#L321) and [loki.libsonnet#L388](https://github.com/observatorium/observatorium/blob/main/configuration/components/loki.libsonnet#L388) to use the `components...resources` fields.

periklis (changes requested):

PR #412, opened 2021-05-31
A few quick grammar updates

brmurphy:


squat (approved):
Thanks :)) this whole doc could use a little bit of proofreading. This is a step in the right direction

PR #413, opened 2021-06-03
components: Decouple thanos & loki secrets from minio

brancz:
I'd like to reuse the minio component for something unrelated to thanos and loki. Regardless I think it didn't make sense for the secrets that are specific to thanos and loki to leak into the minio component. A next step might be to pass the secrets into the thanos/loki components, however, I didn't want to do that as I think that might be too opinionated, as many people prefer configuring secrets via vault or some other custom volume that is not necessarily a Kubernetes secret, so I didn't want to force one opinion or the other even stronger than we already are.

@kakkoyun @metalmatze @onprem @periklis

### DIFF
--- a/configuration/components/minio.libsonnet
+++ b/configuration/components/minio.libsonnet
@@ -3,7 +3,11 @@
 // you can pass an object structured like the default to overwrite default values.
 local defaults = {
   local defaults = self,
+
   namespace: error 'must provide namespace',
+  buckets: error 'must provide buckets',
+  accessKey: error 'must provide accessKey',
+  secretKey: error 'must provide secretKey',
 
   commonLabels:: { 'app.kubernetes.io/name': 'minio' },
 };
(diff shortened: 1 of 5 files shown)


configuration/components/minio.libsonnet
```diff
@@ -5,6 +5,9 @@ local defaults = {
   local defaults = self,
   namespace: error 'must provide namespace',
 
+  buckets: ['thanos', 'loki'],
```
squat, line 8: Should we do the same as above here? Make these values empty and make them required?

brancz: Yeah, makes sense, I'll do the same for access and secret key.

brancz: done! :)

configuration/examples/dev/main.jsonnet
```diff
@@ -136,8 +138,38 @@ dev.manifests +
...
+      namespace: dev.config.namespace,
+    },
+    stringData: {
+      'thanos.yaml': |||
+        type: s3
+        config:
+          bucket: thanos
+          endpoint: %s.%s.svc.cluster.local:9000
+          insecure: true
+          access_key: minio
```
kakkoyun, line 155: just a nit: We can use variables for access_key and secret_key. They are referred more in a couple of places.

kakkoyun (approved):
LGTM :rocket: 

Just a non-blocker nit.

configuration/examples/dev/main.jsonnet
brancz: yeah I wanted to keep the changes minimal and improve bit by bit, it's already better than it was before and I think this can be done as follow ups

PR #419, opened 2021-06-30
Adds rules API proposal

ianbillett:
Signed-off-by: Ian Billett <ibillett@redhat.com>


docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,94 @@
...
+
+* **Related Tickets:**
+  * `<JIRA, GH Issues>`
+
+* **Other docs:**
+  * `<Links…>`
+
+## TL;DR
+
+We propose implementing an API that allows multi-tenant users to create, read, update and delete Prometheus alerting and recording rules.
```
squat, line 14: I think it's clearer to say that the API is multi-tenant as opposed to the users
```suggestion
We propose implementing a multi-tenant API that allows users to create, read, update and delete Prometheus alerting and recording rules.
```

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Why
+
+The single biggest source of frustration from internal Red Hat users of Observatorium right now is the time it takes for changes to their rules to appear in our production infrastructure.
+
+### Pitfalls of the current solution
+
+#### Implicit deploy cycle dependency
+
+Currently, the recording rules deployed into Observatorium are configured in the same repository as the configuration for the service itself. This implicitly ties the deploy-cycle of Prometheus rules to the deploy-cycle of Observatorium itself. If a new set of rules needs to be deployed, our team is required to roll-out our production infrastructure.
```
squat, line 24: Maybe we should clarify that this is our practice within red hat vs how it is done for "Observatorium". It's a result of the fact that the only entry point into modifying recording rules is the jsonnet code in https://github.com/observatorium/observatorium

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,94 @@
...
+
+As the number of tenants we support and users we serve increases, the team responsible for the Observatorium installation is required to roll out production more frequently in order to satisfy user requests. This becomes an increasing impediment to user experience as the size of Observatorium increases.
+
+#### Shared configuration repositories
+
+For rules to be rolled out, users need the ability to raise pull requests against the Observatorium repositories that contain our source configuration. This works fine when the two team co-exist within the same organisation, but this becomes a blocking issue when the two teams share a much more restricted (i.e. inter-company) trust boundary.
```
squat, line 30: Nits + changing to e.g. since it's one example of such a trust boundary.

```suggestion
For rules to be rolled out, users need the ability to raise pull requests against the Observatorium repositories that contain our source configuration. This works fine when the two teams co-exist within the same organisation, but it becomes a blocking issue when the two teams share a much more restricted (e.g. inter-company) trust boundary.
```

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,94 @@
...
+* Users can update their Prometheus rules without intervention from the operating team.
+
+## Non-Goals
+
+* Rate-limiting or accounting of user resources.
+* Define or implement the underlying database technology.
+
+## How
+
+We propose to solve the above problem by implementing an API that allows multi-tenant users to create, read, update and delete Prometheus alerting and recording rules.
```
squat, line 45:
```suggestion
We propose to solve the above problem by implementing  amulti-tenant  API that allows users to create, read, update and delete Prometheus alerting and recording rules.
```

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,94 @@
...
+
+Thanos Ruler operates on the rule configuration provided by users, this is consumed from a file in Ruler's local data directory. Therefore, this configuration that we ingest via the API needs to be replicated into the local data directory of the Thanos Ruler instances. This is achieved with the [thanos-ruler-syncer](https://github.com/observatorium/thanos-rule-syncer), an application that calls the API above and writes the results into a location that Thanos Rule can access.
```
squat, line 47: Pedantic nit: not necessarily the data directory, as that is where TSDB data lives, but rather any directory configured using the `--rule-file` flag

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,94 @@
...
+
+This solution must be backwards compatible with rule configuration currently defined in our repository. That is to say, to ease the transition from our current setup to this new API, Thanos Ruler should be able to operate on Rules both from the API and Supplied as configuration.
```
squat, line 49: Backwards compatibility sounds like a red hat problem rather than an Observatorium concern, no?

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: I love a pedantic nit :)

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Backwards compatibility is - but  the real question here is then: should Observatorium support:
* **only** API driven user rules
* **both** API and config-file user rules?

API-only is easier to support, but I would err on the side of supporting both as config-file are what the upstream communities use. WDYT?

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: good call - updated

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
squat: hm ok so backwards compatibility for the API means not doing anything, right? Because the API is currently totally agnostic to rules. So we are talking about maintaining backwards compatibility in the observatorium jsonnet and how it automatically configures the thanos ruler?
I am not really sure we should be investing effort in maintaining this ad-hoc jsonnet _API_ since it's not formalized or versioned. It was really a hack to have a way of provisioning rules before we had a proper API. The jsonnet interface has been evolving constantly and changes every time we update kube-thanos :p if we want to, we could go the route of ksonnet and version our jsonnet APIs too, but I wouldn't want this to become a bottleneck right now.

I think that if we want to provide a story for administrators to have a read-only, configuration-based approach to managing recording rules for tenants, we should be opinionated and recommend using a rules-file per tenant, e.g. as implemented in the PR, in conjunction with the thanos-rule-syncer. This way the rules are statically provisioned and managed and we can focus on maintaining a single stable API + configuration file syntax.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,94 @@
...
+
+We propose to solve the above problem by implementing a multi-tenant API that allows users to create, read, update and delete Prometheus alerting and recording rules.
+
+Thanos Ruler operates on the rule configuration provided by users, this is consumed from a file in local file defined by Ruler's `--rule.file` flag. Therefore, this configuration that we ingest via the API needs to be replicated into the local data directory of the Thanos Ruler instances. This is achieved with the [thanos-ruler-syncer](https://github.com/observatorium/thanos-rule-syncer), an application that calls the API above and writes the results into a location that Thanos Rule can access.
+
+This solution must be backwards compatible with rule configuration currently defined in our repository. That is to say, to ease the transition from our current setup to this new API, Thanos Ruler should be able to operate on Rules both from the API and Supplied as configuration.
+
+### How should we persist user's rule configuration?
+
+In the first iteration of this API, we will persist the rule configuration to a local directory that is backed by a `PersistentVolumeClaim`.
```
squat, line 53: I don't think implementing this API by storing request data in files is a viable approach for Observatorium. Right now, the Observatorium API is run as a horizontally scalable service, meaning we can have N replicas all serving requests. If we allow users to write data to disk via the Observatorium API, then suddenly each replica of the service will have a different view of the rules since the data in each Pod's PersistentVolumeClaim will be different. Using files to implement an API backend only works for read-only APIs that can be configured statically during deployment. For writable APIs we really need to store the data in an HA location to which all replicas of the Observatorium API have access, e.g. a database.

ianbillett: Ok - that makes sense. The ability to scale horizontally is then another requirement that we need to explicitly call out :+1:

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Ok gotcha, slight misunderstanding there - my intention there was your second point which is a much more eloquent way of putting it.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,100 @@
...
+We propose to solve the above problem by implementing a multi-tenant API that allows users to create, read, update and delete Prometheus alerting and recording rules.
+
+Thanos Ruler operates on the rule configuration provided by users, which is consumed from a file defined by Ruler's `--rule.file` flag. Therefore, this configuration that we ingest via the API needs to be replicated into the local data directory of the Thanos Ruler instances. This is achieved with the [thanos-ruler-syncer](https://github.com/observatorium/thanos-rule-syncer), an application that will call the API above and write the results into a location that Thanos Rule can access.
+
+### How should we persist user's rule configuration?
+
+For the Observatorium API to satisfy our requirement of scaling horizontally with the number of user requests, we need to use a persistent storage layer for our user's recording rules.
+
+TODO
+* Object storage is already used in Thanos. Users will already have this infrastructure provisioned.
```
ianbillett, line 55: @squat I'll need a hand in filling out these sections - happy to do either over meet or happy for you to fill this in async

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,100 @@
...
+
+### How will users perform authentication and authorization?
+
+TODO
+* Probably the same way that Telemeter client and server authentication works?
+* How will users provision new API keys / secret material?
```
ianbillett, line 61: ditto as above

ianbillett:
@squat and I chatted earlier - we're not sure what is the best approach in DB vs object storage so we will lay out the decision and get the team's input.

periklis (changes requested):
Awesome proposal! Thank you so much for taking the time to step back and bring this in verbatim and enable some fruitful discussion. :rocket: 

As a co-maintainer of the logging routes on the observatorium-api, the approach taken to persist and sync the rules to the individual metrics (thanos-ruler) ruler works as expected the same with loki (cortextool sync). Thus, I don't see any obstacles specifically for Loki. Maybe users need to declare which rules are written in `PromQL` and which in `LogQL`. Maybe by amending rules in-flight with the signal as a label on the persisted rules?! WDYT

Nevertheless the more I read about this proposal and the underlying proof-of-concept PR, I have concerns adding the current implementation into the `observatorium-api`. The single purpose of the api is to be a super performant and reliable API-Gateway taking care of tenant authn/autz. Bonding a database to this api will make rollouts more cumbersome for such a crucial component. IMHO, the rules API screams for a separate microservice `observatorium-rulez` that handles persistence and invokes/triggers/whatever the syncs with `thanos-rules-sync` and `cortextool sync`. :warning: 

Saying this  I believe the `observatorium-api` does a great job as of today because it uses extensively delegation to other services (e.g. authn to `OIDC` or authz to `opa-*` sidecars). This approach served us well and keeps maintenance of the API low. In addition scale out of the api is as simple as increasing replicas. Adding a database makes considerations on replica increases more cumbersome.

ianbillett:
Comment from @bwplotka from our slack
> 3 cents from me. I think it would be amazing if Rules API is actually separate service on top of Observatorium. This is to make sure we don't make it mandatory for Postgres to be installed and installed someone could use Kube API and operator for it

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+* **Related Tickets:**
+  * `<JIRA, GH Issues>`
+
+* **Other docs:**
+  * Original [Prometheus Rules in Observatorium API design doc](https://docs.google.com/document/d/1F9Cw6I4qFs__0Dcm19xvxJqRBCAGQBBggYg_PQSV_-g/edit#heading=h.cp0jmcyfj3) (internal Red Hat link)
+
+## TL;DR
+
+We propose implementing a multi-tenant API that allows users to create, read, update and delete Prometheus recording rules.
```
yeya24, line 14: https://github.com/observatorium/observatorium/pull/419/files#diff-8de1f9b976bdab7fdd01a236d8fed2c2be98ee58ea965d33567ba147fe07479eR45 mentions both `alerting` and `recording` rules. But the title and here only mentions `recording` rules, so is `alerting` rule also included in the proposal?

ianbillett: Great question @yeya24 :raised_hands:

The original design doc included both alerting and recording rules in scope. That makes sense given that recording and alerting rules can be freely mixed in a given `rule_files`. 

However, without a corresponding set of features to support user-supplied alertmanager routing configuration instances (which is definitely out of scope of this document) - any alerting rules we ingest will sit idle.

My preference here is to scope this design doc to recording rules only, and explicitly reject any rule files that contain `alert:` keys with an appropriately clear error message to the caller.

ianbillett: I've added this to the non-goals section. Resolving.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Resolving - added explicitly as a requirement.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Answer - we will just use the same RBAC mechanism as we do already.

ianbillett:
Hmm the SVG sequence diagram is failing to render for me under `display rich diff` - but I'm confident it will work when merged. This is the diagram that is included:

image

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,238 @@
...
+
+As the number of tenants we support and tenants we serve increases, the team responsible for the Observatorium installation is required to roll out production more frequently in order to satisfy tenant requests. This becomes an increasing impediment to tenant experience as the size of Observatorium increases.
+
+#### Shared configuration repositories
+
+For rules to be rolled out, tenants need the ability to raise pull requests against the Observatorium repositories that contain our source configuration. This works fine when the two teams co-exist within the same organisation, but it becomes a blocking issue when the two teams share a much more restricted (e.g. inter-company) trust boundary.
+
+This is not a blocker for Red Hat's internal offering, however this is not inline with Observatorium's stated goal (TODO: insert link) of offering a SaaS-like monitoring solution.
```
matej-g, line 71: What link should be here?

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Goals
+
+* Tenants can update their Prometheus recording rules without intervention from the operating team.
+* Observatorium must be able to scale horizontally scale with the number of tenants.
```
matej-g, line 76:
```suggestion
* Observatorium must be able to scale horizontally with the number of tenants.
```

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,238 @@
...
+3. Asks Rule to reload its rules by calling the `/-/reload` endpoint.
+
+This will run as a sidecar to `Rule` and `-file` will be shared between the two containers.
+
+### How do we store tenant rules
+
+#### Service
+
+We will implement the rule storage backend as a separate service within Observatorium (i.e. not part of the API). This follows the established pattern of specifying backends in the API configuration and also maintains desirable properties of the API:
+* The API stays as a super-performant gateway that handles authentication and authorization but no application-specific logic.
```
matej-g, line 110: Optionally we could mention the 'do one thing and do it well' principle (from UNIX philosophy), which I think this design is a great example of.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,238 @@
...
+* Observatorium users are free to specify different storage backends, or no rule backend at all.
+
+The rule backend will be specified via the `--metrics.rules.endpoint` flag, and will route requests via the `/api/metrics/v1/{tenant}/rules` endpoint.
+
+NB: By scoping rules to the `/metrics` endpoint we are consciously creating room for future logging rules.
+
+Since we have explicitly de-scoped alerting rules from this proposal, the service will need to inspect the rules payload. If any rules are found to contain an `alert` field, this request should be rejected and the user returned a `400 Bad Request` error with a helpful message.
```
matej-g, line 117: I was wondering about this already, good job on answering questions before they can be asked :+1:. However, as an alternative, I was thinking whether accepting the mixed 'content' and giving the user warning about ignoring `alert`s would be more user-friendly. I guess it comes down whether we 'value' more user's convenience or user's level of awareness about what we are accepting (warning might as well be overlooked). I'm wondering about others' opinions.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,238 @@
...
+
+Ideally, we would define the rule storage backend API in OpenAPI format, so that it can easily be consumed by downstream services.
+
+#### Storage layer
+
+We will use a persistent storage mechanism that is external to the api deployment (i.e. no local storage). This enables us to satisfy our requirement of horizontally scaling the rule storage backend.
```
matej-g, line 123: nit: capitalizing API everywhere
```suggestion
We will use a persistent storage mechanism that is external to the API deployment (i.e. no local storage). This enables us to satisfy our requirement of horizontally scaling the rule storage backend.
```

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,238 @@
...
+
+Our goal here is to remove our team entirely from the control loop of updating Prometheus rules. We are trading off immediate term technical complexity against longer-term scalability and happier multi-tenancy operation.
+
+This option also does not address the trust-boundary issue identified above.
+
+### Per-tenant GitOps
+
+Another option would be for each tenant to commit their rule configuration into a git repository they control, and then we synchronise that into the right place in our infrastructure based on a URL they provide. This would eliminate the need for us to manage state on their behalf.
+
+This is a neat idea that avoids deploy dependencies, but has some drawbacks:
```
matej-g, line 227: Optional: I think in general having to manage additional git repo sounds already like enough of an overhead for the tenants, even without validation / secrets, so I find our actual proposal much more 'tenant-friendly'.

matej-g (commented):
Such a great job! And the diagram is a real :cherries: on top. I have just few minor comments.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,238 @@
...
+2. API performs authentication and authorization of the tenant then forwards the request to the rules storage backend.
+3. Rules storage backend performs validation of the payload, rejecting any requests found to contain `alert` rules, then stores the rule file in object storage at the path `metrics/rules/{tenant}/{file_name}.yaml`.
+4. Object storage returns success to the Rule Backend.
+5. Rule backend returns success to the API.
+6. API returns success to the Tenant.
+
+#### 2. Sync Rules to Rule
+1. Periodically, thanos-ruler-syncer queries the API for all rules for all tenants.
+   * Do we need to expose an endpoint to return all of the rules? Unclear at the moment.
+   * NB: We can definitely make some optimisations here. Suggestions welcome.
```
periklis, line 192: Why does the `thanos-ruler-syncer` require to poll the API to fetch all rules? Why not polling the `ruler-storage-backend`?

Requiring to poll the API for internal traffic looks like complicating the case w/o more benefits. Any internal client like the `thanos-rule-syncer` needs to work through authn/authz. I suggest to do so only if the syncer needs to scope queries per tenant.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: added in file

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Yes nice touch - will add that too :)

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: updated in doc

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
squat: This means that if you ever change the storage of your rules then you also need to change the rule-sycners to point at new APIs. This is an indication to me that you are too tightly coupling things that should be decoupled.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Hmm that is a very astute point :thinking: I'll add a section that dives into this decision in more detail and we can discuss there.

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Good point :) Added that in the list

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
ianbillett: Good point :raised_hands: Calling the API is a hangover for the original design doc - it doesn't go into details why but I suspect that is so that Observatorium components themselves are designed to work in non-kubernetes environments so we can't just assume that the `syncer` can call `rule-storage-backend` directly.

The API interface will be the same between the API & `rule-storage-backend` so it doesn't matter _too_ much. It is more a question for how this is configured in our reference kubernetes deployment of Observatorium.

I'll add a comment to that effect in the doc :+1:

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,271 @@
...
+
+##### 1. ETCD
+
+One option would be to use the Kubernetes' control plane as the backing data store for tenant's recording rules.
+
+Pros:
+* Lowest possible overhead. No external storage is required to be orchestrated.
+* ETCD data model well suited for recording rules i.e. YAML.
+* Storage permissions are managed as Kubernetes objects.
+* We could leverage Recording Rules CRDs and get validation for free (?)
```
bwplotka, line 168: yes
```suggestion
* We could leverage Recording Rules CRDs and get validation for free
```

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,271 @@
...
+
+Cons:
+* Implicitly ties Observatorium's implementation into a Kubernetes environment. Hard to back out.
```
bwplotka, line 171: Really? What if we do this in this addon/plugin format so user can back it up with RBMS if needed?

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,271 @@
...
+* Large blast radius. With a large number of tenants and recording rules, we could un-intentionally impact the Kubernetes control plane's performance.
+* ETCD has a limit of 1.5MB per key / value pair so we can't store large files (but Prometheus operator seems to do this ok?).
```
bwplotka, line 173: Yea it's fine for those. We can keep this item.

I would add another item: It's might be hard to replicate user configuration across multiple clusters / zones if we would have multi-zone SaaS for single tenant

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,271 @@
...
+
+##### 2. Object Storage
+
+Pros:
+* Users of Thanos project are likely to already have object storage configured and ready to use.
+* Recording rule data model is a good fit for object storage i.e. YAML.
+
+Cons:
+* Unclear consistency semantics depending on object storage provider i.e. in the case of lots of writes, who wins?
+* Requires tenants to manage another set of secret credentials.
```
bwplotka, line 183: Really? I though this storage could be behind our API as in this proposal so we would maintain credentials to storage

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,271 @@
...
+
+##### 3. RDBMS
+
+Another option we have considered is to provision and use a relational database i.e. Postgres.
+
+Pros:
+* Strong isolation and update semantics.
+* Enables us to use multi-dimensional
```
bwplotka, line 191: Enable us to do replication / multi-cluster scalability

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,271 @@
...
+Cons:
+* We don't currently run RDBMS, and don't have any experience doing so.
+* Significant infrastructure overhead - tenants required to BYO database.
+* Requires tenants to manage another set of secret credentials.
+
+##### Storage Decision
+
+This was discussed in the [Observatorium Community Meeting](https://docs.google.com/document/d/1jLvOH0Lllt-ShXVgWeeHBGPb3ZuagirabD_hbq41EKs/edit#heading=h.wxenjcrf4iee) on 2021-07-31.
+
+The consensus among the group was that object storage offered us the best combination of ease of use, familiarity for existing users, and also compatability with the required operations against tenants rules.
```
bwplotka, line 202: Nice! I like this idea, BUT I would also consider pros / cons I commented in. I think we missed important part where object storage is hidden by our API so we can change storage later and iterate

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
```diff
@@ -0,0 +1,271 @@
...
+
+We discussed that we will be likely to employ a RDBMS relatively soon for other features (i.e. RBAC), but we decided to cross that bridge when we came to it.
```
bwplotka, line 204: 👍🏽 YAGNI

docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
bwplotka: Actually your diagram has it all, just above items might be not clear about those

bwplotka (approved):
Amazing work @ianbillett I love it.

LGTM from my side, modulo comments, good work!

matej-g (approved):

ianbillett:
From observatorium communitry meeting - we should accept the alerting rules, but respond with a warning that we don't yet support alert routing.

PR #421, opened 2021-07-09
Add Authz Matchers API proposal

periklis:
The following design proposal provide an extension for the observatorium api to OPA authorizers. In particular this extension addresses the issue to refine queries to read endpoints with label matchers provided by the OPA authorizer. It is inspired by the work on tenant isolation. Although it goes one step further to handle dynamic label matchers provided by domain specific OPA authorizers in form of PromQL's `labels.Matcher` (See [prometheus/pkg/labels/matcher.go](https://github.com/prometheus/prometheus/blob/main/pkg/labels/matcher.go#L45)).

### DIFF
--- a/docs/proposals/drafts/20210709-opa-authorizer-labels-observatorium-api.md
+++ b/docs/proposals/drafts/20210709-opa-authorizer-labels-observatorium-api.md
@@ -1,0 +1,74 @@
+---
+toc: true
+title: 2021-07 Authz Matchers API
+---
+
+## Implement Authz Matchers API for OPA authorizers
+
+* **Owners**
+  * @periklis
+
+* **Related Tickets**
+  * [LOG-1513: Provide an opa-openshift authorizer for OpenShift Cluster Logging using Loki](https://issues.redhat.com/browse/LOG-1513)
+
+* **Other Docs**
+  * Cross-Tenancy proposal???
+
+## TL;DR;
+
+We propose an API that allows OPA authorizers to return label matchers which get enforced read endpoints by the observatorium api.
+
+## Why
+
+The proposed API addresses the issue to use domain-specific authorization knowledge to read queries. This enables a much richer isolation of read queries for the observatroium api.
+
+## Goals
+
+* Users can configure OPA authorizers to pass a set of label matchers back to the observatorium api.
+* The observatorium api can extend the label sets automatically for queries against the read endpoints for all supported observability signals.
+
+## Non-Goals
+
+* Provide a generic label matching engine inside the observatorium api for query result sets.
+
+## How
+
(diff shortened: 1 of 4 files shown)


docs/proposals/drafts/20210709-opa-authorizer-labels-observatorium-api.md
```diff
@@ -0,0 +1,72 @@
...
+* **Other Docs**
+  * Cross-Tenancy proposal???
+
+## TL;DR;
+
+We propose an API that allows OPA authorizers to return label matchers which get enforced read endpoints by the observatorium api.
+
+## Why
+
+OPA authorizers are the single source of through to check if a given subject (e.g. identified by a bearer token) has access to a requested resource (e.g. read/write on metrics and/or logs). Querying either the in-process via rego or any REST OPA authorizer relays this task from the observatorium-api to a separate isolated component. In addition an OPA authorizer can relay the request further to a third-party system (e.g. RedHat AMS, OpenShift/Kubernetes API server). Upon successful authorization the OPA authorizer can request further information (e.g. list of sub-resources the subject has access) and pass it back to the observatorium api. Such further information can be used to further refine the read endpoints (e.g. return logs for kubernetes namespaces the subject has access to).
```
squat, line 23: Here you are describing the "what" or "how". This section should be about "why we need something like this" independently of the implementation.

docs/proposals/drafts/20210709-opa-authorizer-labels-observatorium-api.md
```diff
@@ -0,0 +1,72 @@
...
+
+## How
+
+We propose to extend and further standardize the API between the observatorium-api and OPA authorizers. The current implementation includes only a boolean flag to signal success or failure for the authorization request against an OPA endpoint. This proposal introduces a second optional field to pass label matchers back to the API, e.g.:
+
+Current implementation payload:
+
+```json
+{
+    "result": "true|false"
```
squat, line 42: nit: it's a boolean, not a string representing a boolean
```suggestion
    "result": true|false
```

docs/proposals/drafts/20210709-opa-authorizer-labels-observatorium-api.md
```diff
@@ -0,0 +1,74 @@
...
+{
+    "result": true|false,
+    "matchers": [
+        {
+            "name":  "label key",
+            "type":  "MatchEqual|MatchNotEqual|MatchRegex|MatchNotReqex",
+            "value": "label value"
+        }
+    ]
+}
```
onprem, line 60: For cross-tenant use case, we would need to enforce different matchers depending on the tenant.

For example, Kafka people want to access CPU usage metric from OSD for their clusters, so we need to enforce `{customer="kafka"}` only for OSD metrics accessed by someone from Kafka. A query like 

`kafka_price_per_cpu * cpu_usage_total{tenant="osd"}` 

needs to be converted to 

`kafka_price_per_cpu{tenant="kafka"} * cpu_usage_total{tenant="osd", customer="kafka"}`.

I propose to change this payload to something like 
```suggestion
{
    "result": true|false,
    "matchers": {
        "tenant-a": [
            {
                "name":  "label key",
                "type":  "MatchEqual|MatchNotEqual|MatchRegex|MatchNotReqex",
                "value": "label value"
            }
        ],
        "tenant-b": []
    }
}
```

onprem: Changing this payload right now is not required but we have to do this for cross-tenant queries in the future.

onprem (approved):
This looks good to me! Just a comment about the cross-tenancy use case.

docs/proposals/drafts/20210709-opa-authorizer-labels-observatorium-api.md
periklis: Alright let's do this in a separate step. This doesn't hurt if we change this later.

PR #422, opened 2021-08-05
Propose `mdox` for website pre-process

saswatamcode:
This PR proposes the use of [`mdox`](https://github.com/bwplotka/mdox) for website pre-process using `mdox transform` as an alternative to the previous symlink setup. 

- This ensures that all frontmatter is generated by `mdox` and all `README.md` files in `docs/` are changed to `_index.md`.
- Also adds `make web-pre` to the Makefile and updates netlify job.

PR #430, opened 2021-11-01
Rules API proposal: Accommodate alerting rules

jessicalins:
This PR updates the Rules API proposal to also accommodate alerting rules. I just reword the proposal and removed the "What if a user gives us an alerting rule" section.

### DIFF
--- a/docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
+++ b/docs/proposals/20201019-prometheus-rules-in-observatorium-api.md
@@ -19,13 +19,9 @@
 - [Goals](#goals)
 - [Non-Goals](#non-goals)
 - [How](#how)
-  * [How does Rule get tenant recording rules](#how-does-rule-get-tenant-recording-rules)
+  * [How does Rule get tenant recording and alerting rules](#how-does-rule-get-tenant-recording-and-alerting-rules)
   * [How do we store tenant rules](#how-do-we-store-tenant-rules)
     + [Service](#service)
-      - [What if a user gives us an alerting rule](#what-if-a-user-gives-us-an-alerting-rule)
-        * [1. Reject Request](#1-reject-request)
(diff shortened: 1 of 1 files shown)


squat (approved):
I love it

jessicalins:
@squat I don't really get why tests are failing :( (apparently something in `build.command` from `netfly.toml`? (from the [logs](https://app.netlify.com/sites/observatorium/deploys/618023c18f7386000725c2a0))

squat:
@jessicalins the build is failing because the Makefile target `web` requires the dependency `$(WEBSITE_DIR)/node_modules` however this dependency was removed in this PR (https://github.com/observatorium/observatorium/pull/422/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L37) so `make` complains that it cannot satisfy the dependency.

saswatamcode:
I kept the `$(WEBSITE_DIR)/node_modules` pre-requisite so that `web` or `web-serve` cannot execute fully if `website/node_modules` isn't present. The `web-pre` target creates this directory by executing `npm i` inside of `website/themes/doks` and is a pre-requisite for `web` & `web-serve`. 

It seems like this was a pre-requisite ordering issue. But then it should have come up earlier? 
A reason why it didn't occur earlier might be due to Netlify caching files between builds, so the `website/node_modules` target would be satisfied by that so earlier PRs didn't have this issue. 🤔

According to [docs](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html), it can be solved by changing the `web` & `web-serve` targets to have order-only prerequisites.
I'll raise a PR for this. 🙂

saswatamcode:
@jessicalins Rebasing should now fix the build! 🙂

ianbillett:
Thank you @saswatamcode :raised_hands:

ianbillett (dismissed):

PR #428, opened 2021-10-22
Propose a new pluggable authenticators

tareqmamari:
Propose refactoring the current authenticators setup to provide
pluggable based approach to make it flexible to onboard custom
authentication, this proposal would help new adopters to onboard
their authenticators.

PR #431, opened 2021-11-02
Fix web targets in Makefile

saswatamcode:
This PR fixes broken `web` & `web-serve` targets in Makefile as mentioned in #430 & #428.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,11 +50,9 @@ web-pre: $(MDOX)
 	cd $(WEBSITE_DIR)/themes/doks/ && npm install && rm -rf content
 
 .PHONY: web
-web: web-pre
-web: $(WEBSITE_DIR)/node_modules $(HUGO)
+web: | web-pre $(WEBSITE_DIR)/node_modules $(HUGO)
 	cd $(WEBSITE_DIR) && $(HUGO) -b $(WEBSITE_BASE_URL)
 
 .PHONY: web-serve
-web-serve: web-pre
-web-serve: $(WEBSITE_DIR)/node_modules $(HUGO)
+web-serve: | web-pre $(WEBSITE_DIR)/node_modules $(HUGO)
 	@cd $(WEBSITE_DIR) && $(HUGO) serve


onprem (approved):
Thanks for the quick fix! :100:

Makefile
```diff
@@ -50,11 +50,9 @@ web-pre: $(MDOX)
...
 .PHONY: web
-web: web-pre
-web: $(WEBSITE_DIR)/node_modules $(HUGO)
+web: | web-pre $(WEBSITE_DIR)/node_modules $(HUGO)
 	cd $(WEBSITE_DIR) && $(HUGO) -b $(WEBSITE_BASE_URL)
 
 .PHONY: web-serve
-web-serve: web-pre
-web-serve: $(WEBSITE_DIR)/node_modules $(HUGO)
+web-serve: | web-pre $(WEBSITE_DIR)/node_modules $(HUGO)
```
squat, line 57: Hmm i don't think this is quite right.
Firstly, $(HUGO) should really be a normal dependency as it is a normal file that can be of the wrong version and whose version changes can require rebuilds of the website.

Second, as far as I can see there is still no target that can be called to generate `$(WEBSITE_DIR)/node_modules`. This means that if the directory doesn't exist on your development machine, then make won't know how to generate it and will still fail. If the directory is generated as part of `make web-pre` then we can eliminate the `$(WEBSITE_DIR)/node_modules` dependency as it is redundant and doesn't provide any recipe

saswatamcode: > Firstly, $(HUGO) should really be a normal dependency as it is a normal file that can be of the wrong version and whose version changes can require rebuilds of the website.

Yes, but if treated as a normal dependency, in case of a fresh install (e.g Netlify), hugo would be installed first, and you'd have to wait for it as it's the largest dependency, after which there might be errors in when running `web-pre`, sot it wouldn't fail fast. 

By keeping it at the last like here, hugo would only be installed once all required steps are done. But I can change this in the future if required. 🙂

> If the directory is generated as part of make web-pre then we can eliminate the $(WEBSITE_DIR)/node_modules dependency as it is redundant and doesn't provide any recipe

Yes, I agree. I can fix this in the next PR. 🙂

squat: > Yes, but if treated as a normal dependency

No, it's in the order-only dependencies list. It may be treated as a "normal" dependency by other targets, however, setting our dependency as the wrong type and then implicitly relying on the relationship between dependencies of other targets is fragile and bad practice. Imagine that all of the targets that depend on `$(HUGO)` had listed it as a order-only dependency. If the `hugo` binary were ever to be updated, then our website would not get updated and make use of new hugo generation features.

I would say we should definitely fix this to avoid surprises when another target changes its relationship type on hugo (which could happen for legitimate reasons).

saswatamcode: Will fix it then!

squat: :bow: thanks @saswatamcode

PR #437, opened 2021-11-05
Fix broken website links and web targets

saswatamcode:
This PR fixes broken links on the website i.e links that point to non-markdown or non-image resources, by updating `mdox` to the latest version and changing config. 

It also fixes Makefile web targets according to comments [here](https://github.com/observatorium/observatorium/pull/431#discussion_r740836672). cc @squat


squat (approved):
Great work @saswatamcode
Thanks for fixing this :tada:

PR #439, opened 2021-11-11
Fix Prometheus download instructions for MacOS

esnible:
Signed-off-by: Ed Snible <snible@us.ibm.com>


squat (approved):
Thanks @esnible. This is a good catch :tada:

Issue #442, opened 2021-11-19
Proposal add tracing API and OpenTelemetry collector


pavolloffay:
This is a proposal to expose tracing ingestion API in the Observatorium backed by the OpenTelemetry collector. 

The OTEL collector does not provide any persistence, it is just a data forwarder as of now, however, it can be configured to send data to a variety of observability vendors or OSS projects like Jaeger or Zipkin or simply log data to the standard output. After this proposal is done we will be working on adding a storage/platform for trace data (e.g. Jaeger). 

This proposal contains two main action items:
1. Expose tracing ingestion API in the API service
2. Add deployments manifests for OpenTelemetry collector

### Expose tracing ingestion API
There are a couple of open-source tracing protocols out there - Zipkin, Jaeger and OpenTelemetry. At the moment the OpenTelemetry seems to be the most popular and it is projected to have the biggest adoption. Hence I would propose to use the OpenTelemetry as the trace ingestion API.

The OpenTelemetry protocol primarily supports gRPC for sending traces (I will refer to it as OTLP gRPC) with proto encoding. The HTTP protocol with protobuf encoding is as well supported. The JSON encoding is still in experimental mode. The majority of users use OTLP gRPC and the HTTP is used in environments where gRPC cannot be used (e.g. mobile clients). Because the Observatorium supports HTTP I would propose to start with OTLP HTTP and at the same time discuss how OTLP gRPC could be supported. 

#### Multitenancy 
The multitenancy should be handled the same way as it is for metrics and logs. The API service could add a HTTP header or attribute/label to data. This label would then be used in the collector to identify the tenant. 

### Add deployment manifests for OpenTelemetry collector

This is a pretty straightforward task. There is OpenTelemetry operator or HELM chart, but we could as well use plain Kubernetes manifests given the simplicity of this stateless component. 

#### Other related changes
The Observatorium API is instrumented with OpenTelemetry API. During the deployment it could be automatically configured to report data the OTEL collector as an internal user.

* OTLP HTTP: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp
* OTEL operator: https://github.com/open-telemetry/opentelemetry-operator
* OTEL Helm chart: https://github.com/open-telemetry/opentelemetry-helm-charts
* OTLP proto: https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto/trace/v1

matej-g:
[...]

PR #443, opened 2021-11-23
Add proposal to add tracing ingestion API and OpenTelemtry collector

pavolloffay:
This proposal bootstraps ingestion of tracing data in Observatorium.
The second proposal will build on top of this and add traicng
platform/storage.

Updates: #442 

Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

### DIFF
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -43,7 +43,7 @@ At the moment Observatorium does not expose API to ingest distributed tracing da
 
 ## Goals
 
-* Expose API to ingest tracing data.
+* Expose API in the Observatorium API service to ingest tracing data.
 * Deploy a component that receives tracing data and is able to forward it to tracing platforms with storage.
 
 ## Non-Goals
@@ -75,7 +75,7 @@ The OpenTelemetry collector community already support deploying the collector vi
 
 * Iterate and finalise this design document.
 * Add deployment manifests for OpenTelemetry collector.
-* Add OTLP HTTP support to the Observatorium API.
+* Add OTLP HTTP with binary encoding to the Observatorium API.
 * Add OTLP gRPC support to the Observatorium API.
 
 ## References
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -6,8 +6,6 @@
 * **Related Tickets:**
     * [observatorium/issues/442](https://github.com/observatorium/observatorium/issues/442)
 
-* **Other docs:**
-
 ## Table of Contents
 
 - [TLDR](#tldr)
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -67,6 +67,10 @@ Multitenancy could be handled the same way as it is handled for metrics and logs
 
 ### Add deployments manifests for OpenTelemetry collector
 
+OpenTelemetry collector allows receiving, transforming and forwarding data to multiple systems. For instance, it can receive tracing data in Jaeger, Zipkin and OpenTelemetry formats. The transformation allows changing attributes, remove personally identifiable information, derive metrics from traces, perform tail-based sampling and more. After the processing the data can be sent to a single or multiple systems. These set of features allows users to solve variety of use-cases. The Observatorium users deploying on premise would be able to easily forward data to an observability vendor and the same time keep part of the data locally in e.g. Jaeger with low retention or without removing PII data.
+
+The OpenTelemetry collector deployment will be as well needed for sending data to Jaeger. The Jaeger collector does not support receiving OTLP protocol.
+
 The OpenTelemetry collector community already support deploying the collector via [OpenTelemetry operator](https://github.com/open-telemetry/opentelemetry-operator) or [HELM chart](https://github.com/open-telemetry/opentelemetry-helm-charts). A plain Kubernetes manifests could be as well used given the stateless nature and low complexity of the collector.
 
 ## Action Plan
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -69,7 +69,7 @@ Multitenancy could be handled the same way as it is handled for metrics and logs
 
 OpenTelemetry collector allows receiving, transforming and forwarding data to multiple systems. For instance, it can receive tracing data in Jaeger, Zipkin and OpenTelemetry formats. The transformation allows changing attributes, remove personally identifiable information, derive metrics from traces, perform tail-based sampling and more. After the processing the data can be sent to a single or multiple systems. These set of features allows users to solve variety of use-cases. The Observatorium users deploying on premise would be able to easily forward data to an observability vendor and the same time keep part of the data locally in e.g. Jaeger with low retention or without removing PII data.
 
-The OpenTelemetry collector deployment will be as well needed for sending data to Jaeger. The Jaeger collector does not support receiving OTLP protocol.
+The OpenTelemetry collector deployment is needed for sending data to Jaeger. The Jaeger collector does not support OTLP protocol.
 
 The OpenTelemetry collector community already support deploying the collector via [OpenTelemetry operator](https://github.com/open-telemetry/opentelemetry-operator) or [HELM chart](https://github.com/open-telemetry/opentelemetry-helm-charts). A plain Kubernetes manifests could be as well used given the stateless nature and low complexity of the collector.
 
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -1,4 +1,4 @@
-# Add traces ingestion API and OpenTelemetry collector
+# Traces ingestion API and OpenTelemetry collector
 
 * **Owners:**:
     * `@pavolloffay`
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -1,4 +1,4 @@
-# Rules API Proposal
+# Add traces ingestion API and OpenTelemetry collector
 
 * **Owners:**:
     * `@pavolloffay`
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -15,7 +15,6 @@
 - [How](#how)
   * [Expose tracing ingestion API in the Observatorium API service](#expose-tracing-ingestion-api-in-the-observatorium-api-service)
     + [Multitenancy](#multitenancy)
-  * [Add deployments manifests for OpenTelemetry collector](#add-deployments-manifests-for-opentelemetry-collector)
 - [Action Plan](#action-plan)
 - [References](#references)
 
@@ -33,7 +32,7 @@ Table of contents generated with markdown-toc
 
 ## TLDR
 
-This document proposes to expose tracing ingestion API  backed by the OpenTelemetry collector. This is a first step to support ingestion and storage for tracing data. The OpenTelemetry collector gives flexibility to forward tracing data to variety of tracing platforms, both on-prem or SaaS - Jaeger, Zipkin, Tempo, SigNoz or DataDog, Dynatrace etc.
+This document proposes to expose tracing ingestion API. This is a first step to support ingestion and storage for tracing data.
 
 ## Why
 
@@ -42,7 +41,6 @@ At the moment Observatorium does not expose API to ingest distributed tracing da
 ## Goals
 
 * Expose API in the Observatorium API service to ingest tracing data.
-* Deploy a component that receives tracing data and is able to forward it to tracing platforms with storage.
 
 ## Non-Goals
 
@@ -51,33 +49,21 @@ At the moment Observatorium does not expose API to ingest distributed tracing da
 
 ## How
 
-The how is split into two sections:
-1. Expose tracing ingestion API
-2. Add deployment manifests for OpenTelemetry collector
-
 ### Expose tracing ingestion API in the Observatorium API service
 
 There are a couple of open-source tracing protocols out there - Zipkin, Jaeger and OpenTelemetry. At the moment the OpenTelemetry seems to be the [most popular](https://opentelemetry.io/vendors/), and it is projected to have the biggest adoption in the future. Hence, using OpenTelemetry protocol seems to be the most appropriate choice.
 
-The OpenTelemetry protocol primarily supports gRPC for sending traces (OTLP gRPC) with proto encoding. The HTTP protocol with protobuf encoding is supported as well, but the JSON encoding is still in experimental mode. The majority of users use OTLP gRPC and the HTTP is used in environments where gRPC cannot be used (e.g. mobile clients). Because the Observatorium already supports HTTP and the final state is to support both OTLP HTTP and gRPC, the Observatorium could initially support HTTP and in parallel start working on gRPC.
+The OpenTelemetry protocol primarily supports gRPC for sending traces (OTLP gRPC) with proto encoding. The HTTP protocol with protobuf encoding is supported as well, but the JSON encoding is still in experimental mode. The majority of users use OTLP gRPC and the HTTP is used in environments where gRPC cannot be used (e.g. mobile clients).
 
-#### Multitenancy
+Observatorium should expose OTLP gRPC protocol and later add support for OTLP HTTP if there will be use-case to require it.
 
-Multitenancy could be handled the same way as it is handled for metrics and logs. The API service could add an HTTP header or attribute/label to data. This label would then be used in the collector to identify the tenant. Then the collector could route the data to a specific exporter per tenant or export the data to platform/storage by using a single exporter.
+### Multitenancy
 
-### Add deployments manifests for OpenTelemetry collector
-
-OpenTelemetry collector allows receiving, transforming and forwarding data to multiple systems. For instance, it can receive tracing data in Jaeger, Zipkin and OpenTelemetry formats. The transformation allows changing attributes, remove personally identifiable information, derive metrics from traces, perform tail-based sampling and more. After the processing the data can be sent to a single or multiple systems. These set of features allows users to solve variety of use-cases. The Observatorium users deploying on premise would be able to easily forward data to an observability vendor and the same time keep part of the data locally in e.g. Jaeger with low retention or without removing PII data.
-
-The OpenTelemetry collector deployment is needed for sending data to Jaeger. The Jaeger collector does not support OTLP protocol.
-
-The OpenTelemetry collector community already support deploying the collector via [OpenTelemetry operator](https://github.com/open-telemetry/opentelemetry-operator) or [HELM chart](https://github.com/open-telemetry/opentelemetry-helm-charts). A plain Kubernetes manifests could be as well used given the stateless nature and low complexity of the collector.
+Multitenancy will be handled the same way as it is handled for metrics and logs. The API service will add an HTTP header or attribute/label to data. This label would then be used in the collector to identify the tenant and store data accordingly.
 
 ## Action Plan
 
 * Iterate and finalise this design document.
-* Add deployment manifests for OpenTelemetry collector.
-* Add OTLP HTTP with binary encoding to the Observatorium API.
 * Add OTLP gRPC support to the Observatorium API.
 
 ## References
--- a/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
+++ b/docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
@@ -59,7 +59,7 @@ The how is split into two sections:
 
 ### Expose tracing ingestion API in the Observatorium API service
 
-There are a couple of open-source tracing protocols out there - Zipkin, Jaeger and OpenTelemetry. At the moment the OpenTelemetry seems to be the most popular, and it is projected to have the biggest adoption in the future. Hence, using OpenTelemetry protocol seems to be the most appropriate choice.
+There are a couple of open-source tracing protocols out there - Zipkin, Jaeger and OpenTelemetry. At the moment the OpenTelemetry seems to be the [most popular](https://opentelemetry.io/vendors/), and it is projected to have the biggest adoption in the future. Hence, using OpenTelemetry protocol seems to be the most appropriate choice.
 
 The OpenTelemetry protocol primarily supports gRPC for sending traces (OTLP gRPC) with proto encoding. The HTTP protocol with protobuf encoding is supported as well, but the JSON encoding is still in experimental mode. The majority of users use OTLP gRPC and the HTTP is used in environments where gRPC cannot be used (e.g. mobile clients). Because the Observatorium already supports HTTP and the final state is to support both OTLP HTTP and gRPC, the Observatorium could initially support HTTP and in parallel start working on gRPC.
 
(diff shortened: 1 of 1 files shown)


matej-g (commented):
Thanks for opening this @pavolloffay. I think the proposal is sufficiently scoped and the objectives are clear, I have only one unanswered question.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,86 @@
...
+This document proposes to expose tracing ingestion API  backed by the OpenTelemetry collector. This is a first step to support ingestion and storage for tracing data. The OpenTelemetry collector gives flexibility to forward tracing data to variety of tracing platforms, both on-prem or SaaS - Jaeger, Zipkin, Tempo, SigNoz or DataDog, Dynatrace etc.
+
+## Why
+
+At the moment Observatorium does not expose API to ingest distributed tracing data.
+
+## Goals
+
+* Expose API to ingest tracing data.
+* Deploy a component that receives tracing data and is able to forward it to tracing platforms with storage.
```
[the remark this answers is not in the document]
pavolloffay: >  For example, should tenants be able to configure what kind of exporters or processor should be used in the Collector? 

I don't think so. Processors are used for data manipulation/mutation and this can be done before in the customer's cluster before reporting it to the platform/observatorium.

> Is that goal for this proposal or perhaps will be goal for a future proposal?

I don't see this use case at the moment.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,86 @@
...
+
+This document proposes to expose tracing ingestion API  backed by the OpenTelemetry collector. This is a first step to support ingestion and storage for tracing data. The OpenTelemetry collector gives flexibility to forward tracing data to variety of tracing platforms, both on-prem or SaaS - Jaeger, Zipkin, Tempo, SigNoz or DataDog, Dynatrace etc.
+
+## Why
+
+At the moment Observatorium does not expose API to ingest distributed tracing data.
+
+## Goals
+
+* Expose API to ingest tracing data.
```
jessicalins, line 46: I imagine here we would expand the existent observatorium API (https://github.com/observatorium/api)? if yes, maybe we could rewrite to something like "Expand Observatorium API to support tracing data ingestion"

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,86 @@
...
+
+## How
+
+The how is split into two sections:
+1. Expose tracing ingestion API
+2. Add deployment manifests for OpenTelemetry collector
+
+### Expose tracing ingestion API in the Observatorium API service
+
+There are a couple of open-source tracing protocols out there - Zipkin, Jaeger and OpenTelemetry. At the moment the OpenTelemetry seems to be the most popular, and it is projected to have the biggest adoption in the future. Hence, using OpenTelemetry protocol seems to be the most appropriate choice.
```
jessicalins, line 62: are there any evaluation docs/relevant research that can be added here as a reference that backs up the choice of OpenTelemetry?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Add deployments manifests for OpenTelemetry collector
+
+The OpenTelemetry collector community already support deploying the collector via [OpenTelemetry operator](https://github.com/open-telemetry/opentelemetry-operator) or [HELM chart](https://github.com/open-telemetry/opentelemetry-helm-charts). A plain Kubernetes manifests could be as well used given the stateless nature and low complexity of the collector.
+
+## Action Plan
+
+* Iterate and finalise this design document.
+* Add deployment manifests for OpenTelemetry collector.
+* Add OTLP HTTP support to the Observatorium API.
```
jessicalins, line 78: will the OTLP HTTP support be added with binary (stable) or JSON (experimental) format? from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp

jessicalins (commented):
@pavolloffay thanks for opening the proposal :) 
Overall looks good to me, but as @matej-g suggested I think would be nice if everyone can take a look too. I have a few questions/suggestions

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: I want to add only stable APIs so the binary. JSON could be added later if it is stable and there is a use-case for it.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: yes this is about extending the API service. I will make it clear in the text.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: I will try to find something

pavolloffay: I have added some links to the references section and one link in this paragraph.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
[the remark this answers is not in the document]
matej-g: Ok, so if I understand correctly, for example tenant won't be able to configure how or where the data is exported, but this will be governed by the operator, correct? I'm trying to clarify this with regards to what's stated in multitenancy, i.e. that collector could route to different exporters based on tenants. But my assumption is that after ingestion, only operator will decide how data is exported and the user will have no say in it (maybe could be worth specifying in non-goals).

matej-g (dismissed):

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,89 @@
+# Traces ingestion API and OpenTelemetry collector
+
+* **Owners:**:
+    * `@pavolloffay`
+
+* **Related Tickets:**
+    * [observatorium/issues/442](https://github.com/observatorium/observatorium/issues/442)
+
+* **Other docs:**
```
jessicalins, line 9: I would just remove this line since there are no further links here

jessicalins (approved):
just a tiny suggestion, but overall looks good to me:)

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,87 @@
...
+### Add deployments manifests for OpenTelemetry collector
+
+The OpenTelemetry collector community already support deploying the collector via [OpenTelemetry operator](https://github.com/open-telemetry/opentelemetry-operator) or [HELM chart](https://github.com/open-telemetry/opentelemetry-helm-charts). A plain Kubernetes manifests could be as well used given the stateless nature and low complexity of the collector.
+
+## Action Plan
+
+* Iterate and finalise this design document.
+* Add deployment manifests for OpenTelemetry collector.
+* Add OTLP HTTP with binary encoding to the Observatorium API.
+* Add OTLP gRPC support to the Observatorium API.
```
esnible, line 77: Thanos has a gRPC store API, and I believe the Loki Push API also works with gRPC.
- Is the intention to have a single gRPC endpoint for all three types, or one endpoint for OTLP, one for Loki Push, and one for Prometheus Remote Write
- Is it too early to request that the Observatorium gRPC endpoint support introspection?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: done

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: > Thanos has a gRPC store API, and I believe the Loki Push API also works with gRPC. Yet those endpoints are not exposed on Observatorium. Observatorium should support either none or eventually all of the protocols via gRPC.

Good point, perhaps these gRPC endpoints are just implementation/internal details of mentioned systems. 
If we decide to expose multiple gRPC services then there are various ways how it can be done (single vs multiple ports). 

> Is it too early to request that the Observatorium gRPC endpoint support introspection?

This feature should be discussed separately and not in this proposal.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,87 @@
...
+
+The OpenTelemetry protocol primarily supports gRPC for sending traces (OTLP gRPC) with proto encoding. The HTTP protocol with protobuf encoding is supported as well, but the JSON encoding is still in experimental mode. The majority of users use OTLP gRPC and the HTTP is used in environments where gRPC cannot be used (e.g. mobile clients). Because the Observatorium already supports HTTP and the final state is to support both OTLP HTTP and gRPC, the Observatorium could initially support HTTP and in parallel start working on gRPC.
+
+#### Multitenancy
+
+Multitenancy could be handled the same way as it is handled for metrics and logs. The API service could add an HTTP header or attribute/label to data. This label would then be used in the collector to identify the tenant. Then the collector could route the data to a specific exporter per tenant or export the data to platform/storage by using a single exporter.
+
+### Add deployments manifests for OpenTelemetry collector
```
bwplotka, line 68: I am missing the purpose of this. Why do we need OpenTelemetry Collector anywhere in Observatorium? 🤔 API is in front of the backend, no need for this extra component.

Is it because we want show example way of collecting traces for it?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
bwplotka: > Thanos has a gRPC store API, and I believe the Loki Push API also works with gRPC. Yet those endpoints are not exposed on Observatorium. Observatorium should support either none or eventually all of the protocols via gRPC.

Where this assumption comes from? Thanos storeAPI is an internal API which is never exposed. API is for external users, and it just happen that HTTP APIs are the most popular (Prometheus Query APIs, remote write and so on). 

I see no problem in adding gRPC OTLP if we need, even if other APIs are HTTP. Let's do it if we see grpc otlp more popular. Just a bit more work to adjust our middleware for interceptors if we want them 👍🏽

bwplotka: > Is it too early to request that the Observatorium gRPC endpoint support introspection?

No problem with https://github.com/grpc/grpc/blob/master/doc/server-reflection.md if that is what we want.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,87 @@
...
+
+The OpenTelemetry protocol primarily supports gRPC for sending traces (OTLP gRPC) with proto encoding. The HTTP protocol with protobuf encoding is supported as well, but the JSON encoding is still in experimental mode. The majority of users use OTLP gRPC and the HTTP is used in environments where gRPC cannot be used (e.g. mobile clients). Because the Observatorium already supports HTTP and the final state is to support both OTLP HTTP and gRPC, the Observatorium could initially support HTTP and in parallel start working on gRPC.
+
+#### Multitenancy
+
+Multitenancy could be handled the same way as it is handled for metrics and logs. The API service could add an HTTP header or attribute/label to data. This label would then be used in the collector to identify the tenant. Then the collector could route the data to a specific exporter per tenant or export the data to platform/storage by using a single exporter.
```
bwplotka, line 66: This also mentions collector. Why do we need any collector in Observatorium API? Why it cannot just be L7 proxy that terminates TLS and passes the grpc/http messages further?

bwplotka: The question is then how the tenancy is planned to work. I believe this depends on the storage we use. Can tenancy works by just adding a new "Tag" to traces we can then isolate data from?

Observatorium API purpose is to allow this - is there a way to then query for traces from tenant X but not see traces from tenant Y while using our APIs? This is exactly something I think we missed in this proposal - what's the plan. Even if we want to skip tenancy for now - that should be a clear non goal then.

bwplotka (changes requested):
Thanks for this and sorry for lag in review!

It looks good generally just few missing pieces:

* IMO gRPC otlp is the perfect choice (no need to play with HTTP protocol if it's not finished and not used)
* Why do we need collector anywhere in Observatorium stack? I don't see a need for such heavy dependency anywhere for backend things.
* I think multi-tenancy has some missing details

Thank, and I promise faster review iterations next time🙈

pavolloffay:
> Why do we need collector anywhere in Observatorium stack? I don't see a need for such heavy dependency anywhere for backend things.

Could you please explain "such a heavy dependency". OTEL collector is quite a small component that just receives, transforms and forwards data. I will add some reasoning to the text why we should add the collector.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: > Can tenancy works by just adding a new "Tag" to traces we can then isolate data from?

This will not work with Jaeger at the moment.

> Observatorium API purpose is to allow this - is there a way to then query for traces from tenant X but not see traces from tenant Y while using our APIs? This is exactly something I think we missed in this proposal - what's the plan. Even if we want to skip tenancy for now - that should be a clear non goal then.

The proposal is simply about adding ingestion API and not query part.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,91 @@
...
+Table of contents generated with markdown-toc
+</a>
+
+</i>
+
+</small>
+
+## TLDR
+
+This document proposes to expose tracing ingestion API  backed by the OpenTelemetry collector. This is a first step to support ingestion and storage for tracing data. The OpenTelemetry collector gives flexibility to forward tracing data to variety of tracing platforms, both on-prem or SaaS - Jaeger, Zipkin, Tempo, SigNoz or DataDog, Dynatrace etc.
```
bwplotka, line 36: Hmm, while it sounds great, I don't see why the backend needs such flexibility. We own this path, we can add support of anything to the solution we would use. 

Besides that, sounds to me like our main potential solutions are Jaeger and Tempo and both support just the OTLP tracing protocol, right? 🤔 

PS: And why do we need DataDog? In such case, there is no need for Observatorium, no?

If anyone wants flexibility they can instal Otel collector before Observatorium. IMO we should make as simple an approach as possible. Complexity with generic flexibility has a major cost.

> Could you please explain "such a heavy dependency". OTEL collector is quite a small component that just receives, transforms and forwards data. I will add some reasoning to the text why we should add the collector.

Collector main distro (I assume we want to use that) itself has ~35 dependencies. Repo has 200 issue despite impressive amount of activity and maintainers - that is NOT a lightweight dependency - and easy to maintain. And if we want to put this as a sidecar that is actually a massive complexity addition (another process to maintain, monitor, understand, debug if it's performance hit, etc). Let's put it in different perspective. Adding Otel collector as sidecar to Observatorium API proxy is like adding Prometheus next to Observatorium API. In terms of complexity and size of those projects, our API proxy is tiny. I don't think Otel collector is a free lunch we can treat as black box here.

In order to achieve what we want (tenancy, auth, rate limit) I believe we could get this working correctly for one protocol first. I am afraid making the number of permutations exponential of different outputs (or inputs) will only drop our focus.

At the end, it's really ~100 lines of code to add tag / check stuff per HTTP/gRPC message. Instead of complex collector code, I think that's a win.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
bwplotka: > This will not work with Jaeger at the moment.

So what will work? I don't see other solution proposed.

> The proposal is simply about adding ingestion API and not query part.

Ok, but ingesting dat without tenancy is blocking future use this data in multi-tenant world. I think we should at least discuss options here.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### Multitenancy
+
+Multitenancy could be handled the same way as it is handled for metrics and logs. The API service could add an HTTP header or attribute/label to data. This label would then be used in the collector to identify the tenant. Then the collector could route the data to a specific exporter per tenant or export the data to platform/storage by using a single exporter.
+
+### Add deployments manifests for OpenTelemetry collector
+
+OpenTelemetry collector allows receiving, transforming and forwarding data to multiple systems. For instance, it can receive tracing data in Jaeger, Zipkin and OpenTelemetry formats. The transformation allows changing attributes, remove personally identifiable information, derive metrics from traces, perform tail-based sampling and more. After the processing the data can be sent to a single or multiple systems. These set of features allows users to solve variety of use-cases. The Observatorium users deploying on premise would be able to easily forward data to an observability vendor and the same time keep part of the data locally in e.g. Jaeger with low retention or without removing PII data.
```
bwplotka, line 70: > Observatorium users deploying on premise would be able to easily forward data to an observability vendor and the same time keep part of the data locally in e.g. Jaeger with low retention or without removing PII data.
Write


They can do that BEFORE data reach Observatorium, no? I don't see value in coupling Observatorium with collection transformations Collector offers.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,91 @@
...
+
+The OpenTelemetry collector deployment will be as well needed for sending data to Jaeger. The Jaeger collector does not support receiving OTLP protocol.
```
bwplotka, line 72: I assume you mean a sidecar deployment here?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,91 @@
...
+
+The OpenTelemetry collector community already support deploying the collector via [OpenTelemetry operator](https://github.com/open-telemetry/opentelemetry-operator) or [HELM chart](https://github.com/open-telemetry/opentelemetry-helm-charts). A plain Kubernetes manifests could be as well used given the stateless nature and low complexity of the collector.
```
bwplotka, line 74: I think this is quite unnecessary complexity here. 🤔

bwplotka (changes requested):
Thanks for updates. Hope you don't mind harsh feedback. I love your work and ideas here, but I just want to make sure we don't unnecessarly couple different complex system into one. The power of Observatorium is as "backend", not collector or proxy, so ideally we could treat it like this, minimizing complexity, simplifying the operational model and ensuring scalability. 

If you still think Otel Collector is the way to go, can we add the "Alternative" section which would explain the alternative option (with no collector mode?). I believe it's quite straightforward - I disagree we need multiple outputs for "backend" side.

Also, do you mean to run Otel collector as a sidecar?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
bwplotka: I see you started great discussion on Jaeger itself, nice, great job! Looking though this productive  discussion now https://github.com/jaegertracing/jaeger/issues/3427

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
matej-g: I think if we could take a step back and summarize what we would want to achieve and what are the options. I think we can agree on wanting the API to accept spans in OTLP protocol (whether gRPC or HTTP) as our initial position.

The question is, after ingesting the traces, what are our options to export the spans. So far as I understand the two sides:
1. We can send the traces to Otel Collector, which will forward them to appropriate tracing backend
 - the downside: we need to introduce an extra component (collector) between API and tracing backend = extra dependency, extra overhead etc.
 
2. We send traces to a tracing backend which is directly capable of accepting OTLP and remove the collector
- the downside: this leaves out Jaeger, which does not seem to support OTLP currently; What is the alternative here? @bwplotka you suggested Tempo? How big of a dependency is that in comparison to collector+Jaeger? What are it's other (dis)advantages?

Am I understanding this correctly? I believe we should capture these points in the chosen solution and in the alternatives.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
matej-g: Can you clarify what you mean by sidecar here? I was imagining it being more of a receiving component (similar to e.g. Thanos receive), i.e. API ingests the tracing data -> it is receive by collector -> collector exports it to a tracing backend.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
jessicalins: agree with @matej-g , I think would help systematically listing pros/cons for each option

> At the end, it's really ~100 lines of code to add tag / check stuff per HTTP/gRPC message. Instead of complex collector code, I think that's a win.

@bwplotka could you expand more on this? you mean we could write the necessary/custom code to support tracing for the OTLP protocol instead of adding the collector as a dependency?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
esnible: I took a look at gRPC OTLP.  gRPC doesn't have paths so I imagine the tenant being a header.  This is not a problem for OTLP.
The OTLP collector has an extension that can do AuthN, but not AuthZ.  It is straightforward to have Observatorium listen for gRPC and handle `opentelemetry.proto.collector.trace.v1.TraceService` and the same code as Observatorium to do AuthN and RBAC, then forward the traces along.  Any objections to this approach?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Why
+
+At the moment Observatorium does not expose API to ingest distributed tracing data.
+
+## Goals
+
+* Expose API in the Observatorium API service to ingest tracing data.
+* Deploy a component that receives tracing data and is able to forward it to tracing platforms with storage.
+
```
moadz, line 46: > forward it to tracing platforms with storage

This is quite open ended. It might be worth enumerating by use-case which tracing platforms we are trying to integrate with and deliver the simplest/maximum utility solution for that.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
moadz: Why doesn't OT support multi-tenancy? The ot-collector is intended as a vendor-agnostic implementation on how to receive, process and export telemetry data (including metrics/logs/tracing). The primary disjunct is that Observatorium is positioned as an _opinionated_, _multi-tenanted_ solution to the same problem. OT doesn't currently support soft-multi tenancy. So it seems contradictory to incur the complexity/overhead of maintaining stack agnostic components when the intention is to be opinionated. 

I think it's worth enumerating the core use-case here for the initial tracing implementation. Is it critical to have Jaeger support? Is the OT-collector the only interface we might have with Jaeger?

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: the answer is yes, no, and it depends (as always). It all comes down to what you want to achieve. 

I will repeat myself here. The biggest motivation for using OTELcol is to transform data from OTLP to Jaeger. The second use-case I have in mind is when Observatorium is deployed on-prem, hence running transformations at a single place should be better than maintaining several configurations in multiple places and making sure data is not leaking.

docs/proposals/20201123-tracing-ingestion-api-and-opentelemetry-collector.md
pavolloffay: Is there any reason and benefit to deploy it as a sidecar? 

We could deploy it as Jaeger's sidecar - basically making it Jaeger's internal detail.

pavolloffay:
PR updated based on the discussion we had during the community meeting.

matej-g (approved):
Looks good, now minus the failing tests, but rerun should fix it

pavolloffay:
> Looks good, now minus the failing tests, but rerun should fix it


CI failed on:

> error: unable to read URL "https://raw.githubusercontent.com/coreos/kube-prometheus/master/manifests/setup/prometheus-operator-0servicemonitorCustomResourceDefinition.yaml", server reported 404 Not Found, status code=404

I doubt it is related to this PR. I cannot restart CI.

bwplotka (approved):
Simple and straight forward 👍🏽 Thanks

bwplotka:
You need to run `make docs` to fix lint, but I can do that later, let's merge (:

PR #444, opened 2021-11-25
Add Jaeger traces store proposal

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

Depends on #443


docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,124 @@
...
+
+## How
+
+The how is split into the following sections:
+* Jaeger configuration - describes how Jaeger components are deployed.
+* Jaeger Query - describes how users access data stored in Jaeger.
+
+### Jaeger configuration
+
+Currently, the most adopted and supported storage in Jaeger is Elasticsearch. Red Hat also supports Jaeger with Elasticsearch (6.x) as part of the OpenShift product. Therefore, this proposal assumes Elasticsearch will be used as a storage backend. Distributed tracing team at Red Hat is looking at alternative storages hence the storage technology can change in the future.
```
matej-g, line 68: Since this should be primarily community-driven effort, maybe mentioning Red Hat is not relevant. Could we rephrase this in a way that would say that Elasticsearch is the most used storage solution community-wide? (If we have any 'evidence' to support that statement that would be a cherry on top).

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,124 @@
...
+
+There are multiple Jaeger query APIs, in this case we propose to expose [api_v3](https://github.com/jaegertracing/jaeger-idl/tree/master/proto/api_v3). This API returns OpenTelemetry compatible payload.
+
+#### Expose Jaeger UI
+
+Jaeger Query with UI would be deployed in the Observatorium cluster and exposed outside the cluster. Each tenant would get a unique URL to access Jaeger Query UI.
+
+## Action Plan
+
+Given the issues with the single Jaeger instance topology, the best approach is to deploy a Jaeger instance per tenant that separates tenant's data by using different indices. This approach will require adding a configuration if a new tenant is added to the system.
```
matej-g, line 111: Could we in the section above make it a bit more clear which solution is the chosen one and which one is the alternative that will not be implemented? Right now it seems like both options will be considered.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,124 @@
...
+1. Expose Jaeger Query API in the Observatorium API.
+2. Expose Jaeger UI outside the cluster.
+
+#### Expose Jaeger Query API
+
+Jaeger query API would be exposed in the Observatorium API. The users would have to deploy Jaeger UI externally and use Observatorium API as a data source.
+
+There are multiple Jaeger query APIs, in this case we propose to expose [api_v3](https://github.com/jaegertracing/jaeger-idl/tree/master/proto/api_v3). This API returns OpenTelemetry compatible payload.
+
+#### Expose Jaeger UI
```
matej-g, line 105: Which one is the solution that is being proposed? Expose query API or to expose UI? Are there any additional concerns such as tenant authentication?

docs/proposals/20201125-jaeger-traces-store.md
matej-g: I understand the issues with single Jaeger instance variation, but I'm wondering:
1) How will we be able to ensure seamless addition of new tenants? It looks like this would require a concerted effort of first adding a new Jaeger instance and only then adding tenant to the API. I'm wondering if we can improve on this in any way to make the experience more coherent.
2) Will this solution scale without issues for e.g. hundreds of tenants, meaning hundreds of Jaeger instances? Are there any performance or operation overhead concerns?

matej-g (commented):
Thanks @pavolloffay for opening this, looks like a good draft! I added few suggestions.

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: For this we would like to get feedback from the Observatorium team. 

For us, it would be better to expose just the UI.

esnible: The Query API **must** be exposed outside the cluster for alternative UIs.

- If the UI is outside the cluster it is somewhat simple to marry it to the token-refresher. 
- If the UI is inside the cluster changes to the UI must be made to include the tenant (a part of the Authorization header the UI will receive) in the query the UI makes to the Query API.  I have concerns that this change would be difficult to audit.  For that reason I believe the Observatorium API should only expose query API, not UI.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,124 @@
...
+This deployment strategy will not work with the current Jaeger components. Here are few issues. The list might not be complete:
+* Service and operation name API. The service query API nor storage does not expose/store labels.
+* Dependency/service architecture diagram. To support this feature dependency schema, query, collector and Spark aggregation would have to change.
+* Data retention/TTL configurable pre tenant. The data retention in Elasticsearch is configurable per index.
+
+##### OpenTelemetry collector configuration
+
+The OpenTelemetry collector is configured with a single exporter to export data to a single Jaeger collector.
+
+#### Jaeger instance per tenant
```
esnible, line 85: Do you have a preference, one Jaeger instance per tenant or one Jaeger instance per Observatorium?  Having a single Jaeger per Observatorium would continue the pattern of logs and metrics.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,124 @@
...
+
+There are two ways users could access data stored in Jaeger:
+1. Expose Jaeger Query API in the Observatorium API.
+2. Expose Jaeger UI outside the cluster.
+
+#### Expose Jaeger Query API
+
+Jaeger query API would be exposed in the Observatorium API. The users would have to deploy Jaeger UI externally and use Observatorium API as a data source.
+
+There are multiple Jaeger query APIs, in this case we propose to expose [api_v3](https://github.com/jaegertracing/jaeger-idl/tree/master/proto/api_v3). This API returns OpenTelemetry compatible payload.
```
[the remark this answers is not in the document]
matej-g: It definitely looks somewhat repetitive, but the problem I see we are encapsulating two different APIs in one endpoint. For example for a metrics endpoint:

`api/metrics/v1/{tenant}/api/v1/receive`

The first API / version relates to Observatorium API; the other API / version relates to the underlying query API (whether Prometheus for metrics or Loki for logs). If we want to keep this flexibility (what if in the future there will be Prometheus API `v2` and we want to keep it compatible for both Prometheus API versions?), I think this explicitness is required.

docs/proposals/20201125-jaeger-traces-store.md
matej-g: From my perspective, I think it depends on which use cases are envisioned. If there are alternative UIs, as @esnible mentions, and we want to enable users to use them, then exposing API makes sense. Otherwise, if the Jaeger UI is supposed to be the only way to interact with the storage backend, do we gain anything by having read path in the API?

Just to give full picture, we do both for metrics: Users can interact with the read API, but they can also access the Thanos UI in a browser (provided they can authenticate themselves).

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: > this in a way that would say that Elasticsearch is the most used storage solution community-wide? 

The text already mentions this. I have added a Red Hat there bc the majority of deployed Observatorium instances are seems to be at Red Hat. Red Hat is also mentioned at various parts in the existing docs.

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: >  It looks like this would require a concerted effort of first adding a new Jaeger instance and only then adding tenant to the API.

As far as I know the API does not need any changes if a new tenant is added and this should be case with Jaeger as well. However, to add a new tenant Jaeger deployment will have to change (e.g. add a new collector).

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: I have a preference of a single Jaeger per observatorium, but that is not possible at the moment as the document says. To make that happen storage implementation would have to be made tenant aware.

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: I have updated this section. We need to expose query API that the Jaeger UI uses. If we expose the UI directly it depends. I am fine following the current design on exposing only APIs.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+* Configure OpenTelemetry collector to export data to Jaeger collector.
+* Deploy Jaeger with a persistent storage.
+* Define query API for trace data.
+
+## Non-Goals
+
+* Define trace ingestion API - it is done in [Traces ingestion API and OpenTelemetry collector](https://github.com/observatorium/observatorium/pull/443) proposal.
+
+## How
+
```
bwplotka, line 59: Suggestion for more structure 🤗
 
```suggestion

We plan to allow user to install Observatorium for tracing support (on top of metrics, logging or alone). This mode will deploy Jaeger and X using the following configuration...

Such mode will have the following dependency: Disk/etcd/elastic/... etc  Better integration with simpler and cheaper object storage as metrics, logging have will potentially come in another iterations.
```

docs/proposals/20201125-jaeger-traces-store.md
bwplotka: Red Hat also don't support Elasticsearch at this point, so not sure what's the point of those statements. 🤔

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+The how is split into the following sections:
+* Jaeger configuration - describes how Jaeger components are deployed.
+* Jaeger Query - describes how users access data stored in Jaeger.
+
+### Jaeger configuration
+
+Currently, the most adopted and supported storage in Jaeger is Elasticsearch. Red Hat also supports Jaeger with Elasticsearch (6.x) as part of the OpenShift product. Therefore, this proposal assumes Elasticsearch will be used as a storage backend. Distributed tracing team at Red Hat is looking at alternative storages hence the storage technology can change in the future.
+
+Follows possible multitenant Jaeger deployment topologies. All deployment topologies assume that Jaeger will be deployed by the [Jaeger Operator](https://github.com/jaegertracing/jaeger-operator) which depends on [OpenShift Elasticserach operator](https://github.com/openshift/elasticsearch-operator) and [Strimzi operator](https://operatorhub.io/operator/strimzi-kafka-operator) if Kafka is used. Jaeger can be as well configured to use externally managed Elasticsearch or Kafka instances.
```
bwplotka, line 68: I am bit confused here. So we plan to use Elastic as storage or Kafka AND elastic  or just Kafka or just Elastic?

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Single Jaeger instance
+
+This deployment topology deploys a single Jaeger instance for all tenants with a single Elasticsearch instance. The data for each tenant would contain a unique attribute to identify a tenant. The label would be dynamically injected in the Observatorium API service or in the OpenTelemetry collector.
```
bwplotka, line 72: Tiny vote on this option =D

```suggestion
This deployment topology deploys a single Jaeger instance for all tenants with a single Elasticsearch instance. The data for each tenant would contain a unique attribute to identify a tenant. The label would be dynamically injected in the Observatorium API service.
```

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+
+This deployment strategy will not work with the current Jaeger components. Here are few issues, and the list might not be complete:
+* Service and operation name API. The service query API nor storage does not expose/store labels.
+* Dependency/service architecture diagram. To support this feature dependency schema, query, collector and Spark aggregation would have to change.
+* Data retention/TTL configurable pre tenant. The data retention in Elasticsearch is configurable per index.
```
bwplotka, line 77: Worth to link to https://github.com/jaegertracing/jaeger/issues/3427?

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+
+##### OpenTelemetry collector configuration
+
+The OpenTelemetry collector is configured with a single exporter to export data to a single Jaeger collector.
+
+#### Jaeger instance per tenant
+
+This deployment topology deploys a Jaeger instance per tenant and all Jaeger instances use a single Elasticsearch instance. The data for each tenant is stored in a separate Elasticsearch index. Jaeger was designed to separate tenant data per index. Therefore, all Jaeger functionality works well with this deployment strategy.
+
+##### OpenTelemetry collector configuration
```
bwplotka, line 87: That's not really a soft-tenancy, but could work as a  workaround for now. I think such simple implementation could be embedded directly to Obs API too

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+
+The OpenTelemetry collector is configured with an exporter per tenant that exports data to Jaeger collector allocated per a single tenant.
+
+### Jaeger Query
+
+There are two ways users could access data stored in Jaeger:
+1. Expose only Jaeger Query API in the Observatorium API.
+2. Expose Jaeger UI outside the cluster.
```
bwplotka, line 95: Those sounds like they supplement? In order to put UI outside of cluster you need APIs in Obs API, no?

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+
+* Configure OpenTelemetry collector to export data to Jaeger collector.
+* Deploy Jaeger with a persistent storage.
+* Define query API for trace data.
+
+## Non-Goals
+
+* Define trace ingestion API - it is done in [Traces ingestion API and OpenTelemetry collector](https://github.com/observatorium/observatorium/pull/443) proposal.
+
+## How
```
bwplotka, line 58: As @matej-g mentioned, can we form this proposal in a way:

1. Proposed solution
2. Alternatives

Just pick one you believe in the most for now (: Right now it's hard to say which one we propose

docs/proposals/20201125-jaeger-traces-store.md
bwplotka: I think UI would be fine, we do that with Thanos TBH, right @onprem @matej-g @jessicalins ?

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+#### Expose only Jaeger Query API
+
+Jaeger query API would be exposed in the Observatorium API and users would have to deploy Jaeger UI and use Observatorium API as a data source.
+
+Jaeger query exposes multiple HTTP and gRPC APIs. The Jaeger UI uses HTTP API. This API does not have OpenAPI schema, and it is only defined in [http_handler.go](https://github.com/jaegertracing/jaeger/blob/0dd3e2da0579caed9e24ad2782a1f638ad63214d/cmd/query/app/http_handler.go#L119). The API uses this [data model](https://github.com/jaegertracing/jaeger/blob/9cd7a7ec1aa43b24a8a970eb5b393ca2ffd98a5d/model/json/model.go#L52).
+
+#### Expose Jaeger UI
+
+In addition to exposing Jaeger query API the Observatorium could deploy Jaeger UI as well. At the moment Observatorium does not expose any UI and it is expected that users deploy UI locally.
+
```
bwplotka, line 106: Would be awesome to elaborate on alternatives. I see one major alternative: Tempo. Even you are against of it since for first glance it sounds like a new solution and lacking indexing, do you mind starting some alternative section about it and we can expand it together in terms of pros / cons

bwplotka (commented):
Great job, thanks! Looking forward to having traces in Observatorium 💪🏽 Some comments.

docs/proposals/20201125-jaeger-traces-store.md
matej-g: While it's true we might be the biggest known user, I believe we still should have proposals which are formulated neutrally and take into account needs to the community as a whole. I'm not sure if we justify choices in other proposals by saying Red Hat uses a particular technology - I believe not, but I might be wrong. For me, it would be a stronger statement on why we decided to support a particular storage solution.

docs/proposals/20201125-jaeger-traces-store.md
matej-g: At minimum, adding tenant requires to update the tenant configuration (see [example](https://github.com/observatorium/api/blob/main/examples/tenants/two.yaml)) and this currently necessitates to bump any running instances of API to make the new config take effect. My point was regarding that fact exactly - the user will need to first bump API and then introduce the deployment change on the tracing backend (or rather vice versa). But I'm not sure if with the 'one tenant per instance' model we can improve on this.

docs/proposals/20201125-jaeger-traces-store.md
matej-g: As I mentioned in my comment above, I believe it depends on the uses we foresee. For Thanos, we expose _both_ UI and query endpoint(s) (but because for metrics it makes sense, depending on use case: whether data is being queried as a data source via the `query` / `query_range` endpoint or a tenant is accessing the Thanos UI in a browser to run queries).

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+This proposal adds Jaeger deployment with a persistent storage (e.g. Elasticsearch) to store trace data and it as well defines a query API to search for traces.
+
+## Why
+
+Traces store adds a crucial capability to Observatorium to persist and query tracing data.
+
+## Goals
+
+* Configure OpenTelemetry collector to export data to Jaeger collector.
```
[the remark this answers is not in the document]
pavolloffay: The https://github.com/observatorium/observatorium/pull/443 will add minimal manifests to deploy OTELcol (e.g. to deploy it optionally and log to stdout). 

When executing this proposal the manifest will be changed to report data to Jaeger or configure any necessary things for multi-tenancy.

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: ES support

https://docs.openshift.com/container-platform/4.9/logging/cluster-logging-release-notes.html
https://access.redhat.com/support/policy/updates/openshift#logging
https://catalog.redhat.com/software/containers/openshift-logging/elasticsearch-rhel8-operator/5fd118c5af03b18dd4ff3fc1?container-tabs=gti

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: We need OTELcol to transform from OTLP to Jaeger format.

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: correct, exposing the  UI needs the Query API which is mentioned in the step above.

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: I am not opposed to restructuring the document but the proposed solution is mentioned in the action plan.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+## How
+
+The how is split into the following sections:
+* Jaeger configuration - describes how Jaeger components are deployed.
+* Jaeger Query - describes how users access data stored in Jaeger.
+
+### Jaeger configuration
+
+Currently, the most adopted and supported storage in Jaeger is Elasticsearch. Red Hat also supports Jaeger with Elasticsearch (6.x) as part of the OpenShift product. Therefore, this proposal assumes Elasticsearch will be used as a storage backend. Distributed tracing team at Red Hat is looking at alternative storages hence the storage technology can change in the future.
+
```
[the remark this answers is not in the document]
pavolloffay: I am happy to list any drawbacks that you have in mind. Elasticsearch depends on fast disk storage similarly as Prometheus/thanos/loki/tempo.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,122 @@
...
+This deployment topology deploys a single Jaeger instance for all tenants with a single Elasticsearch instance. The data for each tenant would contain a unique attribute to identify a tenant. The label would be dynamically injected in the Observatorium API service or in the OpenTelemetry collector.
+
+This deployment strategy will not work with the current Jaeger components. Here are few issues, and the list might not be complete:
+* Service and operation name API. The service query API nor storage does not expose/store labels.
+* Dependency/service architecture diagram. To support this feature dependency schema, query, collector and Spark aggregation would have to change.
+* Data retention/TTL configurable pre tenant. The data retention in Elasticsearch is configurable per index.
+
+##### OpenTelemetry collector configuration
+
+The OpenTelemetry collector is configured with a single exporter to export data to a single Jaeger collector.
```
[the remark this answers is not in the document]
pavolloffay: Jaeger collector does not injest OTLP.

pavolloffay:
PR updated based on the discussion we had during the community meeting.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### Expose Jaeger UI
+
+In addition to exposing Jaeger query API the Observatorium could deploy Jaeger UI as well. At the moment Observatorium does not expose any UI and it is expected that users deploy UI locally.
+
+## Alternatives
+
+### Grafana Tempo
+
+[Grafana Tempo](https://github.com/grafana/tempo) is an alternative storage that Observatorium project could use to store traces. The project is inspired and similar to the Prometheus TSDB, Thanos and Grafana Loki, therefore could it could be a good fit for Obeservatorium. The storage is split into two parts: 1. low retention requiring fast local disk (e.g. SSD) to store recent data and 2. object store (S3 compatible) to store historical data. Tempo is a distributed system with [multiple components](https://grafana.com/docs/tempo/latest/operations/architecture/): distributor, ingester, compactor, storage, querier. Distributor uses parts of the OpenTelemetry collector codebase to receive data in multiple formats (OTLP, Jaeger, Zipkin).
```
matej-g, line 141: Might be worth also mentioning it support multitenancy (https://grafana.com/docs/tempo/latest/configuration/multitenancy/) which is one of the features we're actually missing for Jaeger.

docs/proposals/20201125-jaeger-traces-store.md
```diff
@@ -0,0 +1,161 @@
...
+
+The project at the moment supports [search](https://grafana.com/docs/tempo/latest/getting-started/tempo-in-grafana/#tempo-search) only for recent data held in memory with retention about 15 minutes, therefore it is not suitable as a storage for Jaeger that exposes richer search API that supports search by attributes, time range and service name.
```
matej-g, line 143: I glanced over the Tempo documentation and releases, it looks like they are very close to supporting full back-end search (see https://github.com/grafana/tempo/pull/1174), looks like this is slated for the `1.3` release which is next. So this might soon not be a concern at all?

pavolloffay: The proposal is based on what our team currently supports as part of our product. If we will in the future support tempo we will make sure it is used in Observatorium as well.

docs/proposals/20201125-jaeger-traces-store.md
pavolloffay: added multitenant to the text

docs/proposals/20201125-jaeger-traces-store.md
bwplotka: 👍🏽

bwplotka: I would add something along the lines: 

> 
```suggestion
The project at the moment supports [search](https://grafana.com/docs/tempo/latest/getting-started/tempo-in-grafana/#tempo-search) only for recent data held in memory with retention about 15 minutes, therefore it is not suitable as a storage for Jaeger that exposes richer search API that supports search by attributes, time range and service name. NOTE: We might need to revisit this option once [search feature](grafana/tempo#1174) is production ready.

On top of the above, the current tracing team proficiency is around Elastic and Jaeger development. Given that it makes sense to postpone this alternative and revisit it later once there are more arguments supporting this case.
```

bwplotka (approved):
Thanks for this and massive thanks for checking alternatives (Tempo).

Small nits, otherwise it's LGTM. 

**Disclaimer:** I have doubts if this solution will be highly effective (both cost, performance and support burden) on scale & long term e.g to operate bigger tracing SaaS in the RH world. Plus with monitoring team experience and skillset, I think we would have chosen differently, given our presence in TSDB-like/object storage systems (e.g Tempo). Despite this view, to me looks like @pavolloffay made great due diligence vs what the current skill set, experience and capabilities tracing team can offer. We don't have full visibility on that, that's why for us, (reviewers from the monitoring part) the reasons for this decision might be harder to notice. I trust you @pavolloffay  and your team on the judgement here, and I think we can pull pretty decent tracing SaaS with Elastic + Jaeger of medium size, which is our goal for now. 👍🏽 

Nothing blocks us to experiment, iterate with different backends too (even if we would need to keep current Jaeger read API)

so LGTM (:

bwplotka:
Can you run `make docs` and commit changes?

bwplotka:
Thanks LGTM. I pinged my team to make some last pass, and if no objections will be there, we will merge tomorrow!

matej-g (approved):

ianbillett (approved):

PR #445, opened 2021-12-09
Use the new custom resource definition of prometheus operator

tareqmamari:
This is used by the e2e tests, the path to the custom resource
definitions has been changed.

### DIFF
--- a/configuration/tests/e2e.sh
+++ b/configuration/tests/e2e.sh
@@ -26,8 +26,8 @@ dex() {
 }
 
 deploy() {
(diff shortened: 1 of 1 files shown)


periklis (approved):

matej-g (approved):

matej-g:
@periklis Could please re-run CircleCI? Looks like I'm not able to retrigger it.

squat:
let's not only update the manifests but also pin them to a specific commit so that we avoid these kinds of failures in the future.
However, even when fixing this issue, it turns out there were more issues that broke the e2e tests, as explained in #448.
Closing in favor of #448, which fixes the issues from this PR and also fixes other underlying bugs in the e2e tests
(thanks to @saswatamcode for pointing out the duplication with this PR)

PR #449, opened 2021-12-15
Pluggable authorizers proposal

tareqmamari:
Propose refactoring the current authorization providers setup to provide
pluggable based approach to make it flexible to onboard custom
authorization, this proposal would help new adopters to onboard
their custom authorization implementation.

Signed-off-by: Neyyan Sony <neyysony@in.ibm.com>

### DIFF
--- a/docs/proposals/drafts/20211106-plugable-authorizers.md
+++ b/docs/proposals/drafts/20211106-plugable-authorizers.md
@@ -81,7 +81,7 @@ In this approach, the generic interface for authorization providers (Authorizers
 
 This approach is easy to implement, very flexible in terms of adding more Authorizers without any additional implementation and integration complexity. However, there are few concerns when it comes to trust and security, since Observatorium API would consume a pre-built plugin binary as an authorizer.
 
-**Reference**: [Writing Modular Go Programs with Plugins](https://medium.com/learning-the-go-programming-language/writing-modular-go-programs-with-plugins-ec46381ee1a9)
+**Reference**: [Go Plugin Package](https://pkg.go.dev/plugin)
 
 ## Conclusion
 
--- a/docs/proposals/drafts/20211106-plugable-authorizers.md
+++ b/docs/proposals/drafts/20211106-plugable-authorizers.md
@@ -14,7 +14,7 @@ We propose a pluggable authorizers support to enable plugging in custom authoriz
 
 ## Why
 
-This proposed approach allows the adopters of Observatorium to introduce and use their own custom in-process authorizers' implementation. Currently, adopters can build their own OPA integration, which means introducing another component where OPA server/endpoints are running. In one hand, it allows adopters to use their own custom implementation, however, it also comes with additional efforts such as managing the new OPA-integration component's lifecycle. There are adopters, who prefer to have custom in-process authorizers other than using OPA-integration for different reasons, one important reason besides avoid managing a new component, in-process authorizers tend to work better for medium to large workloads.
+This proposed approach allows the adopters of Observatorium to introduce and use their own custom in-process authorizers' implementation. Currently, adopters can build their own OPA integration, which means introducing another component where OPA server/endpoints are running. On one hand, it allows adopters to use their own custom implementation, however, it also comes with additional efforts such as managing the new OPA-integration component's lifecycle. There are adopters, who prefer to have custom in-process authorizers other than using OPA-integration for different reasons, one important reason besides avoiding managing a new component, in-process authorizers tend to work better for medium to large workloads.
 
 ## Goals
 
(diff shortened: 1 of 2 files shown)


docs/proposals/drafts/20211106-plugable-authorizers.md
```diff
@@ -0,0 +1,97 @@
...
+* **Owners**
+  * @tareqmamari
+
+## TL; DR
+
+We propose a pluggable authorizers support to enable plugging in custom authorizers in a self-register manner besides the already supported authorizers in Observatorium API.
+
+## Why
+
+This proposed approach allows the adopters of Observatorium to introduce and use their own custom in-process authorizers' implementation. Currently, adopters can build their own OPA integration, which means introducing another component where OPA server/endpoints are running. In one hand, it allows adopters to use their own custom implementation, however, it also comes with additional efforts such as managing the new OPA-integration component's lifecycle. There are adopters, who prefer to have custom in-process authorizers other than using OPA-integration for different reasons, one important reason besides avoid managing a new component, in-process authorizers tend to work better for medium to large workloads.
```
matej-g, line 17:
```suggestion
This proposed approach allows the adopters of Observatorium to introduce and use their own custom in-process authorizers' implementation. Currently, adopters can build their own OPA integration, which means introducing another component where OPA server/endpoints are running. On one hand, it allows adopters to use their own custom implementation, however, it also comes with additional efforts such as managing the new OPA-integration component's lifecycle. There are adopters, who prefer to have custom in-process authorizers other than using OPA-integration for different reasons, one important reason besides avoiding managing a new component, in-process authorizers tend to work better for medium to large workloads.
```

docs/proposals/drafts/20211106-plugable-authorizers.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Goals
+
+- Do not change the current common interface for the authorizers.
+- Provide a self-registering mechanism of the authorizers so that the adopters do not need to patch the current codebase to introduce their own custom authorization logic.
+- The new changes must be backward-compatible. In other words, the current authorizer implementation and configuration must not be broken by this change and should be supported to use the new pluggable based implementation.
+
+## How
+
+We propose to extend rego evaluation logic to load custom built-in functions in Go, in this way, adopters can implement their own authorization logic in go and then call it in Rego. In addition, we aim to provide a self-registering mechanism, so that introducing any new custom built-in function does not necessary require to do any patch/change in the existing codebase, instead, only the plugin's go source file is needed other than patching the other source files to add the plugin. For that, we propose to use importing for side-effect mechanism, where a protected map of authorizer's provider type and their custom built-in function can be used for such scenario.
```
matej-g, line 27:
```suggestion
We propose to extend Rego evaluation logic to load custom built-in functions in Go. This way, adopters can implement their own authorization logic in Go and then call it in Rego. In addition, we aim to provide a self-registering mechanism, so that introducing any new custom built-in function does not necessary require to do any patch/change in the existing codebase. Instead, only the plugin's Go source file is needed other than patching the other source files to add the plugin. For that, we propose to use importing for side-effect mechanism, where a protected map of authorizers' provider types and their custom built-in function can be used for such scenario.
```

docs/proposals/drafts/20211106-plugable-authorizers.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Implementation Options
+1. Custom Rego Built-in Functions in Go
+2. Importing for Side-Effect
+3. Go Plugin Module
+
+## 1. Custom Rego Built-in Functions in Go
+
+Instead of totally introducing a new authorizer, OPA inprocess authorizer ( using Rego ) can be extended through [custom rego built-in functions in go](https://www.openpolicyagent.org/docs/latest/extensions/#custom-built-in-functions-in-go). In this approach, custom auhtorization go logic can be introduced as a custom built-in rego function in go, which then can be called in the rego logic. Once rego calls this function, the go implementation of it, would be executed. To make it pluggable, importing for side-effects is used, similar to what is proposed in the second option.
```
matej-g, line 36:
```suggestion
Instead of totally introducing a new authorizer, OPA inprocess authorizer (using Rego) can be extended through [custom Rego built-in functions in go](https://www.openpolicyagent.org/docs/latest/extensions/#custom-built-in-functions-in-go). In this approach, custom auhtorization go logic can be introduced as a custom built-in Rego function in go, which then can be called in the rego logic. Once Rego calls this function, the go implementation of it, would be executed. To make it pluggable, importing for side-effects is used, similar to what is proposed in the second option.
```

docs/proposals/drafts/20211106-plugable-authorizers.md
```diff
@@ -0,0 +1,97 @@
...
+
+To onboard custom authorization with this approach, the following is needed:
+1. Add a new go file with the custom authorization logic.
+2. The new custom built-in function should register itself through an onboarding map.
+3. All registered functions will be loaded into the rego env during the evaluation.
```
matej-g, line 41:
```suggestion
3. All registered functions will be loaded into the Rego env during the evaluation.
```

docs/proposals/drafts/20211106-plugable-authorizers.md
```diff
@@ -0,0 +1,97 @@
...
+## Conclusion
+
+We propose to use the first approach (Custom Rego Built-in Functions in Go), as it appears to be the most flexible approach to follow and require no additional changes to onboard a new authorizer or additional resources to manage.
+
+It worth mentioning, that the new changes must be backward compatible, in other words, it must not break neither the existing configuration nor the current authorization flows.
+
+## Action plan
+- [ ] Review and finalize this proposal document.
+- [X] Provide a Proof of Concept (PoC) for pluggable authorizers.
+- [ ] The community to review and provide feedback on the PoC Pull Requests.
```
matej-g, line 95: I think we can check these since the feature is merged already
```suggestion
- [X] Review and finalize this proposal document.
- [X] Provide a Proof of Concept (PoC) for pluggable authorizers.
- [X] The community to review and provide feedback on the PoC Pull Requests.
```

matej-g (approved):
Looking good :+1: , added few minor nits

tareqmamari:
@matej-g Thanks for your comments, this proposal should be ready now.

matej-g (approved):

matej-g:
No more comments and this proposal is already implemented - I'm merging it. Thanks @tareqmamari!

matej-g:
Which reminds we could have moved this directly out of the `drafts` directory...

PR #453, opened 2022-01-07
Add proposal for Prometheus Alerting in Observatorium

onprem:
This proposal talks about adding Prometheus alerting capabilities to Observatorium.

Signed-off-by: Prem Saraswat <prmsrswt@gmail.com>


docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+    interval: 5s
+    rules:
+      - alert: HighRequestLatency
+        expr: job:request_latency_seconds:mean5m{job="second"} > 0.5
+        for: 10m
+        labels:
+          team: alpha
+```
+
+Such rule will be saved through Observatorium API, then Rule-Store to object storage for the specified tenant. It can be then listed through a similar endpoint and GET.
```
moadz, line 67:
```suggestion
Such a rule will be saved through Observatorium API, then Rule-Store to object storage for the specified tenant. It can be then listed through a similar endpoint via GET.
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Alerts
+
+Such configuration will be then eventually synchronized with tenant’s Rulers using ConfigSync. The Ruler itself has to communicate on 4 levels. All connections are presented in the diagram below.
```
moadz, line 71: > eventually synchronized 

At what frequency? What is the `ConfigSync` core loop. Is there a universal `ConfigSync` for all Alertmanagers/tenant combinations?

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+
+![ruler-connections-diagram](../assets/alerting-ruler.png)
+
+1. Alerts will be evaluated based on periodic requests to any replica of  HTTP Query API of Query-frontend (Question: Or querier?).
```
moadz, line 75: > (Question: Or querier?).

Why not directly through Query FE?

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+![ruler-connections-diagram](../assets/alerting-ruler.png)
+
+1. Alerts will be evaluated based on periodic requests to any replica of  HTTP Query API of Query-frontend (Question: Or querier?). 
+2. Triggered alerts will be pushed to all replicas of  Alertmanager (discussed later) through AM HTTP API.
+3. It exposes gRPC Rules API which is consumed by Queriers.
+4. It saves samples to any replica of Receiver Router (or RouterIngestor). 
+
+### Alert eval status
+
+We expect Rulers to be connected to Queriers through gRPC Rules API (NOTE, not Store API as we have now). This allows queries to be able to present federate views through API. In order to fulfil the full story we also need:
```
moadz, line 82:
```suggestion
We expect Rulers to be connected to Queriers through gRPC Rules API (NOTE, not Store API as we have now). This allows queries to be able to present federated views through the API. In order to fulfil the full story we also need:
```

moadz: > This allows queries to be able to present federated views through the API

Observatorium, rules or querier HTTP alerts API?

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+
+* Implement Matching / Filtering by label/tenant on Alerts HTTP API of Querier.
```
moadz, line 84:
```suggestion
* Implement matching/filtering by label/tenant on Alerts HTTP API of Querier.
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+* Expose read Alert API on Observatorium API
+
+###  Alert Routing
+
+Triggered alerts are essential to be reliably forwarded to correct notifications points as soon as they are triggered. We propose deploying 3-replicas Alertmanager HA, through our configuration (Question: Without Prometheus Operator).
```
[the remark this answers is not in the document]
moadz: > (Question: Without Prometheus Operator).

Can you elaborate on the question?

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+NOTE that here we will expect users to pass SECRETS (keys, TLS, passwords etc) to 3rd party receivers. We propose to at the moment save it in object storage in plain text (Question: This sounds bad, what we can do better here?).
+
+We could allow all read API as well of write of Inhibits and Silences of Alertmanager (https://github.com/prometheus/alertmanager/blob/main/api/v2/openapi.yaml) by adding auth and multi-tenancy layer, but we propose to deal with that in a separate epic/proposal.
+
+## Alternatives
+
+### Alertmanager per tenant
+
+
+### ??
```
moadz, line 106: ???

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+
+Triggered alerts are essential to be reliably forwarded to correct notifications points as soon as they are triggered. We propose deploying 3-replicas Alertmanager HA, through our configuration (Question: Without Prometheus Operator).
+
+We propose defining an additional HTTP API for routing on `/api/metrics/v1/{tenant}/api/v1/routing/raw` path in Observatorium API.
+
+Types based on [Prometheus Operator AM Config](https://github.com/prometheus-operator/prometheus-operator/blob/9c0db5656f04e005de6a0413fd8eb8f11ec99757/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L69).
+
+Paste here TBD.
+
+NOTE that here we will expect users to pass SECRETS (keys, TLS, passwords etc) to 3rd party receivers. We propose to at the moment save it in object storage in plain text (Question: This sounds bad, what we can do better here?).
```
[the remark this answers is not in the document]
moadz: Secondary question, I guess we don't have RBAC per tenant inside the Observatorium namespace?

docs/proposals/20220105-observatorium-alerting.md
yeya24, line 84: Does this make sense to bring this ability to prometheus/thanos community, too? Some users already have such use case.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+We expect Rulers to be connected to Queriers through gRPC Rules API (NOTE, not Store API as we have now). This allows queries to be able to present federate views through API. In order to fulfil the full story we also need:
+
+* Implement Matching / Filtering by label/tenant on Alerts HTTP API of Querier.
+* Expose read Alert API on Observatorium API
+
+###  Alert Routing
+
+Triggered alerts are essential to be reliably forwarded to correct notifications points as soon as they are triggered. We propose deploying 3-replicas Alertmanager HA, through our configuration (Question: Without Prometheus Operator).
+
+We propose defining an additional HTTP API for routing on `/api/metrics/v1/{tenant}/api/v1/routing/raw` path in Observatorium API.
```
yeya24, line 91: If there is such write API for alerting routing, then I expect users can use read API to get their routing configuration?
Can you add it to the proposal? How would it work? Is there a alert manager UI for users to see their routes?

spaparaju (commented):
Another option for `Rulers` is reach 'Thanos Queriers' through Observatorium API to reuse the tenancy semantics at Observatorium API  and future-proof `Rulers` to work across multiple instances of Observatorium (possibly at multiple regions)

docs/proposals/20220105-observatorium-alerting.md
ianbillett: +1 I would expect the routing API to support both GET and PUT operations.

Providing an Alertmanager UI is listed under the non-goals section above.

matej-g (commented):
So far it's looking sound :+1: I'd have another look once the current comments and remaining to-dos in the doc are addressed

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+
+Types based on [Prometheus Operator AM Config](https://github.com/prometheus-operator/prometheus-operator/blob/9c0db5656f04e005de6a0413fd8eb8f11ec99757/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L69).
```
ianbillett, line 93: Are we going to use these types directly? Or modify them is some cases?

docs/proposals/20220105-observatorium-alerting.md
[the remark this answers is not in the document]
ianbillett: Nope - Observatorium RBAC only exists as a concept within the API itself.

docs/proposals/20220105-observatorium-alerting.md
ianbillett: ¿¿¿

ianbillett (commented):
Nice start @onprem - the bulk of the thinking is there and seems sensible :+1: 

The biggest bits that are missing for me at the moment are about how we construct the routing trees that alert manager consumes. Users will write routing table configuration to us, and we will store that somehow, but how do we create the final blob of YAML that we sync into AM?

Also - I would be careful about drawing the line between the Rules API proposal and this Alerting proposal, which lets us keep this proposal much more focused. For example - when the Rules API is implemented Ruler will naturally start evaluating alerts, all we need to do is configure Ruler to point at an Alertmanager URL.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+
+Observatorium did not have alerting defined before.
+
+## Goals
+
+Definition: 
+* Alert Routing: Configuration of Alertmanager related to receiver target config and its routing. 
+
+Goals:
+* Tenant should be able to use alerting rules.
```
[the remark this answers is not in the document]
ianbillett: That's already taken care of in the Rules API :)

ianbillett:
Just chatting with @onprem - there will likely be some lag in responding to all the comments in here as we're currently focused on getting the Rules API into staging. That is a hard dependency of this proposal & unblocks other workstreams. Watch this space :)

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
+# Observatorium Alerting
```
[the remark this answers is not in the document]
onprem: The issues are public so I guess we can have them referenced here.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+  * Alerting Rules via GET and PUT via HTTP (Rules API)
+* Tenants should be able to read what is the status of configured alerts (Prometheus Alert API GET)
+* Tenants are able to route alerts:
+  * Alerting Routing via GET and PUT via HTTP API.
+* Both alerts and alerting rules should be multi-tenant.
+
+## Non-Goals
+
+* (?) UI for Alerts/Rules, Alert Status, AM Routing, AM Alerts …
+* (?) Alertmanager Alerts API, Config/Status
```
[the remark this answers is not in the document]
onprem: The question marks indicates an open question. We need to check and decide if we want these.

docs/proposals/20220105-observatorium-alerting.md
onprem: > At what frequency?

It should be configurable, but the default is 60s in Thanos Rule Syncer right now. 

> If that is the case - we need to explain why we think this is good idea, and why thanos-ruler-syncer is not currently sufficient for our use-case here.

Yes, we will add a section explaining the use-case.

docs/proposals/20220105-observatorium-alerting.md
bwplotka: This is placeholder for YOUR ideas for alternatives if any =D

docs/proposals/20220105-observatorium-alerting.md
onprem: > Why not directly through Query FE?

What @matej-g is saying makes sense. It's an implementation detail we can figure out later.

docs/proposals/20220105-observatorium-alerting.md
onprem: We are talking about Thanos [gRPC Rules API](https://github.com/thanos-io/thanos/blob/main/pkg/rules/rulespb/rpc.proto), which is essentially same as [Prometheus Rules HTTP API](https://prometheus.io/docs/prometheus/latest/querying/api/#rules).

We need to add a TODO because we are missing the [Prometheus Alerts API](https://prometheus.io/docs/prometheus/latest/querying/api/#alerts) in Thanos.

docs/proposals/20220105-observatorium-alerting.md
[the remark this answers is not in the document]
onprem: We have many alternatives, and we will explain them. But TLDR: K8s secrets is just another storage we didn't depend on before, it requires special permissions (to create secrets). We could use vault if we even have permissions. But it's an open question we need to decide.

docs/proposals/20220105-observatorium-alerting.md
onprem: I think yes. @saswatamcode is already investigating this.

docs/proposals/20220105-observatorium-alerting.md
onprem: Yes, we will add the `GET` to the proposal as well.

onprem:
> Another option for `Rulers` is reach 'Thanos Queriers' through Observatorium API to reuse the tenancy semantics at Observatorium API and future-proof `Rulers` to work across multiple instances of Observatorium (possibly at multiple regions)

This is indeed possible but I think this might involve more work. Going through Obs. API means we have take care of authenticating all requests, which might involve running token-refresher as sidecar for every Ruler, configuring the tenant secrets, etc.

docs/proposals/20220105-observatorium-alerting.md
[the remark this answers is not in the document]
onprem: We mean
> Tenants must be able to write alert routing configuration to the Observatorium API.

Thanks, I'll update this line to be more specific.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+Such rule will be saved through Observatorium API, then Rule-Store to object storage for the specified tenant. It can be then listed through a similar endpoint and GET.
+
+### Alerts
+
+Such configuration will be then eventually synchronized with tenant’s Rulers using ConfigSync. The Ruler itself has to communicate on 4 levels. All connections are presented in the diagram below.
+
+![ruler-connections-diagram](../assets/alerting-ruler.png)
+
+1. Alerts will be evaluated based on periodic requests to any replica of  HTTP Query API of Query-frontend (Question: Or querier?). 
+2. Triggered alerts will be pushed to all replicas of  Alertmanager (discussed later) through AM HTTP API.
```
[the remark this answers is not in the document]
onprem: Yes, Thanos Ruler pushes the firing alerts to AM.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,124 @@
...
+3. It exposes gRPC Rules API which is consumed by Queriers.
```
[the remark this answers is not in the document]
onprem: The HTTP Rules API exposed by Querier is backed by a gRPC Rules API exposed by all Thanos Rulers.

docs/proposals/20220105-observatorium-alerting.md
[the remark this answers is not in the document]
onprem: > Is this so that users looking at the Thanos Query UI will only see the alerts that matter to them?

Yes, exactly this.

docs/proposals/20220105-observatorium-alerting.md
onprem: This is an implementation detail. It'd be nice if we can use them directly but we might need to modify them in some cases. One such case might be modifying them to fit our tenancy model.

docs/proposals/20220105-observatorium-alerting.md
simonpasquier, line 89: What's the motivation for having 3 replicas here? We used to deploy 3 replicas with the OpenShift monitoring stack but it's 2 replicas now.
Historically it was assumed that 3 replicas provided more resiliency but with proper anti-affinity rules, pod disruption budget and persistent storage, 2 replicas is sufficient IMHO.

docs/proposals/20220105-observatorium-alerting.md
simonpasquier: Happy to have a discussion about it here or at github.com/prometheus-operator/prometheus-operator :)

docs/proposals/20220105-observatorium-alerting.md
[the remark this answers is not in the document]
simonpasquier: This is a pretty important point to clarify.

docs/proposals/20220105-observatorium-alerting.md
bwplotka: Nice, thanks, noted. If 2 replica feels alright, it makes sense then.

docs/proposals/20220105-observatorium-alerting.md
bwplotka: All addressed, except this:

> My understanding is that when the Rules API is active, alerts will already be evaluated in Thanos Ruler? The outcome is that Ruler will create new rules with __name__ = ALERT?

@ianbillett could you elaborate on what do you mean? Outcome is that Recording Rule is evaluated and it's output is pushed to  receiver. Similarly, Alert is evaluated and if triggered it is sending `__name__ = ALERT?` and pushing Alert to AM.

docs/proposals/20220105-observatorium-alerting.md
bwplotka: I actually have idea how to avoid this extra work, added explanation!

bwplotka (commented):
Updated proposal, PTAL 🤗

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+### ConfigSync
+
+The first decision we propose is to move https://github.com/observatorium/rules-objstore to be pure Go library, potentially in new https://github.com/observatorium/config-client repo. Such library would need to support not only saving and retrieving Rules but also Alerting Routing configuration (and potentially silencer, inhibits and other user configuration in the future). Since alerting routing requires secrets we might need to implement various backends in this client (that's why no `objstore` suffix. ). See [chapter about secrets](#managing-user-secrets) for more details. 
+
+Similarly, we propose to move https://github.com/observatorium/thanos-rule-syncer to https://issues.redhat.com/browse/MON-2129[https://github.com/observatorium/config-sync] that will understand Prometheus Rules configuration, Prometheus Alert Routing configuration and in future silences, inhibitors tpp etc. ConfigSync would then import https://github.com/observatorium/config-objstore. This will allow us to have single focused codebase for all different user configurations we will need. This will allow us to maintain consistency around using it, monitoring, debugging, using secrets etc.
```
jessicalins, line 80:
```suggestion
Similarly, we propose to move https://github.com/observatorium/thanos-rule-syncer to https://github.com/observatorium/config-sync (related ticket: https://issues.redhat.com/browse/MON-2129) that will understand Prometheus Rules configuration, Prometheus Alert Routing configuration and in future silences, inhibitors tpp etc. ConfigSync would then import https://github.com/observatorium/config-client. This will allow us to have single focused codebase for all different user configurations we will need. This will allow us to maintain consistency around using it, monitoring, debugging, using secrets etc.
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+We propose potential https://github.com/observatorium/config-sync to be deployed as a sidecar to each component needing configuration to be present and reloaded dynamically.
+
+As a consequence of this decision I propose Observatorium API to also import https://github.com/observatorium/config-objstore drectly to allow User APIs to manage those items (Rules and Alert routing configuration so far).
```
jessicalins, line 84:
```suggestion
As a consequence of this decision I propose Observatorium API to also import https://github.com/observatorium/config-client directly to allow User APIs to manage those items (Rules and Alert routing configuration so far).
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+The Ruler itself has to communicate on 4 levels. All connections are presented in the diagram below.
+
+![ruler-connections-diagram](../assets/alerting-ruler.png)
+
+1. Alerts will be evaluated based on periodic requests to any replica of HTTP Query API, potentially through Query-frontend. 
+2. Ruler evaluates Alerts that can be triggered. Triggered alerts will be pushed to all replicas of the Alertmanager (discussed later) through AM HTTP API.
+3. It exposes gRPC Rules API which is consumed by Queriers. Querier than can show this API via it's HTTP Rule API.
+4. It saves samples to any replica of Receiver Router (or RouterIngestor). 
+
+### Rules API (non row one)
```
jessicalins, line 97:
```suggestion
### Rules API (non raw one)
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+We expect Rulers to be connected to Queriers through [gRPC Rules API](https://github.com/thanos-io/thanos/blob/main/pkg/rules/rulespb/rpc.proto#L29) (NOTE, not Store API as we have now). This allows queries to be able to present federated views of Recording and Alerting Rules loaded and active alerts through the [Promethus Rules HTTP API](https://prometheus.io/docs/prometheus/latest/querying/api/#rules). 
+
+This API is different but a bit overlapping with `/api/metrics/v1/{tenant}/api/v1/rules/raw` which represents what was configured (not loaded).
+
+In order to fulfil the full story we also need:
```
jessicalins, line 103:
```suggestion
In order to fulfill the full story we also need:
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+We propose adding different backends depending on user preference to https://github.com/observatorium/config-client, but for a start we propose adding support to reading secrets via Vault. Unfortunately we have to delegate storing user credentials to 3rd party dependency. See alternatives for different ideas.
+
+## Alternatives
+
+### Alertmanager per tenant
+
+### Having ConfigClient in ConfigSync repo
+
+Since config-client is a dependency for config-sync service, we could simplify development by keeping it in one place. While making sense, this is akwkard with Observatorium API having to import config-client code from config-sync repo. Still this might be something we could consider to simplify release and build process.
```
jessicalins, line 139:
```suggestion
Since config-client is a dependency for config-sync service, we could simplify development by keeping it in one place. While making sense, this is awkward with Observatorium API having to import config-client code from config-sync repo. Still this might be something we could consider to simplify release and build process.
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Different Backend for Secrets
+
+In theory Vault is heavy dependeny. Since we run on Kubernetes, we could just write Kubernetes secrets and store it there. There are some disadventages:
```
jessicalins, line 143:
```suggestion
In theory Vault is a heavy dependency. Since we run on Kubernetes, we could just write Kubernetes secrets and store it there. There are some disadvantages:
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+* Does it really solve secret management in practice? Those secrets are not encrypted
+* We add strong dependency on Kubernetes.
+* Can our deployment have access to writing and reading Kubernetes secrets.
+
+This option has to explored more.
```
jessicalins, line 149:
```suggestion
This option has to be explored more.
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+## FAQ
+
+### Why thanos-ruler-syncer is not currently sufficient for our use-case?
```
jessicalins, line 153:
```suggestion
### Why thanos-rule-syncer is not currently sufficient for our use-case?
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+Thanos rule syncer only support syncing Rules. Does not support AM Routing configuration. It also requires separate service thanos-objstore to be deployed. This unnecessary microservice architecture will become a problem and a fuss to operate and maintain. We don't need to scale those components separatedly, so I propose to keep it in a single binary for now and expected future.
```
jessicalins, line 155: makes sense :+1:

jessicalins (commented):
cool! :a: added a few questions and formatting suggestions

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+        for: 10m
+        labels:
+          team: alpha
+```
+
+Such a rule will be saved through Observatorium API, via Rule-Store to object storage for the specified tenant. It can be then listed through a similar endpoint and GET. Such configuration will be then eventually synchronized with tenant’s Rulers using RuleSync (1m poll by default). This is what was agreed so far and is being tested in downstream RHOBS (Red Hat Observability Service) deployment of Observatorium. 
+
+### ConfigSync
+
+The first decision we propose is to move https://github.com/observatorium/rules-objstore to be pure Go library, potentially in new https://github.com/observatorium/config-client repo. Such library would need to support not only saving and retrieving Rules but also Alerting Routing configuration (and potentially silencer, inhibits and other user configuration in the future). Since alerting routing requires secrets we might need to implement various backends in this client (that's why no `objstore` suffix. ). See [chapter about secrets](#managing-user-secrets) for more details.
```
matej-g, line 78: I think I understand what is meant here, but since it is already a 'pure' Go library, maybe wording it a bit differently:

```suggestion
The first decision we propose is to move https://github.com/observatorium/rules-objstore to be a generic configuration library, potentially in new https://github.com/observatorium/config-client repo. Such library would need to support not only saving and retrieving Rules but also Alerting Routing configuration (and potentially silencer, inhibits and other user configuration in the future). Since alerting routing requires secrets we might need to implement various backends in this client (that's why no `objstore` suffix. ). See [chapter about secrets](#managing-user-secrets) for more details. 
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Alert Evaluation
+
+The Ruler itself has to communicate on 4 levels. All connections are presented in the diagram below.
+
+![ruler-connections-diagram](../assets/alerting-ruler.png)
+
+1. Alerts will be evaluated based on periodic requests to any replica of HTTP Query API, potentially through Query-frontend. 
+2. Ruler evaluates Alerts that can be triggered. Triggered alerts will be pushed to all replicas of the Alertmanager (discussed later) through AM HTTP API.
+3. It exposes gRPC Rules API which is consumed by Queriers. Querier than can show this API via it's HTTP Rule API.
```
matej-g, line 94: What is meant here by show? Maybe specifying that it's consumed by queriers is enough?

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+4. It saves samples to any replica of Receiver Router (or RouterIngestor).
```
matej-g, line 95: Does it matter whether it's router / ingestor for this proposal?
```suggestion
4. It saves samples to any replica of Receiver. 
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+We expect Rulers to be connected to Queriers through [gRPC Rules API](https://github.com/thanos-io/thanos/blob/main/pkg/rules/rulespb/rpc.proto#L29) (NOTE, not Store API as we have now). This allows queries to be able to present federated views of Recording and Alerting Rules loaded and active alerts through the [Promethus Rules HTTP API](https://prometheus.io/docs/prometheus/latest/querying/api/#rules). 
+
+This API is different but a bit overlapping with `/api/metrics/v1/{tenant}/api/v1/rules/raw` which represents what was configured (not loaded).
+
+In order to fulfil the full story we also need:
+
+* Implement matching / filtering by label/tenant on Rules HTTP API of Querier.
+
+This will allow users looking at the Thanos Rules API (or Query UI) to only see the alerts that matter to them.
```
matej-g, line 107:
```suggestion
This will allow users looking at the Thanos Rules API (or Query UI) to only see the alerts that belong to them.
```

docs/proposals/20220105-observatorium-alerting.md
matej-g, line 97: I'm finding this non-raw vs. raw distinction a bit confusing, maybe we could call this one just gRPC Rules API?

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Alert Evaluation Status
+
+On top of mentioned Rules API, Prometheus also exposes [HTTP Alert API](https://prometheus.io/docs/prometheus/latest/querying/api/#alerts). We never expose it within Observatorium and Thanos and the truth is that Rules API gives exactly that information (on top of other data).
```
matej-g, line 111:
```suggestion
On top of mentioned Rules API, Prometheus also exposes [HTTP Alert API](https://prometheus.io/docs/prometheus/latest/querying/api/#alerts). We never expose it within Observatorium and Thanos, but Rules API in fact already provides exactly that information (on top of other data).
```

docs/proposals/20220105-observatorium-alerting.md
ianbillett: I guess what I mean here is that - when the Rules API is implemented, we don't need to do anything else in this Alerting proposal, because the Ruler is responsible for evaluating the alerts and sending to Alertmanager. All we need to do is tell Ruler where to forward the alerts.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+Because of that we propose to not add this API, which will avoid a lot of work. In future iteration (e.g. if we would add UI), we could add Alert HTTP API in Observatorium which would extract data needed by this API from Rules API.
+
+### Alert Routing
+
+Triggered alerts are essential to be reliably forward to correct notifiers as soon as they are triggered. We propose deploying w-replica Alertmanager HA, through our configuration (without Prometheus Operator help which is usually used in our past deployments).
```
matej-g, line 117:
```suggestion
It is crucial that we are able to reliably forward alerts to correct notifiers as soon as they are triggered. We propose deploying 2-replica Alertmanager HA, through our configuration (without the help of the Prometheus Operator, which was usually used in our past deployments).
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+We propose defining an additional HTTP API for routing on `/api/metrics/v1/{tenant}/api/v1/routing/raw` path in Observatorium API. Path that supports both GET, DELETE and PUT. 
+
+We propose to base the type on [Prometheus Operator AM Config](https://github.com/prometheus-operator/prometheus-operator/blob/9c0db5656f04e005de6a0413fd8eb8f11ec99757/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L69). Proposed adding this API to upstream Alertmanager too (TODO).
+
+We could allow all read API as well of write of Inhibits and Silences of Alertmanager (https://github.com/prometheus/alertmanager/blob/main/api/v2/openapi.yaml) by adding auth and multi-tenancy layer, but we propose to deal with that in a separate epic/proposal.
```
matej-g, line 123: I'd either drop the link or link to concrete lines in the OpenAPI YAML (we can perhaps drop it since it's not focus of this proposal):
```suggestion
We could allow all read API as well write of Inhibits and Silences of Alertmanager by adding auth and multi-tenancy layer, but we propose to deal with that in a separate epic/proposal.
```

docs/proposals/20220105-observatorium-alerting.md
matej-g, line 139: I think we could make this workable if we came up with higher level name (e.g. something like `config-utils`?) and then we can even have separate modules for sync and client.

docs/proposals/20220105-observatorium-alerting.md
matej-g, line 155:
```suggestion
Thanos rule syncer only support syncing Rules. Does not support AM Routing configuration. It also requires separate service `thanos-objstore` to be deployed. This unnecessary microservice architecture will become a problem and a fuss to operate and maintain. We don't need to scale those components separately, so I propose to keep it in a single binary for now and near future. 
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+AM: Alertmanager
+
+## Audience of this proposal
+
+Observatorium Devs and Users.
+
+## How
+
+Within work for [Rules API proposal](https://observatorium.io/docs/proposals/20201019-prometheus-rules-in-observatorium-api.md/) we already added Rules APIs for reading and configuring rules which include both recording and alerting rules.
+
```
matej-g, line 59: Should we add a subsection here, something like `### Current solution`, it got me a bit confused since I thought this was already describing proposed solution.

matej-g (commented):
Some non essential comments, but content-wise this sounds good to me :+1:

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+* **Owners:**
+  * `@onprem` `@bwplotka`
+
+* **Related Tickets:**
+  * https://issues.redhat.com/browse/MON-1761
+
+* **Other docs:*
+  * None 
+
+> TL;DR: For monitoring use cases, especially tenants of the metric signal of Observatorium, users want to be able to configure Prometheus-based alerts that will be evaluated for a given interval. If triggered they should notify the desired target (receiver) (e.g PagerDuty using Alertmanager Configuration). Users should also be able to see the status of all alerts through Prometheus-like alerts HTTP API. Users should also be able to see the status of all alerts through Prometheus-like Alerts HTTP API. This proposal specifies how such a system can be deployed within Observatorium.
```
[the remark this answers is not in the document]
bwplotka: Nicely spotted!

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Goals
+
+Definition: 
+* Alert Routing: Configuration of Alertmanager related to receiver target config and its routing. 
+
+Goals:
+
+* Tenants should be able to read the status of active alerts 
+  * via ([Prometheus Alert API GET](https://prometheus.io/docs/prometheus/latest/querying/api/#alerts))
```
bwplotka, line 30: TODO: Remove, since Rules contains already same info!.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+Because of that we propose to not add this API, which will avoid a lot of work. In future iteration (e.g. if we would add UI), we could add Alert HTTP API in Observatorium which would extract data needed by this API from Rules API.
+
+### Alert Routing
+
+Triggered alerts are essential to be reliably forward to correct notifiers as soon as they are triggered. We propose deploying w-replica Alertmanager HA, through our configuration (without Prometheus Operator help which is usually used in our past deployments).
+
+We propose defining an additional HTTP API for routing on `/api/metrics/v1/{tenant}/api/v1/routing/raw` path in Observatorium API. Path that supports both GET, DELETE and PUT. 
+
+We propose to base the type on [Prometheus Operator AM Config](https://github.com/prometheus-operator/prometheus-operator/blob/9c0db5656f04e005de6a0413fd8eb8f11ec99757/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L69). Proposed adding this API to upstream Alertmanager too (TODO).
```
bwplotka, line 121: TODO: I will start a discussion on the Alertmanager repo.

TODO: Clarify what API should look like exactly, before merging

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+* Both alerts and alerting rules should be multi-tenant.
+
+## Non-Goals
+
+* Tenant should be able to create and edit alerting rules:
+  * Alerting Rules via GET and PUT via HTTP (Rules API)
+  * This is not the goal, because it should be done when recording rule API is done.
+* UI for Alerts/Rules, Alert Status, AM Routing, AM Alerts …
+* Alertmanager Alerts API, Silences, Inhibits API
+  * We should focus on that in next iteration.
```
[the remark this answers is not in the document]
bwplotka: We have it in non-goals, so I guess that's fine? (:

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+* Alert Routing: Configuration of Alertmanager related to receiver target config and its routing. 
+
+Goals:
+
+* Tenants should be able to read the status of active alerts 
+  * via ([Prometheus Alert API GET](https://prometheus.io/docs/prometheus/latest/querying/api/#alerts))
+* Tenants are able to configure alert routing:
+  * via new GET, DELETE and PUT HTTP API
+  * Potentially work with Alertmanager Team on upstream API.
+* Both alerts and alerting rules should be multi-tenant.
```
bwplotka, line 34:
```suggestion
* Both Rules API and AM routing API should be multi-tenant.
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
```
bwplotka, line 35: * Tenant's alerts will be evaluated and triggering proper receivers in AM.

docs/proposals/20220105-observatorium-alerting.md
bwplotka: cc `git fetch origin pull453/head:<local branch name>`

docs/proposals/20220105-observatorium-alerting.md
bwplotka, line 94:
```suggestion
3. Ruler exposes gRPC Rules API which is consumed by Queriers. Querier than can expose multi-tenant HTTP Rules API that can be consumed by Observatorium API.
```

docs/proposals/20220105-observatorium-alerting.md
bwplotka: Hmm both are HTTP in most cases.

One is raw configuration you can GET, PUT and DELETE, second is currently LOADED rules and alerts (+ instances of active alerts) you can GET

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+### Managing User Secrets
+
+Something new in Alertmanager Routing configuration is the fact that Alertmanager can integrate with other systemes ("receivers") to send alerts to. This requires TLS, basic auth or other type of authZ and authN to be configured within this configuration. AM supports it, but it expects secrets to be usually baked in file or available in local AM filesystem.
+
+This means we have to enter into user secret management story which is not trivial. We cannot just store those files in object storage in plain text. Encrypting them in rest sounds easy, but it involves lots of complexity around rotating keys, revoking secrets, process when leaked etc.
+
+We propose adding different backends depending on user preference to https://github.com/observatorium/config-client, but for a start we propose adding support to reading secrets via Vault. Unfortunately we have to delegate storing user credentials to 3rd party dependency. See alternatives for different ideas.
```
bwplotka, line 131: TODO: Check if we AppSRE offers VaultaaS

docs/proposals/20220105-observatorium-alerting.md
bwplotka: Yea.. maybe config-sync would be still better, but keeping this in same repo under this name might even fine.

bwplotka: Let's swap alternative with proposal, so simpler solution wins.

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Alternatives
+
+### Alertmanager per tenant
+
+### Having ConfigClient in ConfigSync repo
+
+Since config-client is a dependency for config-sync service, we could simplify development by keeping it in one place. While making sense, this is akwkard with Observatorium API having to import config-client code from config-sync repo. Still this might be something we could consider to simplify release and build process.
+
+### Different Backend for Secrets
```
bwplotka, line 141: TODO: Alternative: Using our secrets and enabling them to contact receivers?

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Action Plan
+
+The tasks to do in order to migrate to the new idea.
+
+* [ ] Start using stateless Ruler in Observatorium
+* [ ] Implement matching / filtering by label/tenant on Rules HTTP API of Querier.
+* [ ] Move https://github.com/observatorium/rules-objstore to be a library called https://github.com/observatorium/config-client
+* [ ] Add ability to talk to Vault in https://github.com/observatorium/config-client
+* [ ] Develop https://github.com/observatorium/config-sync based on https://github.com/observatorium/config-sync
```
bwplotka, line 165:
```suggestion
* [ ] Develop https://github.com/observatorium/config-sync based on thanos-rule-syncer 
```

docs/proposals/20220105-observatorium-alerting.md
```diff
@@ -0,0 +1,169 @@
...
+* **Owners:**
+  * `@onprem` `@bwplotka`
+
+* **Related Tickets:**
+  * https://issues.redhat.com/browse/MON-1761
+
+* **Other docs:*
+  * None 
+
+> TL;DR: For monitoring use cases, especially tenants of the metric signal of Observatorium, users want to be able to configure Prometheus-based alerts that will be evaluated for a given interval. If triggered they should notify the desired target (receiver) (e.g PagerDuty using Alertmanager Configuration). Users should also be able to see the status of all alerts through Prometheus-like alerts HTTP API. This proposal specifies how such a system can be deployed within Observatorium.
```
bwplotka, line 12: TODO:
* How often do we reload AM conf
* How do we trigger reload?
* Does triggering reload affects AM
* Exact config file? (type)
* How tenancy is enabled?

bwplotka (commented):
Thanks for review, some updates.

docs/proposals/20220105-observatorium-alerting.md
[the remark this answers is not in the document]
jessicalins: just wonder if we have non-goals that can be in the future converted to goals - but in case we already have plans/draft on what can be done in the next iteration, I think is fine:)

pavolloffay:
The image is weirdly rendered on the website

Screenshot of Observatorium Alerting - Observatorium

docs/proposals/20220105-observatorium-alerting.md
matej-g: I have checked this for our internal purposes, and although AppSRE does not offer Vault for uses strictly outside of AppSRE Interface, Red Hat IAM services provide a Vault instance. We can discuss details internally.

docs/proposals/20220105-observatorium-alerting.md
matej-g: I think we should clarify what exactly do we want here from the upstream Alertmanager. The fact is that AM already has an API  which would enable us to get currently loaded configuration, see the `/status` endpoint [in the spec](https://github.com/prometheus/alertmanager/blob/main/api/v2/openapi.yaml#L265).

docs/proposals/20220105-observatorium-alerting.md
matej-g: I believe the biggest question mark we have is still around multi-tenancy and Alertmanager configuration. I'll try to summarize the discussions so far.

#### Main concern:
- We do not have (yet) a good idea on how multi-tenancy with configuration should be implemented. The fact is that Alertmanager does not recognize tenancy, meaning the configuration, from global perspective, represents a configuration of a single user. To make tenancy possible, we need to design a solution which would makes it possible to isolate configurations of multiple users (tenants) while still being represented in the global configuration (on top of this, there are related questions such as how to enable such isolation when it comes to UI and other APIs not included in this proposal - i.e. silences, inhibitions - the design choices here might have impact on these as well).

#### Ideas:
1) Have separate Alertmanager instance for each tenant (hard tenancy)
- With this, we would completely side-step the issue of configuration, since the global config = user config. Downsides? Extra overhead of running Alertmanager instance for each tenant?

2) Combine configurations provided by users into a single global configuration (soft tenancy)
- This aligns with how we deal with e.g. rules (see [proposal](https://github.com/observatorium/observatorium/blob/322c22821b4c4b94a82186f45d7e1f2b4b3b3383/docs/proposals/20201019-prometheus-rules-in-observatorium-api.md?plain=1#L197)). This would mean creating one global file with each tenant being represented in the global configuration. The main issue with this solution is how to combine tenants configuration:
 - Alertmanager has a few global options that are applied if the child routes have no configuration set. If tenants are to provide us with their global parameters, it is not clear how it would be possible to combine global parameters from multiple tenants into one config file.
 - An additional suggestion how to solve this would be to accept only non-global parameters, i.e. `routes` and `receivers` and combine these into a global config file. The downside is obviously that users cannot leverage global options and would need to specify configuration for all routes explicitly. 

In this context it is also worth looking at how Cortex works, in which [Alertmanager ](https://cortexmetrics.io/docs/architecture/#alertmanager)is running as a component within Cortex. This makes it possible to run a 'multitenant Alertmanager' as a single component, which encompasses multiple Alertmanagers, each dedicated to a single tenant (meaning each tenant can provide their isolated, global configuration)

PR #456, opened 2022-02-15
Add proposal for `obsctl`

saswatamcode:
This PR adds a proposal that talks about `obsctl` which is a dedicated CLI to interact with Observatorium instances.

cc: @onprem

### DIFF
--- a/docs/proposals/20220215-obsctl.md
+++ b/docs/proposals/20220215-obsctl.md
@@ -9,13 +9,13 @@
 * **Other docs:**
   * None
 
-## Why
+## TL;DR
 
 We can interact with [Observatorium API](https://github.com/observatorium/api) with a convenient and dedicated CLI, which provides capabilities to persist auth data, and allows us to easily & quickly get up and running with our operations.
 
 ## Pitfalls of Current Solution
 
-Currently, we need to manually run [token-refresher](https://github.com/observatorium/token-refresher) with [OIDC](https://openid.net/connect/) information for a particular tenant, and then make cURL requests through that. This is not optimal and generally confusing/trickier for users, both old and new.
+Currently, we do not have a simple and convenient way to interact with the Observatorium API from the command line. Manually crafting cURL commands while authenticating with OIDC or mTLS is not an optimal experience and generally confusing/trickier for users, both old and new.
 
 ## Goals
 
@@ -28,7 +28,7 @@ obsctl will,
 
 ## Non-Goals
 
-* Adding and dynamically configuring tenants through obsctl (No such admin API is available within Observatorium, nor is it planned)
+* Enabling Observatorium administration via obsctl
 
 ## Audience of this proposal
 
@@ -36,32 +36,36 @@ Observatorium Devs and Users.
 
 ## How
 
-As obsctl is a CLI, authentication options (OIDC) need to be persisted across sessions. obsctl aims to achieve this by saving those configurations locally at some configuration directory and maintaining a "current" tenant.
+As obsctl is a CLI, authentication options (OIDC) need to be persisted across sessions. obsctl aims to achieve this by saving those configurations locally at some configuration directory and maintaining a "current" tenant & a "current" instance of Observatorium, which we can refer to as "context".
 
-The "current" tenant will basically represent the logged-in tenant and all operations will be done based on this "current" tenant configuration. In case, the "current" tenant is empty, i.e, there is no tenant currently logged-in, obsctl will error out.
+The "current context" will basically represent the logged-in tenant and the Observatorium API endpoint which the tenant is allowed to use. All operations will be done based on this "current context" configuration. In case, the "current context" is empty, i.e, there is no tenant currently logged-in or no API provided, obsctl will error out.
 
-obsctl will also allow the user to switch from the "current" tenant to another one. This can be achieved by creating a map of tenant configurations that are uniquely referenceable.
+obsctl will also allow the user to switch from the "current context" to another one. This can be achieved by creating a map of tenant configurations and API configurations that are uniquely referenceable.
+
+For OIDC/OAuth2 based authentication, obsctl will try to fetch refresh token from SSO service (if supported) and store it locally as it has longer validity, and automatically refresh the access token using that when needed.
 
 ### Authentication Commands to be supported by obsctl
 
-* `login <OIDC flags>`: Login as a tenant and save their configuration locally.
-* `logout`: Logout current tenant.
-* `switch <tenant>`: Switch from one tenant to another.
-* `current`: Display the “current” tenant.
+* `login --tenant --api <name/URL> <OIDC flags> `: Login as a tenant and save their configuration locally.
+  * `--name`: Optional flag to override the name used to refer the tenant.
+  * `--api-name`: Optional flag to override the name used to refer the api. By default the hostname of the URL is used.
+* `context api add --url <URL> --name <api name>` : Add a referrable instance of Observatorium.
+* `context switch <api name>/<tenant>`: Switch from one context to another.
+* `context current`: Display the “current” context.
 
 Initially, obsctl aims to support one-off metrics-based operations for users.
 
 ### Metric-based Commands to be supported by obsctl
 
-* `rules get`: Get configured Rules for a tenant (YAML, hits `api/v1/rules/raw`).
-* `rules set <config>`: Set Rules for a tenant by passing in Rule file.
-* `read rules`: Read rules for a tenant. (JSON, hits `api/v1/rules`).
-* `read series`: Read series for a tenant.
-* `read labels`: Read labels for a tenant.
-* `query <PromQL query>`: Query data for a particular tenant (JSON).
+* `metrics get rules.raw`: Get configured Rules for a tenant (YAML, hits `api/v1/rules/raw`).
+* `metrics get rules`: Read rules for a tenant. (JSON, hits `api/v1/rules`).
+* `metrics get series`: Read series for a tenant.
+* `metrics get labels`: Read labels for a tenant.
+* `metrics set --rule.file=<rules-config>`: Set Rules for a tenant by passing in Rule file.
+* `metrics query <PromQL query>`: Query data for a particular tenant (JSON).
 
 Additionally, obsctl will also provide statistical information (such as time taken, response size) about the performed operations along with the result, which might be useful for users willing to debug their instances of Observatorium.
 
 ## Alternatives
 
-* Continuing with the existing process of using token-refresher and cURL and documenting that in detail.
+* Continuing with the existing process of manually crafting cURL requests and documenting that in detail.
--- a/docs/proposals/20220215-obsctl.md
+++ b/docs/proposals/20220215-obsctl.md
@@ -36,24 +36,29 @@ Observatorium Devs and Users.
 
 ## How
 
-As obsctl is a CLI, authentication options (OIDC) need to be persisted across sessions. obsctl aims to achieve this by saving those configurations locally at some configuration directory and maintaining a "current" tenant & a "current" instance of Observatorium, which we can refer to as "context".
+As obsctl is a CLI, authentication options (OIDC) need to be persisted across sessions. obsctl aims to achieve this by saving those configurations locally at a configuration directory and maintaining a "current" tenant & a "current" instance of Observatorium, which we can refer to as "context".
 
 The "current context" will basically represent the logged-in tenant and the Observatorium API endpoint which the tenant is allowed to use. All operations will be done based on this "current context" configuration. In case, the "current context" is empty, i.e, there is no tenant currently logged-in or no API provided, obsctl will error out.
 
-obsctl will also allow the user to switch from the "current context" to another one. This can be achieved by creating a map of tenant configurations and API configurations that are uniquely referenceable.
+obsctl will also allow the user to switch from the "current context" to another one. This can be achieved by creating a map of API configurations, each having tenants, that are uniquely referenceable.
 
-For OIDC/OAuth2 based authentication, obsctl will try to fetch refresh token from SSO service (if supported) and store it locally as it has longer validity, and automatically refresh the access token using that when needed.
+For OIDC/OAuth2 based authentication, obsctl will try to fetch access token from SSO service (if supported) and store it locally, and automatically refresh the access token if the locally saved token expires.
+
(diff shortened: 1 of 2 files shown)


docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+We can interact with [Observatorium API](https://github.com/observatorium/api) with a convenient and dedicated CLI, which provides capabilities to persist auth data, and allows us to easily & quickly get up and running with our operations.
+
+## Pitfalls of Current Solution
+
+Currently, we need to manually run [token-refresher](https://github.com/observatorium/token-refresher) with [OIDC](https://openid.net/connect/) information for a particular tenant, and then make cURL requests through that. This is not optimal and generally confusing/trickier for users, both old and new.
+
+## Goals
+
+obsctl will,
+* Manage authentication configuration for multiple tenants by saving them locally
```
matej-g, line 23: Will `obsctl` also take care of automatically renewing credentials, if e.g. a token is expired? Should we state that in `Goals` / `How`?

matej-g (commented):
Nicely done :raised_hands: Good proposal, I'm adding few suggestion

docs/proposals/20220215-obsctl.md
bwplotka: +1

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+* Allow users to switch between tenants conveniently
+* Configure and view rules for a tenant
+* View series, rules, and labels for a tenant
+* Later on, support more such one-off operations for other domains as well (logs, traces, alert-routing configs)
```
bwplotka, line 27: What about admin operations like generating output from our templates or even generating operator as we experimented with https://github.com/observatorium/rndr?

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Audience of this proposal
+
+Observatorium Devs and Users.
+
+## How
+
+As obsctl is a CLI, authentication options (OIDC) need to be persisted across sessions. obsctl aims to achieve this by saving those configurations locally at some configuration directory and maintaining a "current" tenant.
+
+The "current" tenant will basically represent the logged-in tenant and all operations will be done based on this "current" tenant configuration. In case, the "current" tenant is empty, i.e, there is no tenant currently logged-in, obsctl will error out.
```
bwplotka, line 41: How current tenant will work for admin sessions?

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+obsctl will,
+* Manage authentication configuration for multiple tenants by saving them locally
+* Allow users to switch between tenants conveniently
+* Configure and view rules for a tenant
+* View series, rules, and labels for a tenant
+* Later on, support more such one-off operations for other domains as well (logs, traces, alert-routing configs)
+
+## Non-Goals
+
+* Adding and dynamically configuring tenants through obsctl (No such admin API is available within Observatorium, nor is it planned)
```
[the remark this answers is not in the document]
saswatamcode: Yup, we can. I was wondering if this would be a future feature though. Maybe leaving it as `nor is it planned in the near future` would be better here?

docs/proposals/20220215-obsctl.md
saswatamcode: Yes, will clarify this better! 🙂

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+
+The "current" tenant will basically represent the logged-in tenant and all operations will be done based on this "current" tenant configuration. In case, the "current" tenant is empty, i.e, there is no tenant currently logged-in, obsctl will error out.
+
+obsctl will also allow the user to switch from the "current" tenant to another one. This can be achieved by creating a map of tenant configurations that are uniquely referenceable.
+
+### Authentication Commands to be supported by obsctl
+
+* `login <OIDC flags>`: Login as a tenant and save their configuration locally.
+* `logout`: Logout current tenant.
+* `switch <tenant>`: Switch from one tenant to another.
```
bwplotka, line 49: So the flow will be:

obsctl switch tenant1
obsctl login 

In order to login as tenant? Quite opaque, maybe there is way to simplify this. Why do I need login command even? What any command that needs Auth do automatic renewal if needed etc?

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+* `current`: Display the “current” tenant.
+
+Initially, obsctl aims to support one-off metrics-based operations for users.
+
+### Metric-based Commands to be supported by obsctl
+
+* `rules get`: Get configured Rules for a tenant (YAML, hits `api/v1/rules/raw`).
+* `rules set <config>`: Set Rules for a tenant by passing in Rule file.
+* `read rules`: Read rules for a tenant. (JSON, hits `api/v1/rules`).
```
bwplotka, line 58: Let's be consistent. We had <rules> <verb> with rules set and not verb is in beginning?

Check some tips from https://archive.fosdem.org/2019/schedule/event/designingcli/

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+### Authentication Commands to be supported by obsctl
+
+* `login <OIDC flags>`: Login as a tenant and save their configuration locally.
+* `logout`: Logout current tenant.
+* `switch <tenant>`: Switch from one tenant to another.
+* `current`: Display the “current” tenant.
+
+Initially, obsctl aims to support one-off metrics-based operations for users.
+
+### Metric-based Commands to be supported by obsctl
```
bwplotka, line 54: Soo all of this will clash with logs and traces... Should we make this `obsctl metric rules ...`?

bwplotka (changes requested):
Nice, amazing work! Some comments (:

docs/proposals/20220215-obsctl.md
saswatamcode: The flow would be something like, 

We need to perform operations for `tenant1` so we run `obsctl login <tenant1 auth flags>` which doesn't do any HTTP operation but saves the tenant config to a local config file and sets the current tenant to `tenant1`. Any subsequent `obsctl` operations (which will have HTTP calls) would use config for `tenant1`.

Next, we might want to switch to `tenant2` for other operations, so would need to login as `tenant2` by running `obsctl login <tenant2 auth flags>`. Any subsequent operations would then use `tenant2` config as it was set to the current tenant.

But then once we want to switch back to `tenant1` we would run `obsctl switch tenant1`. Any subsequent operation after this would again use `tenant1` configs as current tenant is again set to `tenant1`. 

obsctl would aim to follow similar semantics to kubectl [context](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#define-clusters-users-and-contexts). 🙂

As for token sessions, that is something that needs to be better clarified here. Will discuss this with @onprem!

docs/proposals/20220215-obsctl.md
saswatamcode: Yes, that will be better. Let me change!

docs/proposals/20220215-obsctl.md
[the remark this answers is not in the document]
moadz: In general setting it as a non-goal here doesn't stop us from revisiting when we have figured out how admin permissions would work in multi-tenant deployments.

docs/proposals/20220215-obsctl.md
moadz: My feeling is tenants/observatorium are more analogous to projects/namespaces in a kube cluster as opposed to completely distinct clusters. 

My understanding is switching tenants will always have to be preceded with a login command, so why not merge the two and make it so that a `switch` does a config OIDC config check/token handshake for that particular tenant? It would render the login/logout commands moot and simplify the CLI.

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+
+* `rules get`: Get configured Rules for a tenant (YAML, hits `api/v1/rules/raw`).
+* `rules set <config>`: Set Rules for a tenant by passing in Rule file.
+* `read rules`: Read rules for a tenant. (JSON, hits `api/v1/rules`).
+* `read series`: Read series for a tenant.
```
moadz, line 59: Is this to return all unique timeseries associated with a tenant?

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+* `read labels`: Read labels for a tenant.
+* `query <PromQL query>`: Query data for a particular tenant (JSON).
+
+Additionally, obsctl will also provide statistical information (such as time taken, response size) about the performed operations along with the result, which might be useful for users willing to debug their instances of Observatorium.
+
+## Alternatives
+
+* Continuing with the existing process of using token-refresher and cURL and documenting that in detail.
```
moadz, line 67: Perhaps @bwplotka @matej-g have context on this, but are/where there plans to implement implicit auth flow with obs-api? Instead of needing a refresh token/token-refresher flow all-together

moadz (commented):
LGTM, good first pass!

docs/proposals/20220215-obsctl.md
onprem: When we introduce admin level APIs in Observatroium, we still would need to have authentication on top of it, maybe we can use the same context to store admin credentials as well. The whole picture around how admin endpoints will work is not clear as of now so I think we can think about this in future.

docs/proposals/20220215-obsctl.md
onprem: I see that tenants are similar to namespaces, but a major difference is that each tenant has different creds, while in K8s your credentials are cluster level. I spent some time thinking about how this whole login flow should look like, and I think it will be even trickier that K8s. Because we have multiple instances of Observatorium API, with multiple tenants, and these tenants can have multiple users.


> What any command that needs Auth do automatic renewal if needed etc?

Since we are using OAuth2/OIDC, we can get refresh tokens from SSO provider and implement automatic token refresh flow. I think we should do that.

docs/proposals/20220215-obsctl.md
saswatamcode: Yes, essentially same response as `/api/v1/metrics/{tenant}/api/v1/series`.

docs/proposals/20220215-obsctl.md
saswatamcode: We can certainly add such functionality in the future with obsctl, but for now, this might be out of scope.

saswatamcode:
Updated proposal with some ideas on how auth will be managed better, after 1:1 with @onprem today!

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,77 @@
...
+* **Other docs:**
+  * None
+
+## TL;DR
+
+We can interact with [Observatorium API](https://github.com/observatorium/api) with a convenient and dedicated CLI, which provides capabilities to persist auth data, and allows us to easily & quickly get up and running with our operations.
+
+## Pitfalls of Current Solution
+
+Currently, we do not have a simple and convenient way to interact with the Observatorium API from the command line. Manually crafting cURL commands while authenticating with OIDC or mTLS is not an optimal experience and generally confusing/trickier for users, both old and new.
```
jessicalins, line 18:
```suggestion
Currently, we do not have a simple and convenient way to interact with the Observatorium API from the command line. Manually crafting cURL commands while authenticating with OIDC or mTLS is not an optimal experience and is generally confusing/trickier for users, both old and new.
```

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,77 @@
...
+obsctl will,
+* Manage authentication configuration for multiple tenants by saving them locally
+* Allow users to switch between tenants conveniently
+* Configure and view rules for a tenant
+* View series, rules, and labels for a tenant
+* Later on, support more such one-off operations for other domains as well (logs, traces, alert-routing configs)
+
+## Non-Goals
+
+* Enabling Observatorium administration via obsctl
```
jessicalins, line 31: what does it mean 'Observatorium administration'? Perhaps we could clarify this better?

jessicalins: does it refer to https://github.com/observatorium/observatorium/pull/456/files#r807905979 for example?

jessicalins (commented):
looking good, just a small suggestion from my side :)

docs/proposals/20220215-obsctl.md
saswatamcode: Sure, let me clarify! 🙂

saswatamcode: Clarified this here: https://github.com/observatorium/observatorium/pull/456/commits/16ccd1d1c2fd096def38d20b2322960042c08b02.

docs/proposals/20220215-obsctl.md
```diff
@@ -0,0 +1,67 @@
...
+
+* **Related Tickets:**
+  * https://issues.redhat.com/browse/MON-2199
+
+* **Other docs:**
+  * None
+
+## Why
+
+We can interact with [Observatorium API](https://github.com/observatorium/api) with a convenient and dedicated CLI, which provides capabilities to persist auth data, and allows us to easily & quickly get up and running with our operations.
```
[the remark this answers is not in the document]
matej-g: Minor thingy, I think it's still unresolved :slightly_smiling_face:

matej-g:
@saswatamcode what's the status? It seems like most of the things have been addressed and this is ready for final reviews, correct?

saswatamcode:
Yup, it's good to go @matej-g! 🙂

docs/proposals/20220215-obsctl.md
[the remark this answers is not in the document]
saswatamcode: I've changed the header to TL;DR in https://github.com/observatorium/observatorium/pull/456/commits/d9c5e150be2bbd69210687766071ad623d76232f 🙂

saswatamcode:
I think I've addressed the minor comments, so this should be good to go! 🙂

jessicalins (approved):
LGTM :)

PR #463, opened 2022-03-25
loki: Add query scheduler and index gateway components

periklis:
The following PR adds two new Loki components namely the `query-scheduler` and the `index-gateway`. The rational having this two components is:
1. The `index-gateway` incorporates loading and caching the `boltdb` index files on a persistent volume that used to live in the `querier` components. In turn queriers can be transformed from `StatefulSet` to a `Deployment` enabling horizontal scaling w/o requiring any persistent volumes anymore.
2. The `query-scheduler` incorporates the scheduling logic of multi-tenant and time-range shardable queries that used to live in the `query-frontend` and `querier` components. In turn both `query-frontend` and `querier` can be scaled horizontally without tuning the concurrency settings (e.g. `max_concurrent`). These settings will be tuned by the querier automatically based on the amount of available schedulers rather query-frontends.

In summary the proposed changes simplify horizontal scaling of the query path by making externalizing operations to dedicated components that can operate with a static replica count over time. This enables a more elastic scaling behavior of the main work horses in the query path, i.e.:
- The `query_frontend` can be scale out based on the incoming http traffic requests w/o a need to inform queries of their current replica count and tune latter's internal concurrency settings.
- The `queriers` can be scaled out as simple deployments and thus provide more query processing power w/o requiring persistent volumes.

(cc @xperimental)

### DIFF
--- a/configuration/components/loki.libsonnet
+++ b/configuration/components/loki.libsonnet
@@ -299,9 +299,11 @@ local defaults = {
         },
       } else {
         results_cache: {
-          enable_fifocache: true,
-          fifocache: {
-            max_size_bytes: '500MB',
+          cache: {
+            enable_fifocache: true,
+            fifocache: {
+              max_size_bytes: '500MB',
+            },
           },
         },
       }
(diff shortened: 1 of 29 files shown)


bwplotka (approved):
A number of different services we need to run are getting scary... But we don't have a choice, all of them are needed (: Thanks for explaining all.


Wonder if we can avoid those in Thanos. Index GW we need atm, since we have bigger blocks. We don't have persistent queue because of federated query mechanism (but we will need some queues soon). (:

PR #464, opened 2022-03-28
loki: Add service monitors for query_scheduler and index_gateway

periklis:
This PR adds the missing service monitor support for the loki query-scheduler and index-gateway components. Missed to add in #463


squat (approved):
Nice

PR #465, opened 2022-03-29
Fix optional manifests

saswatamcode:
This PR fixes an issue where importing latest `observatorium.libsonnet` causes errors like 
```
RUNTIME ERROR: Unexpected type null, expected object
	vendor/github.com/observatorium/observatorium/configuration/components/observatorium.libsonnet:(145:5)-(162:14)	object <anonymous>
	services/observatorium-template.jsonnet:11:36-49	thunk from <object <anonymous>>
	<std>:1293:24-25	thunk from <function <anonymous>>
	<std>:1293:5-33	function <anonymous>
	services/observatorium-template.jsonnet:11:19-50	object <anonymous>
	services/observatorium-template.jsonnet:(7:5)-(15:6)	
	During manifestation	
```
This is due to if-else blocks not being evaluated correctly, in `manifests` field.

### DIFF
--- a/configuration/components/observatorium.libsonnet
+++ b/configuration/components/observatorium.libsonnet
@@ -157,11 +157,12 @@ local api = (import 'observatorium-api/observatorium-api.libsonnet');
     } + {
       ['thanos-' + name]: obs.thanos.manifests[name]
       for name in std.objectFields(obs.thanos.manifests)
-    } + if std.objectHas(obs.loki, 'manifests') then {
-      ['loki-' + name]: obs.loki.manifests[name]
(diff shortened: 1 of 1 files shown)


squat:
Hmm it would be really great if we added some tests to the CI of this repo to babe sure that we don't break things more in the future. It's kind of amazing that we don't have any yet :)

saswatamcode:
Yup! I think that would be really useful to have. We already have some e2e tests, but nothing to test this kind of case. Maybe we can use something like: https://github.com/yugui/jsonnetunit?

matej-g (approved):

jessicalins (approved):
:tada:

Issue #470, opened 2022-04-07
`make generate` fails on macOS


esnible:
`make generate` fails on OSX, producing a file named `{}.yaml` and deleting most of the .yaml deployment examples.

To reproduce, on a Mac,

```bash
cd configuration
touch components/observatorium.libsonnet
make generate
git status
```

The problem is very small `xargs` defaults in Darwin.  I am preparing a PR.

Note: This was the most confusing bug I have seen in a long time.  It took me a long time to realize that `make generate` was only deleting 1/3 of the example deployment .yaml, rather than all of it.  I spent a long time looking for problems with shell, filename, and escape sequence differences between Mac and Linux.

PR #471, opened 2022-04-07
macOS compatibility

esnible:
Resolves https://github.com/observatorium/observatorium/issues/470

macOS includes a BSD-derived `xargs`, with very different flags from the GNU `xargs` on Linux systems.
See: https://www.freebsd.org/cgi/man.cgi?xargs

### DIFF
--- a/configuration/Makefile
+++ b/configuration/Makefile
@@ -11,6 +11,12 @@ RNDR ?= ../../rndr/.bin/rndr # DEBUG
 
 JSONNET_SRC = $(shell find . -type f -not -path './*vendor/*' \( -name '*.libsonnet' -o -name '*.jsonnet' \))
 
+UNAME := $(shell uname)
+ifeq ($(UNAME), Darwin)
+   # Overcome Mac default limiting xargs -I replace size.
+   XARGS_FLAGS ?= -S 4096
+endif
+
 all: generate validate
 
 render:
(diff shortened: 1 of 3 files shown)


configuration/Makefile
```diff
@@ -11,6 +11,12 @@ RNDR ?= ../../rndr/.bin/rndr # DEBUG
 
 JSONNET_SRC = $(shell find . -type f -not -path './*vendor/*' \( -name '*.libsonnet' -o -name '*.jsonnet' \))
 
+UNAME := $(shell uname)
+ifeq ($(UNAME), Darwin)
+   # Overcome Mac default limiting xargs -I replace size.
+   XARGS_FLAGS ?= -S 4096
```
saswatamcode, line 17: Alternatively, we could just look for GNU version of `xargs` on macOS and use that using something like `XARGS ?= $(shell which gxargs 2>/dev/null || which xargs)` (you can install `gxargs` with `brew install findutils`).

But this works as well, so I don't feel too strongly either way! 🙂

saswatamcode (dismissed):
Thanks for fixing it! This must've been a tricky bug to figure out! 💪

configuration/Makefile
moadz: I agree with saswata, it's probably better to prompt users to use gnu `findutils`.

esnible: @moadz The Makefile's behavior on unmodified Mac/BSD is bizarre and it will be hard to communicate the solution (`brew install findutils`) using @saswatamcode 's idea to have `gxargs` override `xargs` -- but only if present.

This is similar but will fail in an obvious way:

```bash
ifeq ($(UNAME), Darwin)
   # macOS/BSD xargs not compatible with this Makefile (because of -S replsize default of 256)
   XARGS ?= gxargs
else
   XARGS ?= xargs
endif
```

Alternately we could use the work-around in this PR which does not require installing anything.

moadz: If one is installing `findutils` they're most likely adding it to their `PATH` as well. 

I like that your approach requires little intervention on behalf of the user. But, I also think in this case where GNU/BSD `xargs` (and other tools like `sed`)  have diverged significantly, it is more useful that a contributor is aware of explicit GNU compatibility requirement on BSD. This means any new changes + steps can be done without having to consider BSD backwards compatibility.

esnible: We have talked about three approaches:
- My first approach, which detects Macs and adds BSD-specific flags
- My second approach https://github.com/observatorium/observatorium/pull/471#discussion_r845008628 which requires Mac users to install GNU _xargs_ as _gxargs_
- @saswatamcode 's approach which uses gxargs if present.

I believe my second approach best addresses @moadz 's comments.  I will revise the PR toward that approach.

saswatamcode (approved):

matej-g (approved):
I am not able to test on MacOS, but the changes are looking good and I trust @saswatamcode tested this before :+1:

PR #474, opened 2022-04-12
add jaeger ui default settings

frzifus:
In Jaeger, ui modifications can be made via a [ui config](https://www.jaegertracing.io/docs/1.32/frontend-ui/#configuration). The operator supports a configuration directly via [CR entry](https://www.jaegertracing.io/docs/1.32/operator/#configuring-the-ui). This pr extends the tracing lib by this field.

Tracing is enabled in the local example, there i added an example configuration.

Happy to hear your thoughts on this.

---

cc @esnible @pavolloffay

### DIFF
--- a/configuration/components/tracing.libsonnet
+++ b/configuration/components/tracing.libsonnet
@@ -93,6 +93,29 @@ function(params) {
       },
       spec: {
         strategy: 'allinone',
+        ui: {
+          options: {
+            dependencies: {
+              menuEnabled: false,
+            },
+            menu: [
+              {
+                items: [
+                  {
+                    label: 'Documentation',
+                    url: 'https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html/distributed_tracing/index',
+                  },
+                ],
+                label: 'About',
+              },
+              {
+                anchorTarget: '_self',
+                label: 'Log Out',
+                url: './login',
+              },
+            ],
+          },
+        },
       },
     },
   manifests: {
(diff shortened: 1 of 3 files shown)


configuration/components/tracing.libsonnet
```diff
@@ -92,7 +92,8 @@ function(params) {
         labels: newCommonLabels(component),
       },
       spec: {
-        strategy: 'allinone',
+        strategy: tracing.config.jaeger.strategy,
```
frzifus, line 95: wait that does not make sense, because its the default configuration?

pavolloffay: We should establish some convention here. This PR essentially forces us to change `tracing.libsonnet` every time we need to make a new field configurable in the Jaeger CR. A more flexible solution might be better option. For instance a consumer of the `tracing.libsonnet` could define full Jaeger CR.

frzifus: Meaning that we store an `JaegerSpec` with default values somewhere like `components/jaeger.libsonnet`, which could be adapted in main and passed via the tracing configuration?

frzifus:
@esnible , @pavolloffay could you have a look on the changes again?

configuration/components/tracing.libsonnet
```diff
@@ -91,8 +91,23 @@ function(params) {
...
+      spec: tracing.config.jaegerSpec,
+    } + {
+      spec+: {
+        ui+: {
+          options+: {
+            menu+: [
+              {
+                anchorTarget: '_self',
+                label: 'Log Out',
+                url: './login',
```
esnible, line 103: This URL won't work unless https://github.com/observatorium/api/pull/274 is merged.  We might want to hold off on this one until we get a positive review on that one.

Issue #472, opened 2022-04-08
API configured for traces even with tracing disabled


esnible:
Notice that https://github.com/observatorium/observatorium/blob/main/configuration/examples/dev/manifests/api-deployment.yaml#L43 has

```yaml
        - --traces.write.endpoint=observatorium-xyz-otel-collector:4317
        - --grpc.listen=0.0.0.0:8090
```

Of the three examples, only _local_ has tracing enabled, yet all three examples supply now the flags to the API to forward traces.  (Since https://github.com/observatorium/observatorium/pull/461 ).

I am looking into the fix but Jsonnet debugging is new to me.  I have the syntax figured out, but Jsonnet's lazy declarative semantics make it tricky to figure out why part of the Jsonnet respects tracing `.enabled` and another part doesn't.

cc @pavolloffay

squat:
I wonder if `.enabled` is kind of an anti-pattern. We don't use this field to control the enabling of logging or metrics, only the existence of the configuration is enough. Perhaps the solution is rather to eliminate `enabled` entirely?

esnible:
@squat You have a good point.  Perhaps this is why the previous version enabled it.
It troubled me that we use `enabled` to turn off deployment of the back-end, but configure the API to forward traces.  If the user has chosen to deploy without traces we shouldn't offer the APIs.

bwplotka:
Right. I wonder if this makes sense to do as @squat mention - treat no endpoint config as `enabled=false`

PR #475, opened 2022-04-12
Don't deploy API to forward unless tracing enabled

esnible:
This PR implements half of a fix for https://github.com/observatorium/observatorium/issues/472

The other half is https://github.com/observatorium/api/pull/280 in the observatorium/api repo.
This PR will not change the behavior of the generated manifests until the other PR is merged.

I am creating this PR as a draft, so that it doesn't accidentally get merged before [271](https://github.com/observatorium/api/pull/280).

cc @pavolloffay @jessicalins

### DIFF
--- a/configuration/components/observatorium.libsonnet
+++ b/configuration/components/observatorium.libsonnet
@@ -79,7 +79,13 @@ local api = (import 'observatorium-api/observatorium-api.libsonnet');
       ],
     },
     traces: {
+      // gRPC OpenTelemetry collector for traces endpoint
       writeEndpoint: obs.tracing.manifests.otelcollector.metadata.name + '-collector:4317',
+
+      // Jaeger V2 HTTP query for traces using Observatorium tenancy pattern
+      templateEndpoint: 'http://jaeger-{tenant}-collector:16686',
+
+      enabled: false,
     },
     rateLimiter: {
       grpcAddress: '%s.%s.svc.cluster.local:%d' % [
--- a/configuration/examples/base/manifests/api-deployment.yaml
+++ b/configuration/examples/base/manifests/api-deployment.yaml
@@ -40,8 +40,6 @@ spec:
         - --logs.read.endpoint=http://observatorium-xyz-loki-query-frontend-http.observatorium.svc.cluster.local:3100
         - --logs.tail.endpoint=http://observatorium-xyz-loki-querier-http.observatorium.svc.cluster.local:3100
         - --logs.write.endpoint=http://observatorium-xyz-loki-distributor-http.observatorium.svc.cluster.local:3100
-        - --traces.write.endpoint=observatorium-xyz-otel-collector:4317
-        - --grpc.listen=0.0.0.0:8090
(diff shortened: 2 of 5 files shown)


configuration/components/observatorium.libsonnet
```diff
@@ -79,7 +79,13 @@ local api = (import 'observatorium-api/observatorium-api.libsonnet');
       ],
     },
     traces: {
+      // gRPC OpenTelemetry collector for traces endpoint
       writeEndpoint: obs.tracing.manifests.otelcollector.metadata.name + '-collector:4317',
+
+      // Jaeger V2 HTTP query for traces using Observatorium tenancy pattern
+      templateEndpoint: 'http://jaeger-{tenant}-collector:16686',
+
+      enabled: false,
```
pavolloffay, line 88: `tracing.libsonnet` has enabled flag as well https://github.com/observatorium/observatorium/blob/main/configuration/components/tracing.libsonnet#L10 which controls tracing manifests (jaeger, OTELcol CRs) generation - https://github.com/observatorium/observatorium/blob/main/configuration/components/observatorium.libsonnet#L164 
Could we unify those two enabled fields? 


I don't see where the `enabled` field from this PR is used e.g. in a condition.

esnible: This enabled field is being used in the draft PR https://github.com/observatorium/api/pull/280/files#diff-d79e05fae1c17a80c15dfd8e05409c355f8eb20c179667763c26ef6351c18fbfR133

See also this comment: https://github.com/observatorium/observatorium/issues/472#issuecomment-1097011831

I'm not a Jsonnet expert but I spent several days trying to access the other `.enabled` or to use an empty `writeEndpoint` field.  Those fields don't seem to be visible from _jsonnet/lib/observatorium-api.libsonnet_.

pavolloffay: I think we could define the `enabled` field only in `observatorium-api.libsonnet` as the condition where it is used is in that file.

esnible: @pavolloffay I will defer to you and others for the best pattern to enable/disable in Jsonnet.

I am a Jsonnet beginner.  I saw in _jsonnet/lib/observatorium-api.libsonnet_ the code 

```Jsonnet
                if api.config.traces != {} then
                  [
                    '--traces.write.endpoint=' + api.config.traces.writeEndpoint,
                    '--grpc.listen=0.0.0.0:' + api.config.ports['grpc-public'],
```

... and I assumed that these ports would be closed unless tracing was defined.  At that time I hadn't looked at observatorium/observatorium.

When I looked at the examples in observatorium/observatorium, the _base_, _dev_, and _local_ in _configuration/examples_, I thought that only _local_ had tracing and thus only _local_ should have a _manifests/api-deployment.yaml_ with `--traces.write.endpoint`.  It was confusing to see the API server enabling traces when the collector wasn't deployed.

This PR is my attempt to pass the flag into _observatorium-api.libsonnet_ so that a .jsonnet file can enable or disable tracing.

esnible:
Closing; will work on different changes for same effect.

Issue #476, opened 2022-04-14
probably invalid twitter link


frzifus:
Probably the twitter link in the [governance](https://github.com/observatorium/observatorium/blob/3d467717ca3008a74ce645c2e7c4dd2e5e01f5a3/docs/community/governance.md#maintainers-team) file is not quite up to date?

> New maintainer is announced on the [Observatorium Twitter](https://twitter.com/0bservatorium) by an existing team member.

It points to https://twitter.com/0bservatorium.

![image](https://user-images.githubusercontent.com/10403402/163388452-3f8e5289-b976-445d-b265-0af145eff37d.png)

jessicalins:
fixed by https://github.com/observatorium/observatorium/pull/477

PR #477, opened 2022-04-18
Repoint Twitter link

esnible:
The current Twitter link points to a non-Observatorium-related Twitter user.

### DIFF
--- a/.mdox.validate.yaml
+++ b/.mdox.validate.yaml
@@ -7,3 +7,6 @@ validators:
   # Always returns 403.
   - regex: 'cloudflare\.com'
     type: 'ignore'
+  # May return a 404 in some environments
+  - regex: 'twitter\.com'
+    type: 'ignore'
--- a/.mdox.validate.yaml
+++ b/.mdox.validate.yaml
@@ -7,6 +7,6 @@ validators:
   # Always returns 403.
   - regex: 'cloudflare\.com'
     type: 'ignore'
-  # May return a 404 in some environments
+  # May return a 404 in some environments.
   - regex: 'twitter\.com'
     type: 'ignore'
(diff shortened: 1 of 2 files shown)


squat (dismissed):

squat:
🥲maybe we need a more unique/recognizable Twitter handle

squat:
Hmm for some reason the docs check is failing even though both links work.

Cc @saswatamcode do you think this is Twitter rejecting mdox's automation or something?

saswatamcode:
Yes, I think, but only when it's through GitHub Actions. I tried it locally and `make docs/check-docs` works. But not on GitHub Actions, I even tried it on a separate branch: https://github.com/saswatamcode/observatorium/runs/6064026598?check_suite_focus=true. I wonder why? 🤔

squat:
Twitter blocking GitHub actions IPs to protect against crawling + scraping?

saswatamcode:
Yup, I thought so, but then it would also break on any repo with mdox running in Actions, but it doesn't, for example: https://github.com/saswatamcode/thanos/runs/6064578300?check_suite_focus=true.

esnible:
In the past, Twitter has cut off access to URLs via GET.  They want to work via the browser.  See https://twitter.com/zarfeblong/status/1339742840142872577

That can't be the case here; I can run `make check-docs` and it works fine for me.  Perhaps Twitter rate-limits Github?  Perhaps Observatorium should tell _mdox_ or _colly_ to ignore or transform Twitter URLs?

matej-g:
I actually wanted to completely remove mention of Twitter from after seeing https://github.com/observatorium/observatorium/issues/476, since the Observatorium Twitter account does not seem to be actively used (at least at the moment) and that part of the procedure - announcing new maintainers - I have never seen in action, so it seemed a bit redundant. WDYT?

saswatamcode:
We can just ignore the twitter link by adding it to [`.mdox.validate.yaml`](https://github.com/observatorium/observatorium/blob/main/.mdox.validate.yaml), but I'd be interested to know why this happens and why only in this repo. It started when this commit landed in main: https://github.com/observatorium/observatorium/commit/c09470dc853125e5a3699d9c681165b4c77a5bf1

jessicalins:
https://github.com/observatorium/observatorium/pull/477#issuecomment-1102209175
+1

squat (approved):

PR #502, opened 2022-10-04
Remove volumes for loki ruler when using s3 storage

periklis:
This is a follow-up on #500 to remove static rules from a ConfigMap when the Loki Ruler config is using object storage as the default storage config.


douglascamata (approved):
LGTM

jessicalins (approved):
LGTM

PR #504, opened 2023-01-09
jsonnet: Update gubernator rate limiting component to latest

PhilipGough:


squat (approved):
Lgtm, I think netlify had a hiccough

PhilipGough:
@squat yeah it seems to be having issues but some circle CI failures too so will investigate!

saswatamcode:
Seems like the Netlify build images have been deprecated, and need to updated in the Observatorium team account.

squat:
> Seems like the Netlify build images have been deprecated, and need to updated in the Observatorium team account.

who has access to the Netlify team account so we can update the build image?

matej-g:
> who has access to the Netlify team account so we can update the build image?

Hm I'm afraid it might be only @bwplotka who's on a break 😞

saswatamcode (approved):
Thanks!

PR #505, opened 2023-01-09
Update expired certificates for e2e manifests/tests

PhilipGough:
While working on #504, I noticed the e2e tests begin to fail due to expired SSL certificates.

This regenerates the certs and updates the tests manifests accordingly.


configuration/tests/main.jsonnet
```diff
@@ -80,12 +80,12 @@ local metricsConfig = {
   writeEndpoint: 'http://%s.%s.svc.cluster.local:%d/api/metrics/v1/test/api/v1/receive' % [
     obs.api.service.metadata.name,
     obs.api.service.metadata.namespace,
-    obs.api.service.spec.ports[1].port,
+    obs.api.service.spec.ports[2].port,
```
saswatamcode, line 83: Seems like this was the wrong port previously?

saswatamcode (approved):
Thanks! LGTM!
Netlify fails due to https://github.com/observatorium/observatorium/pull/504#issuecomment-1376649855

One small question.

configuration/tests/main.jsonnet
PhilipGough: No I don't think so. But something has changed since these manifests were last generated causing the current port (`1`) to point at the internal  port on 8081. This change is required to avoid the diff in the generated manifests.

matej-g (dismissed):
⭐

PR #506, opened 2023-01-27
Upgrade Loki to 2.7.2

aminesnow:
This PR upgrades Loki from 2.6.1 to 2.7.2

### DIFF
--- a/configuration/components/loki.libsonnet
+++ b/configuration/components/loki.libsonnet
@@ -172,9 +172,9 @@ local defaults = {
         },
       } else {
         chunk_cache_config: {
-          enable_fifocache: true,
-          fifocache: {
-            max_size_bytes: '500MB',
+          embedded_cache: {
+            enabled: true,
+            max_size_mb: 500,
           },
         },
       }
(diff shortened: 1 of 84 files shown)


configuration/components/loki.libsonnet
```diff
@@ -294,6 +301,7 @@ local defaults = {
       per_stream_rate_limit: '3MB',
       per_stream_rate_limit_burst: '15MB',
       split_queries_by_interval: '30m',
+      deletion_mode: 'disabled',
     },
     querier: {
```
periklis, line 306: According to upgrade docs (https://grafana.com/docs/loki/latest/upgrading/#engine-query-timeout-is-deprecated) we should drop `query.engine.timeout` in favor of `query.query_timeout`. This is probably a hindsight from the last update to 2.7.0

periklis (changes requested):

configuration/components/loki.libsonnet
aminesnow: I saw that. Thing is, I don't know which value to keep. WDYT the value should be?

periklis: we should keep the `query.query_timeout` with the value as is.

periklis (commented):
Looks good to me

periklis (approved):
Approving this as per upcoming v2.7.3 is not going to provide changes for the loki component jsonnet here.

PR #507, opened 2023-01-30
Add proposal + PoC for Observatorium Go-based configuration

saswatamcode:
This PR contains two things,

- Proposal for Go-based configuration for Observatorium using [mimic](https://github.com/bwplotka/mimic): https://github.com/saswatamcode/observatorium/blob/config-go/docs/proposals/20230130-go-configuration.md
- PoC of Go-based configuration lib for Observatorium API and its usage: https://github.com/saswatamcode/observatorium/tree/config-go/configuration_go

cc: @moadz @bwplotka


docs/proposals/20230130-go-configuration.md
```diff
@@ -0,0 +1,189 @@
...
+* Have a consistent installation and usage model for Observatorium, as a customizable simple static configuration generation solution, with sane defaults
+
+## Non-Goals
+
+* Deprecating/removing jsonnet entirely. This is only proposing Go-based configuration as an alternative method of usage, but does not take away from exisiting jsonnet templating in any way
+* Operator or operator generation. This proposal only aims for static configuration generation, and is not related to applying this configuration or automating it
+
+## Audience of this proposal
+
+Observatorium Devs and Users.
```
douglascamata, line 42: Users are part of the audience but the proposal doesn't outline what's the user's point of view. For example, if I deploy Observatorium today using the jsonnet libs & apply some customizations there I only need to: create a jsonnetfile, add the dependency, run the jsonnet bundler tools (`jb`), and then use the `jsonnet` cli to render them. 

What would be the steps to reuse configuration with this Go system?

docs/proposals/20230130-go-configuration.md
```diff
@@ -0,0 +1,189 @@
...
+			api.WithAPIResources(apiResources),
+			api.WithServiceMonitor(),
+		).Manifests(),
+		"config",
+	)
+```
+
+### Future Installation Model
+
+We can even leverage [`obsctl`](https://github.com/observatorium/obsctl) in the future to allow generating boilerplate Go code that imports Observatorium Go config and provides the user with commands to generate sane default YAML config.
```
douglascamata, line 163: Isn't `obsctl` becoming a tool that does many very different things? Is this okay for the project's direction?

docs/proposals/20230130-go-configuration.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Pitfalls of Current Solution
+
+Currently, we primarily use [jsonnet](https://jsonnet.org/) in Observatorium, to compose a "meta-project" library that allows generating configuration for a multi-tenant scalable observability system.
+
+While jsonnet is a commonly used configuration templating language, there are certain issues with how jsonnet acts as a front-level configuration interface, which causes it to not be preferable to end-users.
+
+Templating via jsonnet, while achievable, often results in complex, unreadable code, and due to no doc generation and low IDE support becomes virtually undiscoverable. Codifying and abstracting operational practices and knowledge in jsonnet is difficult and time-consuming, due to the lack of type safety and testing practices. In certain cases, pre-existing upstreams also cannot be used and jsonnet templating needs to be written manually.
+
+Furthermore, for a project like Observatorium, this leads to a complex-to-adopt and messy installation model. An end-user is forced to go through one of the below,
```
douglascamata, line 21: Or what if the end-user already uses jsonnet in their environment due to wider community adoption (i.e. from Grafana, Prometheus, etc)?

docs/proposals/20230130-go-configuration.md
saswatamcode: I don't think the steps for jsonnet are quite clear, here. There are a few more steps to using jsonnet: create a jsonnetfile, add the dependency, run the jsonnet bundler tools (jb), understand how Observatorium's jsonnet lib works and is constructed (without docs/IDE support) and hunt through generated YAMLs to figure out how end state is affected, customize by writing complex jsonnet for diff envs to get to state that fits user deployment pipeline, use the jsonnet cli to render them to JSON and then use a tool like gojsontoyaml to convert to YAML.

The path for Go-based one is: 
Have Go tools installed, create a Go module and import obs/obs, check an example to figure out observatorium config abstractions or just go to pkg.go.dev for godoc or just click through + autocomplete in IDE to understand abstractions, use said abstractions and customize struct for diff envs to fit user deploy pipeline, `go run` to get YAML you need

In both cases above, we don't consider steps about learning Go or jsonnet.

docs/proposals/20230130-go-configuration.md
saswatamcode: `obsctl` is a tool for making observatorium easier to interact with IMO. This just generates a boilerplate, and doesn't do anything special, customization is still a manual operation for user. However I don't feel too strongly about this, can remove this idea if others agree. 🙂

docs/proposals/20230130-go-configuration.md
saswatamcode: That's fine! 

It still means the end-user would have to learn how the Observatorium jsonnet lib is constructed and then spend time navigating and modifying tricky jsonnet to customize it for their pipelines. Maybe spend even more time trying to work observatorium into their jsonnet.

With this, they can have separate easier-to-operate Go code which they can customize to output YAML they need.

In fact, this is what we do downstream with our Observatorium RBAC. We maintain it in Go because it's easier to modify than jsonnet templating: https://github.com/rhobs/configuration/blob/main/configuration/observatorium/rbac.go. 🙂

docs/proposals/20230130-go-configuration.md
douglascamata: The point of my questions was more to think about these steps and outline them in the proposal, food for thought. Particularly I am missing the point of view of the user of Observatorium in there. 

i.e. why would I have to understand Observatorium's jsonnet lib? Aren't the default values good enough there? Why in the Go based config I can simply "check examples and figure out" and not in jsonnet? 

Better outlining these things will make the proposal more attractive, imho.

docs/proposals/20230130-go-configuration.md
douglascamata: Totally get the objective of `obsctl`. But it's too generic and going against "do one thing and do it well". Which can be a totally fine direction, so if that's the case, all good to me.

docs/proposals/20230130-go-configuration.md
douglascamata: What I mean is that the proposal seems biased toward people that do not already know jsonnet. So saying Observatorium is complex to adopt and has a messy installation model is not a fact, it's relatively to who does it. 

This part of the proposal could be probably rewritten something in the lines of "we assume (and assumptions could be wrong) that there aren't many people out there with enough jsonnet knowledge and for people new to it its usage is complex and messy".

docs/proposals/20230130-go-configuration.md
saswatamcode: I think this is covered by the "Pros" section of the proposal. All of those point out what can be done in Go and not jsonnet or which one is easier. If this feels inadequate, would adding a few examples of this help here?

docs/proposals/20230130-go-configuration.md
saswatamcode: I mean, we are jsonnet users, we deal with it on a daily basis, and know how to use it. And still, find it difficult and tricky to maintain, adopt and scale with it! I'm not sure if this is a biased view?

saswatamcode:
Will add some examples comparing jsonnet and Go tooling for Observatorium specifics as well, as discussed offline. 🙂

periklis (commented):
This is a very refreshing proposal indeed and I very like the idea switching to a Go-based code base for the benefits listed in the proposal document. Thank you @saswatamcode for this effort :rocket:

**As a first note:** We (OpenShift Logging Log Storage) have tried a similar thing back in the inception days of the Loki Operator (See [loki-broker](https://github.com/grafana/loki/blob/main/operator/cmd/loki-broker/main.go)). We experimented on how to present the same CRD e.g. LokiStack and get in return a static but parameterized (via CLI flags) list of YAML manifests. The goal was to use the same manifest machinery as in the operator by re-using code and in parallel deliver the manifests statically for RHOBS. I must admit we never completed the integration into RHOBS, because our Go code base was always behind in AppSRE terms, e.g. sidecars, template-packaging, overwrites, special annotations for deployment recycle just to mention a couple of them. Thus loki-broker lingers around as a never-finished tool but good enough to take notice :)

Back to this proposal. Besides the refreshing ideas outlined in the proposed approach (static typing, IDE support, provider-based templating, go doc generation), I want to highlight a couple of things to bring some contrast in evaluating the purpose and cost/benefits of accepting/rejecting this proposal.

From a user standpoint:
1. The users of this repository is not really interested neither if it is a Go or a JSONNet project. They need imho some solid documentation (still missing) and a driver tool (currently Make) to execute manifests generation for their deployment.
2. User customizations consist of a handful cases: environments, parameterizing manifests, augmenting runnables (deployments/statefulsets). All three cases can be solved nowadays with non-JSONNet and non-Go-based tooling, e.g. [kubectl kustomize](https://kustomize.io/) which is good start for k8s users anyway.

From a developer standpoint:
1. Developers of Observatorium are contributors and as such need to learn a project based on decisions made by the maintainers. They need to first acknowledge that there is some rationale (neither always complete nor always correct) behing these decisions. In fact the most important argument for JSONNet back in 2019/2020 (when Observatorium was born) was that the ecosystem (Prometheus, Thanos, Cortex, Loki, etc.) should be our main JSONnet provider. Keeping configuration-as-code close to the projects ensures that it gets continuously maintained. This still holds as the supreme argument for continuouing with JSONnet besides it's complex nature.
2. Developers contributing here benefit from the fact that this is a collection repo of upstream JSONNet. We need to strive treating it as a collection repo without adding too much secret sauce. Besides some testing facility, I believe this repo still acts as a collection. The only secret (but I would rather call it auxilliary) sauce consists of components like dex, hydra and minio that drive the homerun having a click-and-run-like example/testing environment.

In summary taking into account the above user and developer perspectives the benefits for switching to Go-based configuration-as-code approach do not outweigh the cost over time (e.g. refactoring the code base, re-vendor upstream code as Go, ecosystem disconnect) although it would be very appealing. From what I can tell the only real use case to switch to Go would be if the manifests generation Go code could be re-used by (or vendored from) kuberentes operators (e.g. Loki Operator, Red Hat Observability Operator, Jaeger Operator, Tempo Operator). This would make the static manifest generation to use CRDs as inputs (See loki-broker). In addition it ensures what we write-once can be used in managed SAAS environments (static manifests) as well as on-prem operator-based environments (e.g. Red Hat customer/user ecosystem). This re-usability seems novel to me but again the cost is huge and the overall release coordination requires a solid process. This would remidy older failed experiments like providing an observatorium operator (asked many times but never delivered) (see [observatorium/operator](https://github.com/observatorium/operator)). Nevertheless the make-it-operator-friendly goal is something that in the end will defeat the collection-like core goal of Observatorium's existence. In fact it would became a complex library and who knows how well maintained.

Saying the above I would rather favor to keep the JSONNet code base and improve other neglected areas of this project, i.e.:
1. Improve quickstart documentation.
2. Provide how-to-customize (environments, parameters, augmenting) documentation using ecosystem tools like kustomize.
3. Refactor some JSONNet components to make use of upstream mixins (e.g. components like Loki, Memcached, tracing)
4. Establish releases and tags for this collection (e.g. like apt-get/yum collections `observatorium-2023.01`)

PS: For VSCode users (not me who is sold to Emacs his entire life) there is a great JSONNet language server provided by Grafana: https://github.com/grafana/vscode-jsonnet

configuration_go/k8sutil/k8sutil.go
```diff
@@ -0,0 +1,93 @@
+package k8sutil
+
+import (
+	"fmt"
+
+	mon "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring"
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
```
bwplotka, line 7: nice idea, to simplify exact k8s resources.

Long term, perhaps even those utils have to be versioned by path, so you could migrate and host resources for different API version in the same config.

docs/proposals/20230130-go-configuration.md
bwplotka: I think even wider community does not adopt jsonnet as fast as we envisioned.

So I assume @douglascamata you just want to rephrase existing to something like 
```suggestion
Furthermore, for a project like Observatorium, this leads to a complex-to-adopt installation model, especially for those not familiar with jsonnet. An end-user is forced to go through one of the below,
```

docs/proposals/20230130-go-configuration.md
```diff
@@ -0,0 +1,189 @@
...
+* Scale the steep learning curve of jsonnet initially (if needed), and understand complex Observatorium jsonnet library internals and abstractions (without documentation or discoverability) to be able to configure it exactly according to needs (sometimes even rewrite for customizability)
+* Copy over pre-generated YAML in observatorium/observatorium and change it manually
+* Reference generated YAML and rewrite it in templating the end-user is comfortable with
+
+## Goals
+
+Compose Observatorium configuration as a Go-based library using configuration-as-code concepts, that allows end-users to,
+* Easily discover configuration abstractions, without having to go through steep configuration templating language learning curves and having to explore templating code in-depth
+* Easily utilize pre-defined abstractions for their specific environments, and be able to override configuration at will, without hassle
+* Be able to rely on tested configuration generation, battle-tested versioning via Go modules and automatic documentation via [goDoc](https://blog.golang.org/godoc-documenting-go-code)
```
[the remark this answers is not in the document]
bwplotka: I fully agree. But still, having familiar and well-maintained version manager is a nice thing. Not sure if a strong goal, though. 

I am missing good IDE support with autocompletion as a goal. 🤔

docs/proposals/20230130-go-configuration.md
bwplotka: I understand your concern @douglascamata, but I wouldn't say it's too much. obsctl still interacts with Observatorium resources in this case.

bwplotka (approved):
I like it, but I am bias (Go dev). Would be amazing to experiment with this. I like @periklis comment that at the end, perhaps we could share how somebody can extend things with kustomize, even if we will choose Go path.

The main downsides I see are:
* Importing Go structs for config from source is amazing, but relies on project being nice and maintaining their marshalling as well unmarshalling. Ideally those structs are lightweight and we don't import the whole project e.g Prometheus. Without extending Go module, we either have big deps we don't care about or we copy things over for now. 
* Some boilerplate of Go is unavoidable. Especially frustrating is creating new structs (both ref as well as copy). Normally for configuration compiler can detect type of struct (only when used with interfaces this is impossible), so can be more verbose than it should. Ideally we could improve this someday... (https://github.com/bwplotka/mimic/issues/20)

configuration_go/manifests/config/api-serviceMonitor.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
```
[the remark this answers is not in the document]
saswatamcode: That's a great idea

docs/proposals/20230130-go-configuration.md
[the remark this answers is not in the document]
saswatamcode: Will add the goal to the proposal

moadz:
Thank you @periklis for the in-depth feedback, I wanted to include some final rebuttals here before we close the PR. 


I absolutely accept that there once was inertia with Jsonnet, that meant that the path of least resistance is to adopt. But our integration is not just with Jsonnet, it's with Jsonnet-bundler (which afaics has very limited adoption, is in alpha, is not actively being developed).

> **the ecosystem (Prometheus, Thanos, Cortex, Loki, etc.) should be our main JSONnet provider.**

This is a very good point. At least from a metrics perspective, we are the exclusive maintainers of the kube-thanos jsonnet providers which also see very limited use, so we are hamstrung by the limitations of jsonnet (no static typing, no tests, confusing overwrites driven configuration model, poor IDE support) everywhere we go. If we want to write a go template for loki-stack it would just be a go interface in a project somewhere like kube-thanos. Sure it has to be maintained, but so does jsonnet. 

The community already has a standard, it's the k8s manifests that any of these projects generate. _That_ is the static API, not the method by which you generate the yaml. If we want to consume jsonnet providers upstream, this approach does not preclude us from doing so (either through seperate code generation path, or by stubbing our own interfaces for these projects which is the same cost as creating a jsonnet provider for static k8s). 

> **We need to strive treating it as a collection repo without adding too much secret sauce.**

Well at the moment, for metrics at least, we need to start upstreaming our configuration first, using a versatile DSL that allows us to do the branching we need to for static config in RHOBS. And even if we were to create a Thanos operator instead of static config, we would write that in Go too. 

> **Developers of Observatorium are contributors and as such need to learn a project based on decisions made by the maintainers. They need to first acknowledge that there is some rationale**

I agree that there is a forked use-case here. For people consuming configuration, their experience will not differ. They will still invoke generation via a makefile, and they'll need to make sure they have a local go toolchain installed. 

For people making changes to our config (primarily the RHOBS team) in the observatorium repo (either by importing/overwriting or by raising an MR), they have to make changes to a fully jsonnet toolchain (of which we own and maintain the jsonnet providers for in kube-thanos). 

So in the interests of compatibility (with ourselves) we are hamstringing ourselves with a templating language that is not fit for configuration at the scale that we are using it in. It's not just a conclusion arrived at by Saswata, people have mentioned our Jsonnet toolchain in exit interviews 😆 from our team, and the situation is simply untenable. 

Why not just refactor the jsonnet to make it better? Because the cost would probably be similar to just writing it in go, and we don't solve any of the fundamental problems with the language (bad type support, not testing, overwrites driven templating) and the distribution (jsonnet bundler is not the industry standard, is in alpha and is not being worked on anymore). Jsonnet is trying to do too much. 

If all you want is a simple way to define env specific configuration that can override a set of defaults, use Kustomize, don't use jsonnet. If you need to do anything more than that, you should use a language contributors are familiar with, that allows you to scale contributions to said code (tests/sensible types/build tooling).  

To summarise: I would prefer a slightly deviated toolchain in an active repository that can deliver value to _most_ of the consumers of this project (k8s manifest consumers), than something that is a homogenous e2e toolchain with no support (jsonnet-bundler/jsonnet config in its current form).

moadz (approved):

bwplotka (approved):
💪🏽

saswatamcode:
Thanks for all the comments and reviews! Let's merge this and iterate on it! 🙂

PR #508, opened 2023-03-16
Refactors loki.libsonnet to use loki mixins to generate services

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/LOG-3626

In this PR refactors loki.libsonnet to use the loxi mixins to generate the services for each loki component. If you review the PR you will notice that before each component had 2 services, one per protocol (HTTP and GRPC). In the new version both these services were merged into a single service that exposes two ports one for each protocol.

The loki mixins library is object-based, which means that we cannot simply call a function to generate a resource. However, in this PR we noticed that in components k8s services in loki mixins have a uniform name. The service's name follows the following convention `$COMPONENT_NAME_service` (e.g `distributor_service`), this allowed us to build a function to dynamically generates resources based on their name https://github.com/observatorium/observatorium/pull/508#discussion_r1139006086. Then we simply have to loop through all the services and call the function https://github.com/observatorium/observatorium/pull/508#discussion_r1139007486

### Issues

- [x] Fix CI, loki mixins depend on a file `k.libsonnet` to be present in the $JSONNET_PATH, on my machine I manually created the file importing `github.com/jsonnet-libs/k8s-libsonnet/1.26/main.libsonnet`, this, however, doesn't work for the CI due the vendor folder not being under version control. UPDATE: added `lib` at the `./configuration` that is then used as a library with the flag `-J` in the `make generate` command.
- [x] Remove the function `newGrpcService`, this function is no longer needed, however, it's still used in `newStatefulSet`. UPDATE: this was solved by just using the statefulSet name as now the service has the same name as the statefulSet.
- [x] Update loki.mixins to make compactor service expose GRPC port https://github.com/grafana/loki/pull/8855

### Improvements 
- [ ] Open Issue/PR upstream to refactor ServiceMonitors so only one is  generated per component 
- [X] Open Issue/PR upstream to make service port names uniforms namely `query_frontend`, `query_scheduler` UPDATE: this is by design in order for DNS SRV discovery to work, added comment
- [x] PR to normalize service name for both query frontend/scheduler https://github.com/grafana/loki/pull/8880

### Unanswered questions

- [x] Loki mixins generates Services with two ports instead of two separate services, is this a problem for observatorium? Was there any historic reason to have two separate services?  UPDATE: see https://github.com/observatorium/observatorium/pull/508#issuecomment-1474998694 TL;DR there is no problem
- [x] Loki compactor service mixin only exposes one port (http), whereas the old svc exposes two (http & grpc), is this a problem? UPDATE: see https://github.com/observatorium/observatorium/pull/508#issuecomment-1474998694 TL;DR this is a bug
- [x] When gossip ring us enabled the services of the components will also expose that port, is that an issue? UPDATE: no problem https://github.com/observatorium/observatorium/pull/508#issuecomment-1479982233


periklis:
@JoaoBraveCoding Awesome work done here! Can you please elaborate in the description any issues, remaining differences, things that need to migrated to the new manifests e.g. on rollout time?

periklis:
> Loki mixins generates Services with two ports instead of two separate services, is this a problem for observatorium? Was there any historic reason to have two separate services?

For historic reasons when we started with Loki we required/assumed all GRPC client/server discovery to be done by GRPC itself. In fact GRPC does load-balancing itself (when using DNS SRV records) and this required to use a dedicated headless k8s service for that purpose. 

OTOH using memberlist (our choice for the hash ring based on a gossip protocol) means that the peer pods (e.g. distributors, ingester, queriers) use the k8s service for DNS resolution and after joining the gossip ring they advertise their instance IP address and port to all peers. In turn GRPC connections become point-to-point (e.g. from a distributor IP (source) to an ingester IP (destination) which makes GRPC load-balancing obsolete.

FWIW, although I cannot track if the Loki mixin ever used GRPC via k8s headless services in their code base, I believe we introduced the service separation out of uncertainty on how things work (back on 2020 when we started using Loki). AFAICS in this code base the only GRPC headless service required is for registering the queriers back to schedulers in `_config.loki.frontend.scheduler_address`

> Loki compactor service mixin only exposes one port (http), whereas the old svc exposes two (http & grpc), is this a problem?

This sounds like a bug or miss in the official Loki mixin. FWIW the compactor never talked to the rest of the system except exposing HTTP metrics thus a single service port makes sense. However with the introduction of the deletion API two weird things happened:
1. The initial internal deletion call stack from the queriers to the compactor was accidentally implemented via HTTP. This however was a hindsight introducing this feature too aggressively in the Loki code base. As per all internal component to component communication should be implemented via GRPC.
2. In Loki v2.7.1 @sandeepsukhani implemented the GRPC compactor client via GRPC in grafana/loki/pull/7804

When we introduced 2.7 in Observatorium we bumped directly to v2.7.2 thus we started using the GRPC compactor client right ahead. At least we put that configuration in [#506](https://github.com/observatorium/observatorium/pull/506/files#diff-53a0b44e1437e6a3490d2adb55ff1c6952ed979de33f7045c225421f9d2ed385R198-R203). In general we don't expose the Deletion API in observatorium or RHOBS at all. We still have concerns with that API.

JoaoBraveCoding:
> FWIW, although I cannot track if the Loki mixin ever used GRPC via k8s headless services in their code base, I believe we introduced the service separation out of uncertainty on how things work (back on 2020 when we started using Loki). AFAICS in this code base the only GRPC headless service required is for registering the queriers back to schedulers in _config.loki.frontend.scheduler_address

Correct, this is confirmed in this PR https://github.com/grafana/loki/pull/4154. However apart from this apparently there is no issue with both the HTTP and GRPC services being merged in a single one

> This sounds like a bug or miss in the official Loki mixin.

I agree, before merging this PR that should be corrected

configuration/components/loki.libsonnet
```diff
@@ -485,6 +486,18 @@ function(params) {
...
+  rhobsLoki::
+    lokiMixins {
+      _config+:: {
+        multi_zone_ingester_enabled: false,
+        // Compactor
+        using_boltdb_shipper: true,
+        ruler_enabled: true,
+        query_scheduler_enabled: true,
+        use_index_gateway: true,
+      },
```
periklis, line 498: Do you consider adding the [membelist_ring_enabled](https://github.com/grafana/loki/blob/5a08a6bcb9ecc816328cb7859bbdcc7cf9d004c5/production/ksonnet/loki/memberlist.libsonnet#L18) in this PR or a separate one? This options enables another k8s service for the gossip ring.

periklis (changes requested):
Looks overall good to me. I suggest that in case your approach is to make small self-containing installments of moving manifests to the upstream Loki mixing to have additionally to the ServiceMonitors to make this PR work end-to-end and self-sufficient:
https://github.com/observatorium/observatorium/blob/37ee741596df29a3f2dbcdebd3a108f8e2ffc1f4/configuration/components/loki.libsonnet#L776-L798

AFAICS the port name used to be `metrics` and the upstream mixin renames it to `http-metrics`.

JoaoBraveCoding:
> AFAICS the port name used to be metrics and the upstream mixin renames it to http-metrics.

Very good point! I was completely forgetting this, I'll make sure to package in this PR ServiceMonitors as well!

> Do you consider adding the [membelist_ring_enabled](https://github.com/grafana/loki/blob/5a08a6bcb9ecc816328cb7859bbdcc7cf9d004c5/production/ksonnet/loki/memberlist.libsonnet#L18) in this PR or a separate one? This options enables another k8s service for the gossip ring.

I see, I'll check this! If it doesn't have major implications I'm happy to also ship it in this PR

JoaoBraveCoding:
@periklis I'm not a big fan of how the ServiceMonitor from the loki.mixins looks like. It's a single ServiceMonitor to monitor all the different components (see img). So in this PR I'll just fix the port name and then we can open a PR upstream to refactor it.
<details>
  <summary>Image</summary>

image

</details>

In regards to the gossip service I've also added it to this PR, I added however 1 "Unanswered question" related to it

periklis:
>  When gossip ring us enabled the services of the components will also expose that port, is that an issue?

No this is ok to expose the port. I guess we were doing this before the same and that's ok:
https://github.com/observatorium/observatorium/blob/37ee741596df29a3f2dbcdebd3a108f8e2ffc1f4/configuration/components/loki.libsonnet#L634-L635

periklis:
> @periklis I'm not a big fan of how the ServiceMonitor from the loki.mixins looks like. It's a single ServiceMonitor to monitor all the different components (see img). So in this PR I'll just fix the port name and then we can open a PR upstream to refactor it.
> Image
> 
> In regards to the gossip service I've also added it to this PR, I added however 1 "Unanswered question" related to it

I agree this one huge ServiceMonitor is a code smell anyhow. Let's keep such things on list to deviate from. We do not necessarily need to go 100% with the mixin if justifiable. IMHO I would keep our ServiceMonitor code for better visibility on the production.

periklis (commented):
LGTM. This is a huge improvement :+1:

periklis (approved):

configuration/components/loki.libsonnet
```diff
@@ -285,34 +277,19 @@ function(params) {
...
             // Mixins don't support prefixes in resources names so we have to overwrite
-            // local schedulerService = newService('query_scheduler'),
-            // scheduler_address: '%s.%s.svc.cluster.local:9095' % [schedulerService.metadata.name, loki.config.namespace],
-            // TODO(JoaoBraveCoding): replace when working on services
-            scheduler_address: '%s.%s.svc.cluster.local:9095' % [
-              loki.config.name + '-query-scheduler-grpc',
-              loki.config.namespace,
-            ],
+            local schedulerService = newService('query_scheduler'),
+            scheduler_address: '%s.%s.svc.cluster.local:9095' % [schedulerService.metadata.name, loki.config.namespace],
```
periklis, line 281: Doesn't the scheduler service need to be headless for this configuration to work? See:

https://github.com/grafana/loki/blob/de2b0c6e9dcce028ed69c6e40f75e1ff0c081358/production/ksonnet/loki/query-scheduler.libsonnet#L7-L11

periklis (changes requested):

configuration/components/loki.libsonnet
```diff
@@ -261,13 +258,8 @@ function(params) {
...
             // before removing this
-            // local compactorService = newService('compactor'),
-            // compactor_grpc_address: '%s.%s.svc.cluster.local:9095' % [compactorService.metadata.name, loki.config.namespace],
-            // TODO(JoaoBraveCoding): replace when working on services
-            compactor_grpc_address: '%s.%s.svc.cluster.local:9095' % [
-              loki.config.name + '-compactor-grpc',
-              loki.config.namespace,
-            ],
+            local compactorService = newService('compactor'),
+            compactor_grpc_address: '%s.%s.svc.cluster.local:9095' % [compactorService.metadata.name, loki.config.namespace],
```
periklis, line 262: nit. Suggestion to move all ports (grpc, http, gossip) to constants. WDYT?

configuration/components/loki.libsonnet
```diff
@@ -285,34 +277,19 @@ function(params) {
...
             // Mixins dont support prefixes in resources names so we have to overwrite
-            // local schedulerService = newService('query_scheduler'),
-            // scheduler_address: '%s.%s.svc.cluster.local:9095' % [schedulerService.metadata.name, loki.config.namespace],
-            // TODO(JoaoBraveCoding): replace when working on services
-            scheduler_address: '%s.%s.svc.cluster.local:9095' % [
-              loki.config.name + '-query-scheduler-grpc',
-              loki.config.namespace,
-            ],
+            local schedulerService = newService('query_scheduler'),
+            scheduler_address: '%s.%s.svc.cluster.local:9095' % [schedulerService.metadata.name, loki.config.namespace],
```
periklis, line 292: Same here, shouldn't that be the headless service for the scheduler?

periklis (changes requested):

configuration/components/loki.libsonnet
JoaoBraveCoding: IIUC, if we were NOT using the membership ring yes this would need to be a headless service. However, since we always run with the membership ring AFAIK this doesn't need to be a headless service. This reply is based on the fact that we got a green CI and what I understood from this comment https://github.com/observatorium/observatorium/pull/508#issuecomment-1474998694 but maybe I misunderstood it, let me know!

The mixins provides the two services https://github.com/grafana/loki/blob/de2b0c6e9dcce028ed69c6e40f75e1ff0c081358/production/ksonnet/loki/query-scheduler.libsonnet#L70-L89

periklis: According to the docs the `frontend` block to discover scheduler pods is not part of memberlist, it simplified by using the DNS SRV records and leave anything else to GRPC: https://grafana.com/docs/loki/v2.8.x/configuration/#frontend

```yaml
# DNS hostname used for finding query-schedulers.
# CLI flag: -frontend.scheduler-address
[scheduler_address: <string> | default = ""]
```

configuration/components/loki.libsonnet
periklis: According to the docs the `frontend_worker` block to discover scheduler pods is not part of memberlist, it simplified by using the DNS SRV records and leave anything else to GRPC: https://grafana.com/docs/loki/v2.8.x/configuration/#frontend_worker

```yaml
# Hostname (and port) of scheduler that querier will periodically resolve,
# connect to and receive queries from. Only one of -querier.frontend-address or
# -querier.scheduler-address can be set. If neither is set, queries are only
# received via HTTP endpoint.
# CLI flag: -querier.scheduler-address
[scheduler_address: <string> | default = ""]
```

PR #509, opened 2023-03-23
Add AntiAffinity to Loki and Memcached

aminesnow:
This PR introduces a flag to add `AntiAffinity` to Loki component manifests and Loki caches' manifests.

### DIFF
--- a/configuration/components/loki.libsonnet
+++ b/configuration/components/loki.libsonnet
@@ -48,7 +48,7 @@ local defaults = {
     compactor: {
       withLivenessProbe: true,
       withReadinessProbe: true,
-      withAffinity: true,
+      withPodAntiAffinity: true,
       resources: {
         requests: { cpu: '100m', memory: '100Mi' },
         limits: { cpu: '200m', memory: '200Mi' },
@@ -58,7 +58,7 @@ local defaults = {
     distributor: {
       withLivenessProbe: true,
       withReadinessProbe: true,
-      withAffinity: true,
+      withPodAntiAffinity: true,
       resources: {
         requests: { cpu: '100m', memory: '100Mi' },
         limits: { cpu: '200m', memory: '200Mi' },
@@ -79,7 +79,7 @@ local defaults = {
     ingester: {
       withLivenessProbe: true,
       withReadinessProbe: true,
-      withAffinity: true,
+      withPodAntiAffinity: true,
       resources: {
         requests: { cpu: '100m', memory: '100Mi' },
         limits: { cpu: '200m', memory: '200Mi' },
@@ -102,7 +102,7 @@ local defaults = {
     index_gateway: {
       withLivenessProbe: true,
       withReadinessProbe: true,
-      withAffinity: true,
+      withPodAntiAffinity: true,
       resources: {
         requests: { cpu: '100m', memory: '100Mi' },
         limits: { cpu: '200m', memory: '200Mi' },
@@ -112,7 +112,7 @@ local defaults = {
     querier: {
       withLivenessProbe: true,
       withReadinessProbe: true,
-      withAffinity: true,
+      withPodAntiAffinity: true,
       resources: {
         requests: { cpu: '100m', memory: '100Mi' },
         limits: { cpu: '200m', memory: '200Mi' },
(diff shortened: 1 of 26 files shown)


aminesnow:
CC @periklis

configuration/components/loki.libsonnet
```diff
@@ -48,6 +48,7 @@ local defaults = {
     compactor: {
       withLivenessProbe: true,
       withReadinessProbe: true,
+      withAffinity: true,
```
douglascamata, line 51: Sounds like this could be called `withPodAntiAffinity`, as that's effectively what's added?

douglascamata (approved):
LGTM. @periklis might also want to have a look at this.

configuration/components/loki.libsonnet
```diff
@@ -652,6 +662,29 @@ function(params) {
       if std.length(config.resources) > 0
     },
 
+  local affinity(component, namespace) = {
+    podAntiAffinity: {
+      preferredDuringSchedulingIgnoredDuringExecution: [
+        {
+          weight: 100,
```
periklis, line 669: Why is the weight here `100`. Does this have a meaning?

configuration/components/loki.libsonnet
```diff
@@ -652,6 +662,29 @@ function(params) {
...
+          podAffinityTerm: {
+            topologyKey: 'kubernetes.io/hostname',
+            namespaces: [namespace],
```
periklis, line 672: Can you elaborate a bit in the description on what happens if the target cluster will not have enough distinct nodes, e.g. for 3 ingesters schedule each on a different node?

periklis (changes requested):
LGTM, however I would rather put `withPodAntiAffinity: false` in the default configuration because this node placement directive requires checking the node capacity upfront. Thus the user should opt-in by making a explicit choice.

configuration/components/loki.libsonnet
aminesnow: Here we use `preferredDuringSchedulingIgnoredDuringExecution`, which is a soft matcher and is ignored when unable to be fulfilled.

periklis (commented):
What is the reason to prefer `preferredDuringSchedulingIgnoredDuringExecution` over `requiredDuringSchedulingIgnoredDuringExecution`? Why can't the user make a choice?

FYI, rendering the upstream [loki mixing](https://github.com/grafana/loki/tree/44f1d8d7f6b1fb9b7fac32e913bd6517d15bad3f/production/ksonnet/loki) tells me that they use `requiredDuringSchedulingIgnoredDuringExecution` for ingesters, rulers, query-frontend and query-scheduler pods. This means they apply strict scheduling for at least the components which:
- Either have impact/ensure replication (i.e. ingester)
- Or spread the load on executing queries (query-frontend, query-scheduler)
- Or spread the load on executing rules

WDYT to get in touch with the Loki team and ask for advice?

aminesnow:
> What is the reason to prefer `preferredDuringSchedulingIgnoredDuringExecution` over `requiredDuringSchedulingIgnoredDuringExecution`? Why can't the user make a choice?
> 
> FYI, rendering the upstream [loki mixing](https://github.com/grafana/loki/tree/44f1d8d7f6b1fb9b7fac32e913bd6517d15bad3f/production/ksonnet/loki) tells me that they use `requiredDuringSchedulingIgnoredDuringExecution` for ingesters, rulers, query-frontend and query-scheduler pods. This means they apply strict scheduling for at least the components which:
> 
>     * Either have impact/ensure replication (i.e. ingester)
> 
>     * Or spread the load on executing queries (query-frontend, query-scheduler)
> 
>     * Or spread the load on executing rules
> 
> 
> WDYT to get in touch with the Loki team and ask for advice?

The reason is because I don't want this constraint to be hard, ie, prevent the deployment when the condition is not fulfilled. 
I don't see what advice could the upstream give me, since we don't run nearly anywhere near the same conditions. We should focus on what we want to achieve for our use case. 
For me, this is about making one step towards spreading the pods around the nodes without making it impossible to deploy them. If this is not the goal here then I should close this PR and we should spike this.

periklis (approved):

PR #516, opened 2023-04-11
logs: moves some loki CLI args to the config

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/LOG-3902

Moves some loki configuration that previously was being passed by CLI arguments to the loki config directly in the form of ENV vars that will then be expanded to their actual values. We need this because upstream we will want to update these values without having to regenerate the templates. Furthermore, some of the values are secrets and are only known at rollout time

### DIFF
--- a/configuration/components/loki.libsonnet
+++ b/configuration/components/loki.libsonnet
@@ -207,7 +207,7 @@ function(params) {
   assert std.isObject(loki.config.etcd) : 'etcd has to be an object',
   assert std.isArray(loki.config.etcdEndpoints) : 'etcdEndpoints has to be an array',
 
-  rhobsLoki::
+  mixins::
     lokiMixins {
       // Some values in the loki config are ENV var, we do this because upstream
       // we will want to update them without having to regenerate the templates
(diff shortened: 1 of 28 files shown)


configuration/components/loki.libsonnet
```diff
@@ -582,83 +618,43 @@ function(params) {
...
+        if component == 'querier' then [
+          // env only supports strings hence the addition with ''
+          envVarFromValue('LOKI_QUERIER_MAX_CONCURRENCY', '' + loki.config.query.concurrency),
+        ] else []
+      ) + (
+        if component == 'ingester' then [
+          // env only supports strings hence the addition with ''
+          envVarFromValue('LOKI_INGESTER_WAL_REPLAY_MEMORY_CEILING', '' + loki.config.wal.replayMemoryCeiling),
+        ] else []
+      ) + (
```
JoaoBraveCoding, line 637: @periklis I want to double-check something with you. I'm afraid that if we do things like this then components that don't have a given env var will not correctly perform ENV expansion and will have a problem parsing the configuration, the CI is green so apparently this is not an issue but I wanted your opinion.

One solution if this is indeed a problem is to provide default values which AFAIK the loki config supports
https://grafana.com/docs/loki/latest/configuration/#use-environment-variables-in-the-configuration

JoaoBraveCoding: Me and Peri talked about it and decided to simply add the ENV vars to all the components deployments/statefulsets

configuration/components/loki.libsonnet
```diff
@@ -492,6 +523,45 @@ function(params) {
...
+    local envVarFromValue(name, value) = { name: name, value: value };
+    local envVarFromSecret(name, secretName, secretKey) = { name: name, valueFrom: { secretKeyRef: { name: secretName, key: secretKey } } };
+    local osc = loki.config.objectStorageConfig;
+    local rsc = loki.config.rulesStorageConfig;
+    [
+      envVarFromValue('LOKI_LOG_LEVEL', loki.config.logLevel),
+      // The following values are added with '', because env only supports strings
+      envVarFromValue('LOKI_REPLICATION_FACTOR', '' + loki.config.replicationFactor),
+      envVarFromValue('LOKI_QUERIER_MAX_CONCURRENCY', '' + loki.config.query.concurrency),
+      envVarFromValue('LOKI_INGESTER_WAL_REPLAY_MEMORY_CEILING', '' + loki.config.wal.replayMemoryCeiling),
```
JoaoBraveCoding, line 543: Still not sure if this should be considered an int or a string since as far as I saw this can may or may not have the unit suffix (KB, MB, GB) applied [1]. Currently, it's coded as if it was an int. I'll test tomorrow morning to be sure

[1] https://grafana.com/docs/loki/latest/configuration/#ingester

periklis: It is not a value with unit: https://github.com/rhobs/configuration/blob/main/services/observatorium-logs-template.jsonnet#L80

periklis (approved):

PR #518, opened 2023-04-12
Fix local quickstart environment

periklis:


squat (approved):
Sounds good! We should have some way of validating that our configuration is actually correct! This whole repo is about configuration but we don't really validate it all

PR #527, opened 2023-05-24
Gubernator peer discovery 

philipgough:
This change ensures we use the correct peer discovery mechanism by specifying the `GUBER_PEER_DISCOVERY_TYPE ` env var.

It also changes the Service to a headless one to ensure we can load balance between peers.


saswatamcode (approved):
LGTM!

PR #530, opened 2023-07-18
Add initial Thanos Query lib in Go config

saswatamcode:
This PR adds a Thanos Query library for our Go configuration. 🙂 

It does the following,

- Adds ability to generate Querier K8s config from Go, and adds placeholder files for the rest.
- Refactors k8sutil, to have DeploymentGenericConfig struct, that allows passing in certain generic k8s deployment options to various deployments. The struct can be embedded into every deployment we generate via go i.e Observatorium API and Thanos Querier
- Adds a schema dir, for containing certain tracing structs from Thanos upstream. We copy them over to add `omitempty` to struct fields (which is forbidden upstream due to doc gen) and also avoid adding a dep to `go.mod`

Would love any thoughts on this.


configuration_go/k8sutil/genericdeployment.go
```diff
@@ -0,0 +1,155 @@
...
+	AdditionalPodVolumes          []corev1.Volume
+	AdditionalServicePorts        []corev1.ServicePort
+	AdditionalServiceMonitorPorts []monv1.Endpoint
+}
+
+// GenericDeploymentConfig represents certain config fields
+// that might be useful to add/override in a Deployment. It contains
+// fields of both DeploymentSpec and PodSpec.
+// It also has method defined for overriding any default values.
+type GenericDeploymentConfig struct {
```
philipgough, line 22: `DeploymentConfig` was its own thing in OpenShift for a long time. Do you think the naming here could become any source of confusion in that respect?

philipgough (commented):
This is in really good shape already. Nice work!

configuration_go/k8sutil/genericdeployment.go
saswatamcode: Ah, I wasn't aware..let's rename then

saswatamcode: Have renamed it.

configuration_go/k8sutil/genericdeployment.go
```diff
@@ -0,0 +1,154 @@
...
+	AdditionalPodVolumes          []corev1.Volume
+	AdditionalServicePorts        []corev1.ServicePort
+	AdditionalServiceMonitorPorts []monv1.Endpoint
+}
+
+// DeploymentGenericConfig represents certain config fields
+// that might be useful to add/override in a Deployment. It contains
+// fields of both DeploymentSpec and PodSpec.
+// It also has method defined for overriding any default values.
+type DeploymentGenericConfig struct {
```
thibaultmg, line 22: I feel we could do something simpler here by avoiding filling transitory fields that will then be mapped to the `appsv1.Deployment` struct. Couldn't we directly manipulate this target structure here and avoid having to map everything back to it in the `Manifests()` methods used in `abstr/kubernetes`.
Also, we are coupling this implementation with the `Manifests()` methods as additional capability provided here will probably need modifications in `Manifests()` to map the fields in the target `appsv1.Deployment` struct.
I might have overlooked some details preventing this approach. Feel free to discard this proposition.

thibaultmg (approved):
Awesome! I like this much more than jsonnet.
Left a comment for your consideration.

configuration_go/k8sutil/genericdeployment.go
saswatamcode: Thanks! This seems doable and a good idea, will iterate on this in a later PR 🙂

saswatamcode:
Thanks! Let's merge and iterate on this!

PR #531, opened 2023-07-28
[WIP] Add thanos compactor to configuration_go

thibaultmg:
I am using the new Go configuration and it is great! I feel though we could simplify some things, resulting in less code lines to write:

- I introduce a `clioption` package that use struct tags to map fields to the cmd option. I like being able to map directly the code to cli options/flags. Represents less mental overhead to me. Also, less code lines to generate the container args.
- I refactored the k8sutils way of handling deployment config to reduce repetition and code lines. I removed the `DeploymentGenericConfig` struct in favor functions updating directly the `Deployment` object. The idea is to cover common use cases. And for specifics, the compactor exposes kube objects for custom post processing.

This is very much work in progress but it compiles. Sharing early for feedbacks.

Cheers


thibaultmg:
Opening new pull request for final version. Closing this one.

PR #537, opened 2023-08-23
Add Thanos compactor to go config

thibaultmg:
Adds Thanos Compactor deployment Go code. Here are few changes I propose for handling these new go configurations:

- Added the `cmdoption` package to streamline args generation in containers. This allows mapping the component commands flags to struct fields using annotations. These fields entirely align with the command flags, making it more easy to use.
- Proposed an updated version of the `k8sutil` package. It now implements kubernetes resources providers. The package also adds some helpers for generating common k8s structures.
- The Thanos Compactor implements these changes: 
  - `CompactorOptions` sets the command options/flags of the compactor.
  - Some helper functions create default providers for kubernetes objects.
  - It is up to the caller to ensure the coherence of the `CompactorOptions` with the providers values. No verification is done to see if an environment variable is missing, and so on. This is deliberate as the CI is responsible to make these checks. Doing them in this code would overly complicate things.


moadz (dismissed):
LGTM for now to unblock, but I think we need to have a design discussion around the pattern we use for common templates in observatorium/observatorium. Querier currently uses a builder, whilst your approach uses a factory of sorts. 

The decision from the above will influence how the community will hold these templates in the future. 

wrt `cmdopts` +1000, I really prefer that approach for template CLI arg declaration. _very idiomatic`

configuration_go/main.go
```diff
@@ -290,6 +293,50 @@ func main() {
...
+	kubeCfg.ServicePorts = []corev1.ServicePort{
+		{
+			Name:       "https",
+			Port:       httpsPort,
+			TargetPort: intstr.FromInt(int(httpsPort)),
+		},
+	}
+
+	// Create compactor, generate manifests, add annotations as post processing, generate files.
+	cpt := compactor.NewCompactor(compactorOptions, kubeCfg)
```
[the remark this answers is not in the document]
moadz: With your approach, it would require shared `compactorOptions` and differentiated `kubeCfg` as opposed to a shared core struct that represents 'a compactor (and all its related bits)'

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
```diff
@@ -1 +1,375 @@
...
+	dataVolumeName      string = "data"
+	defaultHTTPPort     int    = 10902
+	defaultNamespace    string = "observatorium"
+	servicePortName     string = "http"
+	thanosContainerName string = "thanos"
+)
+
+// CompactorOptions represents the options/flags for the compactor.
+// See https://thanos.io/tip/components/compact.md/#flags for details.
+type CompactorOptions struct {
```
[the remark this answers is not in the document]
thibaultmg: Yes this is probably feasible with additional work. I'll think about it. 
But even today, it would be great if Thanos exported all these flags so that we would simply import them. @saswatamcode wanted to do that, but does not seem straightforward at all with current Thanos flags implementation.
With `cmdopts` we still have to maintain the flags list ourselves. Seems reasonable to me.

configuration_go/main.go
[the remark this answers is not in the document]
thibaultmg: Yes, you framed it well. Two points:

- The builder pattern used in the querier implementation looks more elegant to use but is more verbose to implement. As it is done currently, you both have to declare the `GenericConfig` with all its fields, as well as one builder per field. This doesn't provide simplification in my opinion. Directly setting the attribute of the struct seems more straightforward.
- I agree that if you want to create several deployments with a few changes:
  - The builder will be simpler to use as you just need to derive the base compactor object using the builder `With...` methods 
  - While with the factory approach here, you must clone, modify the `compactorOptions` and `kubeCfg` and finally instantiate a new compactor
  - This limitation in this PR could be circumvented by exposing these two attributes in the compactor. Then if you want to derive a base configuration, you simply modify the compactor struct. Here I made them private to limit the public part of the compactor to the minimum but did not think about this use case.
  
Should I make the `compactorOptions` and `kubeCfg` fields public in the compactor?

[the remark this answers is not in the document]
thibaultmg: I have heavily refactored the code. I think it is cleaner now. I was able to implement some kind of factory to generate kubernetes objects that should be versatile enough to be used by all thanos components. This further reduces the amount of code needed to add these components. Let me know what you think.

configuration_go/k8sutil/v2/providers.go
```diff
@@ -0,0 +1,430 @@
...
+	return c.MonitorPorts
+}
+
+// GetVolumeClaims returns the volume claims that the container requires.
+func (c *Container) GetVolumeClaims() []VolumeClaim {
+	return c.VolumeClaims
+}
+
+// AddConfigMap adds a config map to the container.
+func (c *Container) AddManifest(filename string, mf ManifestProvider) {
```
douglascamata, line 151: Missmatching name according to signature and function's docs. `AddConfigMap` seems to be correct.

configuration_go/k8sutil/v2/providers.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// PersistentVolumeClaimProvider is the interface to be implemented by pods that require volume claims.
+type PersistentVolumeClaimProvider interface {
+	GetVolumeClaims() []VolumeClaim
+}
+
+// PodProvider is the interface to be implemented by pods to be used in a StatefulSet or Deployment.
+type PodProvider interface {
+	MakePodSpec() corev1.PodSpec
```
douglascamata, line 182: Why is the `PodProvider` the only one with a function like this? I have the impression it should be the `ManifestProvider` interface.

configuration_go/k8sutil/v2/builders.go
```diff
@@ -0,0 +1,132 @@
...
+	defaultTerminationGracePeriodSeconds = 120
+	ServiceManifestKey                   = "service"
+	ServiceMonitorManifestKey            = "serviceMonitor"
+	ServiceAccountManifestKey            = "serviceAccount"
+	StatefulSetManifestKey               = "statefulSet"
+)
+
+type StatefulSetBuilder struct {
+	MetaConfig                    MetaConfig
+	Containers                    []ContainerProvider
```
douglascamata, line 20: This sounds weird. A statefulset contains pods, not container. 🤔

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
```diff
@@ -1 +1,168 @@
...
+				Port:           "http",
+				RelabelConfigs: k8sutilv2.GetDefaultServiceMonitorRelabelConfig(),
+			},
+		},
+	}
+}
+
+// NewStatefulSet returns a new statefulset container provider for the compactor.
+// It includes the compactor container and the volume claims.
+func NewSSContainerProvider(c *CompactorOptions) *k8sutilv2.Container {
```
douglascamata, line 140: This returns a new **statefulset** **container**? 🤔 

Shouldn't it be only statefulset or only container? Of course a statefulset has pods, that has containers.

I am missing a high-order function to create the Compactor here. Something that creates the `compactorStatefulSet` variable like in there:  https://github.com/observatorium/observatorium/pull/537/files#diff-cd4defa334f8d951bb98d5d340bd8afb7901fc72a89fcf8fa8bd5888f6666fc7R314-R337

douglascamata (commented):
Few questions and suggestions 🤓

configuration_go/abstr/kubernetes/cmdoption/cmdopt_test.go
```diff
@@ -0,0 +1,144 @@
...
+	Float        float64       `opt:"float"`
+	Bool         bool          `opt:"bool"`
+	Duration     time.Duration `opt:"duration"`
+	Sub          SubStruct     `opt:"sub"`
+	NoValue      bool          `opt:"no-value,noval"`
+	Repeat       []string      `opt:"repeat"`
+	SingleHyphen int           `opt:"single,single-hyphen"`
+}
+
+func TestCmdOptions(t *testing.T) {
```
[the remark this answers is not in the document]
thibaultmg: Sure, I'm always up for good coverage. Regarding the tests you propose, I am excluding some because there seem to be some confusion about the tags purpose because of how I named them in the tests. 
The first tag value is just the name that is used to generate the arg, it is not interpreted.

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
```diff
@@ -1 +1,168 @@
...
+			k8sutil.InstanceLabel:  "observatorium",
+			k8sutil.PartOfLabel:    "observatorium",
+			k8sutil.ComponentLabel: "database-compactor",
+			k8sutil.VersionLabel:   defaultImageTag,
+		},
+	}
+}
+
+// NewBaseContainerProvider returns a new container provider for the compactor.
+func NewBaseContainerProvider(c *CompactorOptions) *k8sutilv2.Container {
```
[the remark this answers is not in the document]
thibaultmg: yes good point changing that

configuration_go/abstr/kubernetes/cmdoption/cmdopt.go
```diff
@@ -0,0 +1,228 @@
...
+
+// GetOpts is used to generate command line options from a struct
+// by mapping the fields values to the option name defined in the opt tag (first position).
+// GetOpts returns a slice of strings, each string representing an option.
+// Following types are supported: string, int, bool, float64, time.Duration, slice of supported types, sub struct implementing the Stringer interface.
+// Pointer types are used if not nil. Private fields are ignored.
+// Additional tags can be added to the opt tag, separated by a comma, to modify its default behavior:
+// - noval: the option is added without a value if the field is true.
+// - single-hyphen: the option is prefixed with a single hyphen instead of a double hyphen.
+func GetOpts(obj interface{}) []string {
```
saswatamcode, line 26: Ok, first off I think this is a really cool way of getting the flags from a struct, and I see it being super clean approach. I'd actually love to see some form of CLI package with this at some point. 🙂 

But, this is still a maintenance burden in the context of Observatorium. I'm keen to merge this, but with a TODO, to upstream this in the future, so that this config is derived from the same place as where the original struct is defined (i.e in Thanos).

configuration_go/k8sutil/v2/builders.go
```diff
@@ -0,0 +1,132 @@
...
+	Replicas                      int32
+	TerminationGracePeriodSeconds int64
+	Affinity                      *corev1.Affinity
+	SecurityContext               *corev1.PodSecurityContext
+	ServiceAccountName            string
+}
+
+func (s *StatefulSetBuilder) makeAffinity() *corev1.Affinity {
```
saswatamcode, line 28: Any reason to keep these methods private?

configuration_go/k8sutil/v2/utils.go
saswatamcode, line 1: I really like these "sane defaults", that can be used throughout! 🙂 

What does v2 indicate tho?

configuration_go/k8sutil/v2/builders.go
saswatamcode, line 1: I'm not sure I understand the purpose of this builders file. 🤔 
It seems to encode quite a lot of logic, just to avoid some repetition.

As this code is exclusive to generate config, we should avoid doing DRY too much, and be verbose here. Repetition is fine here, for the sake of clarity. Refer to https://github.com/bwplotka/mimic/blob/main/docs/best_practice.md 🙂

configuration_go/k8sutil/v2/providers.go
thibaultmg: Yup there is a mismatch. The objective here is to make the container self contained with everything it needs to work properly, including configmaps and secrets. To do that, it implements the `ManifestsProvider` interface. Here this is a function to add any type of manifest with its expected file name. `AddManifest` is what I wanted. 
Updating.

configuration_go/k8sutil/v2/providers.go
thibaultmg: The Container also has a `GetContainer` function. It does not implement directly the `ManifestProvider` interface because it is not a final object. It is embedded in a StatefulSet here that needs to know what volume claims are expected for example.

configuration_go/k8sutil/v2/builders.go
thibaultmg: The idea with these "builders" is to standardise how StatefulSet or Deployments are created. It flattens and provides most important parameters for the Pod and StatefulSet and then takes care for building everything, from the Pod itself, the statefulSet and all resources expected by the containers. 
It does not do much, but I expect that the need will be the same for all components, and they won't need to implement this logic.
The alternatives are:
- this is implement in each component (in a simplified manner as here it is more generic)
- this is done by the caller that follows each step

Also the purpose of this builder is that it can be easily derived. You don't have to rebuild everything yourself chaining from the container, to final objects

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
thibaultmg: The idea was to provide the possibility to deploy the compactor either as a statefulSet or a Deployment (not sure it make sense). Providing more flexibility for the lib user.
But, yes I'll add a default StatefulSet returned by the compactor, and probably remove this function. It seems to align more with what is expected from this lib and @saswatamcode also expects that.

configuration_go/abstr/kubernetes/cmdoption/cmdopt.go
thibaultmg: Cool! But I don't know if it will ever make it to the other side (parsing args and map them to the struct). There are probably better libs for that.

For now, I don't see how we can escape the maintenance burden of these flags. Because either we provide a tailored API for these components like what is done in `query` component. But we still have to add new methods when new flags appear. Either we import directly from Thanos, but this may take a while before being possible.

Also, if a flag is not yet listed here but present on Thanos, you can still add it as an ExtraOpt or manually in the container.Args.

configuration_go/k8sutil/v2/builders.go
thibaultmg: It is not meant to be called by the user. If the user wants a custom affinity, he can set the attribute directly. This is just code organisation. I can push them at the bottom.

configuration_go/k8sutil/v2/utils.go
thibaultmg: I am not sure that v2 brings consensus, didn't want to rewrite the query if not. Thus I isolated it for now.

configuration_go/k8sutil/v2/builders.go
thibaultmg: Interesting doc. I'll read it in details. Regarding doing DRY too much, I agree we must find the right balance. Maybe this builder is too much. But as we are deploying a cohesive set of components, they will all do it the same way. So it might make sense having a common way of building StatefulSets or Deployments. This is the idea. But yes, you lose some clarity.
If component creators don't think it is needed, we can totally remove it, and implement simpler logics in each observatorium component

thibaultmg: removed as discussed

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
thibaultmg: This function has been removed.

configuration_go/k8sutil/v2/builders.go
thibaultmg: The builder has been removed.

configuration_go/k8sutil/v2/providers.go
thibaultmg: I changed it to handle config maps.

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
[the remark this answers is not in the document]
thibaultmg: This function has been removed.

configuration_go/k8sutil/v2/utils.go
thibaultmg: Merged the versions

configuration_go/k8sutil/v2/builders.go
thibaultmg: Removed the builder

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
```diff
@@ -1 +1,269 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	dataVolumeName   string = "data"
+	defaultHTTPPort  int    = 10902
+	defaultNamespace string = "observatorium"
+	servicePortName  string = "http"
+	defaultImage     string = "quay.io/thanos/thanos"
+	defaultImageTag  string = "v0.31.0"
```
philipgough, line 20: can probably bump this to the latest release?

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
```diff
@@ -1 +1,269 @@
...
+			},
+		},
+		VolumeType: "gp2-csi",
+		VolumeSize: "50Gi",
+	}
+}
+
+// Manifests returns the manifests for the compactor.
+// It includes the statefulset, the service, the service monitor, the service account and the config maps required by the containers.
+func (c *CompactorStatefulSet) Manifests() k8sutil.ObjectMap {
```
philipgough, line 151: Why do we treat Secrets differently here? The output here depends on a Secret and it seems fairly rigid in that regards. Is it out of scope here for some reason?

configuration_go/main.go
```diff
@@ -285,11 +284,47 @@ func main() {
...
+	// Example
+	// Thanos Compactor deployment as statefulSet with sidecar.
+
+	// Set the compactor options.
+	compactorSatefulset := compactor.NewCompactor()
+	compactorSatefulset.Options.LogLevel = "debug"
+	compactorSatefulset.Options.AddExtraOpts("--debug.accept-malformed-index")
+
+	// Set the kube config.
+	compactorSatefulset.ImageTag = "v0.32"
```
philipgough, line 301: missing a minor version?

configuration_go/manifests/compactor-tenant-a/compactor-statefulSet.yaml
```diff
@@ -0,0 +1,176 @@
...
+      name: data
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: 50Gi
+      storageClassName: gp2-csi
+    status: {}
+status:
```
philipgough, line 174: is it possible to strip status from the output? This cant be supplied by the client

configuration_go/main.go
thibaultmg: updated

configuration_go/manifests/compactor/compactor-statefulSet.yaml
```diff
@@ -0,0 +1,177 @@
...
+        - name: AWS_ACCESS_KEY_ID
+          valueFrom:
+            secretKeyRef:
+              key: name
+              name: aws_access_key_id
+        - name: AWS_SECRET_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              key: name
+              name: aws_secret_access_key
```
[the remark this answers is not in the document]
thibaultmg: I have updated the example, putting these creds in the objstore, removing these envs. The objective was more about showing what is possible to customise the container configuration.

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
thibaultmg: This is a mistake. I have added capability for associating secrets to containers. The same way this is done for configMaps.

configuration_go/manifests/compactor-tenant-a/compactor-statefulSet.yaml
philipgough: Yep, totally.  Just an observation and comment as opposed to being a problem.

philipgough (approved):

saswatamcode (approved):
LGTM! Thanks for awesome work on this! 
Let's merge and iterate! 🙂

PR #538, opened 2023-08-24
Add disjointed relabel configurator for compactor in go configuration

thibaultmg:
This helps generating relabel configurations in soft tenancy environments for example.
I think it can be useful in this repo. Let me know what you think.

### DIFF
--- a/configuration_go/abstr/kubernetes/thanos/compactor/relabel.go
+++ b/configuration_go/abstr/kubernetes/thanos/compactor/relabel.go
@@ -1,0 +1,78 @@
+package compactor
+
+import (
+	"fmt"
+	"strings"
+
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+)
+
+// MakeDisjointRelabelConfigForGroups returns a list of RelabelConfig for each group that ensures the action "keep" is applied to all label values within the group.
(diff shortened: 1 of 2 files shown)


moadz:
cc: @philipgough but I believe hard tenancy on buckets negates the need for managing this via disjunct relabel config generator? We can chat more about this in standup, but a hard-compactor tenant with its own tenant means we don't have to worry about hard tenancy compactor. 

That being said, we may still want to revisit this for large soft-multi tenants that need their own compactor args.

thibaultmg:
Ok, then I misunderstood the need. Thought we would have to deal with soft tenancy on a shared bucket. If that is not the case, we can indeed drop this PR.

moadz:
Yup correct @thibaultmg miscommunication on my part. The original plan was to move hard tenant blocks in-place and differentiate the configuration via compactor, but as the receiver hashrings for each hard-tenant are seperate, it's simpler to persist their blocks to a seperate bucket, and then spawn a compactor for each bucket. 

There is a wider issue at hand, which is parallel compactions and how to scale compactor beyond a single process (there is no elegant way to do this currently) but we can solve for this in upstream thanos.

thibaultmg:
Thanks for the details @moadz . Closing the PR as not needed.

PR #543, opened 2023-09-15
Add thanos store to go config

thibaultmg:


configuration_go/abstr/kubernetes/thanos/store/store.go
```diff
@@ -1 +1,265 @@
...
+	TracingConfig                    trclient.TracingConfig          `opt:"tracing.config"`
+	TracingConfigFile                string                          `opt:"tracing.config-file"`
+	WebDisable                       bool                            `opt:"web.disable"`
+	WebDisableCors                   bool                            `opt:"web.disable-cors"`
+	WebExternalPrefix                string                          `opt:"web.external-prefix"`
+	WebPrefixHeader                  string                          `opt:"web.prefix-header"`
+
+	// Extra options not officially supported by the compactor.
+	cmdopt.ExtraOpts
+}
```
douglascamata, line 73: You write all this manually? :O

thibaultmg: No, you copy paste the cli help, comment it, do the first struct field, and then tab tab with github copilot. Or you can ask chatGPT to generate it 😉 . 
You have some light post processing to do, but works well!

configuration_go/schemas/thanos/cache/cache.go
```diff
@@ -0,0 +1,17 @@
...
+
+type IndexCacheProvider string
+
+// Taken from github.com/thanos-io/pkg/store/cache/factory.go
+
+const (
+	INMEMORY  IndexCacheProvider = "IN-MEMORY"
+	MEMCACHED IndexCacheProvider = "MEMCACHED"
+	REDIS     IndexCacheProvider = "REDIS"
+)
```
douglascamata, line 11: Are these used? 🤔

configuration_go/schemas/thanos/cache/cache.go
```diff
@@ -0,0 +1,17 @@
...
+
+// IndexCacheConfig specifies the index cache config.
+type IndexCacheConfig struct {
+	Type   IndexCacheProvider `yaml:"type"`
+	Config interface{}        `yaml:"config"`
```
douglascamata, line 16: Is there a way to avoid the empty interface type here and use something else? Maybe generics to state it could be `MemcachedClientConfig | RedisClientConfig | MemoryCacheConfig`?

configuration_go/schemas/thanos/objstore/objstore.go
```diff
@@ -1,6 +1,6 @@
 package objstore
 
-// Taken from https://github.com/thanos-io/objstore/client/factory.go
+// Taken from github.com/thanos-io/objstore/client/factory.go
```
douglascamata, line 3: Why these changes? I think we have more changes that code editors identify the link and make it clickable if it's prefixed with `http[s]`.

douglascamata (changes requested):
Some minor comments. I like this, it looks pretty neat!

configuration_go/schemas/thanos/objstore/objstore.go
thibaultmg: Because the URL does not resolve, you would have to add `/tree/main` and because I am referencing a package here, not a website. So package notation seemed more accurate to me. And it aligns with what is done in `/schemas` of this repo.

configuration_go/schemas/thanos/cache/cache.go
thibaultmg: What do you mean?
These are the supported types of caches supported by Thanos, extracted from Thanos. So, when you want to set the config as a flag content for a given component of Thanos, you can use the `IndexCacheConfig` and serialize it to yaml.

By the way, the reason all this code is copy pasted form Thanos and not directly imported is because `yaml` struct tags lack the `omitempty` tag and some of these struct also have additional fields not related to the input.

configuration_go/schemas/thanos/objstore/objstore.go
douglascamata: Maybe we should link it to an specific commit then? Because the code in the linked files can change and this comment then become obsolete and wrong. If we link to the commit it will always be correct (unless a force push happens, of course).

configuration_go/schemas/thanos/cache/cache.go
thibaultmg: So the best I could come up with is something like that:

```Go
type cacheConfig interface {
	redis.RedisClientConfig | memcached.MemcachedClientConfig | memory.MemoryCacheConfig
	Type() string
}

// IndexCacheConfig specifies the index cache config.
type IndexCacheConfig[T cacheConfig] struct {
	configType string `yaml:"type"`
	config     T      `yaml:"config"`
}

func NewConfig[T cacheConfig](c T) *IndexCacheConfig[T] {
	return &IndexCacheConfig[T]{
		configType: c.Type(),
		config:     c,
	}
}
```

Is that what you are looking for? If so I can change with that.

configuration_go/schemas/thanos/objstore/objstore.go
thibaultmg: fixed

douglascamata (dismissed):
LGTM

configuration_go/k8sutil/genericdeployment.go
```diff
@@ -35,6 +36,9 @@ type DeploymentGenericConfig struct {
 	ConfigMaps map[string]map[string]string // maps a configmap name to its data of type map[string]string
 	Secrets    map[string]map[string][]byte // maps a secret name to its data of type map[string][]byte
 	Sidecars   []ContainerProvider
+
+	// Post processing of the generated objects (e.g. for adding annotations, modifying monitor namespace, etc.)
+	PostProcess []ObjectProcessor
```
saswatamcode, line 41: I think this is a tricky thing to maintain. The users of this lib will have the freedom to modify the `runtime.Object` map in any way they wish to. 

I think it's just easier for user to get the map, and then modify whatever components they need to in some loop, on _their_ configuration codebase, instead of having to construct functions to pass into struct where they have to conditionally check which element of map to apply to.

saswatamcode (commented):
Thanks for awesome work! Some nits!

configuration_go/abstr/kubernetes/thanos/store/store.go
```diff
@@ -1 +1,271 @@
...
+	GrpcServerMaxConnectionAge       time.Duration                   `opt:"grpc-server-max-connection-age"`
+	GrpcServerTlsCert                string                          `opt:"grpc-server-tls-cert"`
+	GrpcServerTlsClientCa            string                          `opt:"grpc-server-tls-client-ca"`
+	GrpcServerTlsKey                 string                          `opt:"grpc-server-tls-key"`
+	HttpAddress                      net.TCPAddr                     `opt:"http-address"`
+	HttpGracePeriod                  time.Duration                   `opt:"http-grace-period"`
+	HttpConfig                       string                          `opt:"http.config"`
+	IgnoreDeletionMarksDelay         time.Duration                   `opt:"ignore-deletion-marks-delay"`
+	IndexCacheSize                   string                          `opt:"index-cache-size"`
+	IndexCacheConfig                 string                          `opt:"index-cache.config"`
```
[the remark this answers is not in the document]
thibaultmg: I realise I did the contrary for tracing option...

configuration_go/abstr/kubernetes/thanos/store/store.go
```diff
@@ -1 +1,271 @@
...
+	SelectorRelabelConfig            relabel.Config                  `opt:"selector.relabel-config"`
+	SelectorRelabelConfigFile        string                          `opt:"selector.relabel-config-file"`
+	StoreEnableIndexHeaderLazyReader bool                            `opt:"store.enable-index-header-lazy-reader"`
+	StoreGrpcDownloadedBytesLimit    int                             `opt:"store.grps.downloaded-bytes-limit"`
+	StoreGrpcSeriesMaxConcurrency    int                             `opt:"store.grps.series-max-concurrency"`
+	StoreLimitsRequestSamples        int                             `opt:"store.limits.request-samples"`
+	StoreLimitsRequestSeries         int                             `opt:"store.limits.request-series"`
+	SyncBlockDuration                time.Duration                   `opt:"sync-block-duration"`
+	TracingConfig                    trclient.TracingConfig          `opt:"tracing.config"`
+	TracingConfigFile                string                          `opt:"tracing.config-file"`
```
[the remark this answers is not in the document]
thibaultmg: Ok. So I also remove `IndexCacheConfigFile`, `ObjstoreConfigFile`, `RequestLoggingConfigFile`, `SelectorRelabelConfigFile`.

configuration_go/abstr/kubernetes/thanos/store/store.go
```diff
@@ -1 +1,271 @@
...
+	WebDisable                       bool                            `opt:"web.disable"`
+	WebDisableCors                   bool                            `opt:"web.disable-cors"`
+	WebExternalPrefix                string                          `opt:"web.external-prefix"`
+	WebPrefixHeader                  string                          `opt:"web.prefix-header"`
+
+	// Extra options not officially supported by the compactor.
+	cmdopt.ExtraOpts
```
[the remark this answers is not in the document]
thibaultmg: I find it quite useful. This allows using unreferenced flags easily and in an encapsulated way at the time you declare the rest of the options:
`store.Options.AddExtraOpts("--store.caching-bucket.config=...")`

If you don't have that, you must do some post processing on the manifests as the container.Args is generated when calling ` Manifests()` on the store object.

configuration_go/k8sutil/genericdeployment.go
thibaultmg: I agree that you don't need this for basic modifications like changing the serviceMonitor namespace.
But it is trickier when you need to know values defined at the container level.
This is the case that made me do this addition:

- the store container has a side car
- this side car (oauth proxy) consumes a secret name `mytlscerts` that is created by the platform through an annotation on the service
- If I encapsulate the store statefulset configuration in a given function like `makeStore()` and I generate the manifests from another place, I have to return this secret name `mytlscerts` in the function where the `Manifests()` are generated. Because at that point I will be able to do the needed post processing.

How do you think we should handle such cases?

configuration_go/schemas/thanos/cache/memcached/memcached.go
```diff
@@ -0,0 +1,67 @@
+package memcached
+
+import (
+	"time"
+
+	"github.com/thanos-io/thanos/pkg/model"
+)
+
+// Taken from github.com/thanos-io/thanos/pkg/cacheutil/memcached_client.go v0.32.2
```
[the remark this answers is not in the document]
thibaultmg: Ok. I was making the case to @douglascamata that I was referencing packages and not websites. But both are fine to me. I see you both agree, I'll update all these comments.

configuration_go/abstr/kubernetes/thanos/store/store.go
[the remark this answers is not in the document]
thibaultmg: done

configuration_go/schemas/thanos/cache/cache.go
```diff
@@ -0,0 +1,25 @@
...
+	Type() string
+}
+
+// IndexCacheConfig specifies the index cache config.
+type IndexCacheConfig[T cacheConfig] struct {
+	configType string `yaml:"type"`
+	config     T      `yaml:"config"`
+}
+
+func NewConfig[T cacheConfig](c T) *IndexCacheConfig[T] {
```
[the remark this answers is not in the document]
thibaultmg: Now injected in the store options. I had to modify it as I cannot use a generic as a type.

configuration_go/k8sutil/genericdeployment.go
thibaultmg: Removed. Can be handled nicely by consumers.

configuration_go/abstr/kubernetes/thanos/store/store.go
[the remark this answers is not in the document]
saswatamcode: Sorry, misread this, thought it meant store didn't support it. But seems to be compactor. :)

configuration_go/k8sutil/genericdeployment.go
saswatamcode: Yup!

configuration_go/abstr/kubernetes/thanos/store/store.go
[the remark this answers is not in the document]
saswatamcode: Discussed offline, but removing everything except objstore seems to be the way to go! Can change in future!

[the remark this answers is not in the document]
thibaultmg: I am not so sure about this one anymore. Removing those will prevent us from implementing the block sharding strategy with a single statefulset as it need to consume the relabel config as a file generated with a configmap.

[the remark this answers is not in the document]
saswatamcode: Ok discussed offline again, I think it'd be ok to preserve the file variants for all, but consumer, should always prefer struct.

configuration_go/abstr/kubernetes/thanos/store/store.go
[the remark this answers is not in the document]
thibaultmg: oh yes, will update

saswatamcode (approved):
LGTM! Let's actually start using this and iterate!

configuration_go/abstr/kubernetes/thanos/store/store.go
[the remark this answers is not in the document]
thibaultmg: restored as discussed. We can add those guidelines in the documentation later 👍

douglascamata (approved):
:shipit:

thibaultmg:
After using this lib, I found a few quirks that I fixed in the preceding commit before this gets merged:

- in k8utils, for antiAffinity generation, match expressions are sorted to avoid random order when generating the yaml that results in noisy diffs.
- in providers, removed the serviceMonitor namespace target to avoid unnecessary synchronisation with the namespace specified in the `DeploymentGenericConfig` as by default, the service monitor target its own namespace.
- did the same for affinity namespace
- added the grpc port of the store for thanos api that was missing
- now cloning the metaConfig before assigning it to k8s providers to avoid that when doing post processing, a change in one object impacts all of them (like adding an annotation)

Can you please review and hopefully merge these last changes? @douglascamata @saswatamcode 
Thanks!

saswatamcode (dismissed):

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
```diff
@@ -15,55 +16,54 @@ const (
...
+	DownsampleConcurrency              int                         `opt:"downsample.concurrency"`
+	DownsamplingDisable                bool                        `opt:"downsampling.disable"`
+	HashFunc                           string                      `opt:"hash-func"`
+	HttpAddress                        *net.TCPAddr                `opt:"http-address"`
+	HttpGracePeriod                    time.Duration               `opt:"http-grace-period"`
+	HttpConfig                         string                      `opt:"http.config"`
+	LogFormat                          common.LogFormat            `opt:"log.format"`
+	LogLevel                           common.LogLevel             `opt:"log.level"`
+	MaxTime                            *common.TimeOrDurationValue `opt:"max-time"`
+	MinTime                            *common.TimeOrDurationValue `opt:"min-time"`
```
douglascamata, line 49: I'm not the biggest fan of naming packages like this (`common`, `utils`, etc). But this is not a blocker at all, we can change it later. I would suggest a `log` and `time` packages.

configuration_go/schemas/thanos/cache/cache.go
```diff
@@ -0,0 +1,35 @@
...
+// IndexCacheConfig specifies the index cache config.
+type IndexCacheConfig struct {
+	ConfigType string      `yaml:"type"`
+	Config     interface{} `yaml:"config"`
+}
+
+func (c IndexCacheConfig) String() string {
+	ret, err := yaml.Marshal(c)
+	if err != nil {
+		mimic.Panicf("error mashalling IndexCacheConfig to yaml: %v", err)
```
douglascamata, line 20: Feels a bit weird that a method to transform something to a string needs a very specific helper from a third party library to panic and all the helper does is `panic(fmt.Sprintf("mimic: "+format, a...))`. I think it's something we could think about in the future. It raises an eyebrow but it's fine for now. 🤔

configuration_go/schemas/thanos/common/common.go
```diff
@@ -0,0 +1,40 @@
...
+	case tdv.Time != nil:
+		return tdv.Time.String()
+	case tdv.Dur != nil:
+		if v := *tdv.Dur; v < 0 {
+			return "-" + (-v).String()
+		}
+		return tdv.Dur.String()
+	}
+
+	return "nil"
```
douglascamata, line 39: This is a bit weird in the Go world. Could we instead return the empty value of the `string` type instead? I think Go developers will be more accustomed to this.

configuration_go/schemas/thanos/reqlogging/reqlogging.go
```diff
@@ -0,0 +1,48 @@
...
+// Taken from https://github.com/thanos-io/thanos/blob/release-0.32/pkg/logging/yaml_parser.go
+
+type LogLevel string
+
+const (
+	LogLevelDebug LogLevel = "DEBUG"
+	LogLevelInfo  LogLevel = "INFO"
+	LogLevelWarn  LogLevel = "WARNING"
+	LogLevelError LogLevel = "ERROR"
+)
```
douglascamata, line 12: These constants are almost the same as the ones in `configuration_go/schemas/thanos/common/common.go`, except that these are uppercase. Could some code be reused, maybe? Not a blocker for the PR.

configuration_go/schemas/thanos/cache/redis/redis.go
```diff
@@ -0,0 +1,103 @@
+package redis
+
+import (
+	"time"
+
+	"github.com/thanos-io/thanos/pkg/model"
```
douglascamata, line 6: We can use the local import now from `configuration_go/schemas/thanos/units/units.go`.

douglascamata (changes requested):
This is shaping very well. 

Left a few comments and some are no blockers, so please feel free to only address them in a follow up PR in the future.

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
thibaultmg: Yup I agree.

configuration_go/schemas/thanos/cache/cache.go
thibaultmg: Yes I agree, I would rather simply use the standard panic without any ref to mimic. There were some other `mimic.Panic` in other places, followed blindly this convention. Will change.

configuration_go/schemas/thanos/common/common.go
thibaultmg: I just copy pasted the relevant part for us from this lib. I agree that this does not make sense for our case. They are probably using this do distinguish a zero value of a Time or Duration from a nil. But we're not using this nil. Will remove.

douglascamata: Hummm, we can ignore if it'll break something because we're changing this. 🤔

configuration_go/schemas/thanos/cache/cache.go
thibaultmg: @saswatamcode are you ok with this change?

saswatamcode: Sure! As long as we panic, I'm good

configuration_go/abstr/kubernetes/thanos/compactor/compactor.go
thibaultmg: done

configuration_go/schemas/thanos/cache/cache.go
thibaultmg: done

configuration_go/schemas/thanos/common/common.go
thibaultmg: Let's leave like that then. It's just copy paste.

configuration_go/schemas/thanos/cache/redis/redis.go
thibaultmg: done

douglascamata (approved):
LGTM!

saswatamcode (approved):
🚀

PR #544, opened 2023-10-06
Add receive to Go configs

thibaultmg:
Adds the receive component to the Go configuration files.

### DIFF
--- a/configuration_go/abstr/kubernetes/thanos/receive/controller.go
+++ b/configuration_go/abstr/kubernetes/thanos/receive/controller.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	thanosReceiveControllerImage string = "quay.io/observatorium/thanos-receive-controller"
+	ThanosReceiveControllerImage string = "quay.io/observatorium/thanos-receive-controller"
 )
 
 type ControllerOptions struct {
@@ -43,7 +43,7 @@ func NewController() *Controller {
 
 	genericDeployment := k8sutil.DeploymentGenericConfig{
 		Name:            fmt.Sprintf("%s-%s", commonLabels[k8sutil.InstanceLabel], commonLabels[k8sutil.NameLabel]),
-		Image:           thanosReceiveControllerImage,
+		Image:           ThanosReceiveControllerImage,
 		ImageTag:        "main",
 		Replicas:        1,
 		CommonLabels:    commonLabels,
--- a/configuration_go/abstr/kubernetes/thanos/receive/controller.go
+++ b/configuration_go/abstr/kubernetes/thanos/receive/controller.go
@@ -1,0 +1,198 @@
+package receive
+
+import (
+	"fmt"
+
+	cmdopt "github.com/observatorium/observatorium/configuration_go/abstr/kubernetes/cmdoption"
+	"github.com/observatorium/observatorium/configuration_go/k8sutil"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	thanosReceiveControllerImage string = "quay.io/observatorium/thanos-receive-controller"
+)
+
+type ControllerOptions struct {
+	ConfigMapName          string `opt:"--configmap-name"`
+	ConfigMapGeneratedName string `opt:"--configmap-generated-name"`
+	FileName               string `opt:"--file-name"`
+	Namespace              string `opt:"--namespace"`
+}
+
+type Controller struct {
+	Options ControllerOptions
+	k8sutil.DeploymentGenericConfig
+}
+
+func NewController() *Controller {
+	options := ControllerOptions{
+		ConfigMapName:          "observatorium-thanos-receive-controller",
+		ConfigMapGeneratedName: "observatorium-thanos-receive-controller-generated",
+		FileName:               "hashrings.json",
+		Namespace:              "observatorium",
+	}
+
+	commonLabels := map[string]string{
+		k8sutil.NameLabel:      "thanos-receive-controller",
+		k8sutil.InstanceLabel:  "observatorium",
+		k8sutil.PartOfLabel:    "observatorium",
+		k8sutil.ComponentLabel: "kubernetes-controller",
+	}
+
+	genericDeployment := k8sutil.DeploymentGenericConfig{
+		Name:            fmt.Sprintf("%s-%s", commonLabels[k8sutil.InstanceLabel], commonLabels[k8sutil.NameLabel]),
+		Image:           thanosReceiveControllerImage,
+		ImageTag:        "main",
+		Replicas:        1,
+		CommonLabels:    commonLabels,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Env: []corev1.EnvVar{
+			k8sutil.NewEnvFromField("NAMESPACE", "metadata.namespace"),
+		},
+		PodResources: k8sutil.NewResourcesRequirements("10m", "24Mi", "64m", "128Mi"),
+		ConfigMaps:   map[string]map[string]string{},
+		Secrets:      map[string]map[string][]byte{},
+	}
+
+	return &Controller{
+		options,
+		genericDeployment,
+	}
+}
+
+func (c *Controller) Manifests() k8sutil.ObjectMap {
+	container := c.makeContainer()
+	ret := k8sutil.ObjectMap{}
+
+	commonObjectMeta := k8sutil.MetaConfig{
+		Name:      c.Name,
+		Labels:    c.CommonLabels,
+		Namespace: c.Namespace,
+	}
+	commonObjectMeta.Labels[k8sutil.VersionLabel] = container.ImageTag
+
+	pod := &k8sutil.Pod{
+		TerminationGracePeriodSeconds: &c.TerminationGracePeriodSeconds,
+		SecurityContext:               c.SecurityContext,
+		ServiceAccountName:            c.Name,
+		ContainerProviders:            []k8sutil.ContainerProvider{container},
+	}
+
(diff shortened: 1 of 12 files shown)


philipgough (dismissed):
lgtm

configuration_go/schemas/thanos/reqlogging/reqlogging.go
```diff
@@ -17,6 +19,15 @@ type RequestConfig struct {
 	Options OptionsConfig       `yaml:"options,omitempty"`
 }
 
+func (c RequestConfig) String() string {
+	// We use "gopkg.in/yaml.v2" instead of "github.com/ghodss/yaml" for correct formatting of this config.
```
saswatamcode, line 23: This comment can be put on the function as a goDoc comment, so that user can see it?

configuration_go/schemas/thanos/tracing/client/client.go
```diff
@@ -17,3 +19,12 @@ type TracingConfig struct {
 	Type   TracingProvider `yaml:"type"`
 	Config interface{}     `yaml:"config"`
 }
+
+func (c TracingConfig) String() string {
+	// We use "gopkg.in/yaml.v2" instead of "github.com/ghodss/yaml" for correct formatting of this config.
```
saswatamcode, line 24: This comment can be put on the function as a goDoc comment, so that user can see it?

configuration_go/schemas/thanos/option/option.go
```diff
@@ -0,0 +1,40 @@
+package option
+
+// ConfigFile is a struct that represents a config file.
+// It includes data needed to mount the file as a configMap or secret in a container.
+type ConfigFile[T any] struct {
+	Name      string
+	Value     T
```
saswatamcode, line 7: Does this need to be a generic?

configuration_go/schemas/thanos/objstore/objstore.go
```diff
@@ -8,6 +10,15 @@ type BucketConfig struct {
 	Prefix string      `yaml:"prefix,omitempty"`
 }
 
+func (c BucketConfig) String() string {
+	// We use "gopkg.in/yaml.v2" instead of "github.com/ghodss/yaml" for correct formatting of this config.
```
saswatamcode, line 14: Ditto

configuration_go/schemas/thanos/cache/cache.go
```diff
@@ -16,6 +16,7 @@ type IndexCacheConfig struct {
 }
 
 func (c IndexCacheConfig) String() string {
+	// We use "gopkg.in/yaml.v2" instead of "github.com/ghodss/yaml" for correct formatting of this config.
```
saswatamcode, line 19: Ditto

configuration_go/abstr/kubernetes/thanos/store/store.go
```diff
@@ -119,7 +119,6 @@ func NewStore() *StoreStatefulSet {
 			Replicas:             1,
 			PodResources:         k8sutil.NewResourcesRequirements("500m", "1", "200Mi", "400Mi"),
 			Affinity:             *k8sutil.NewAntiAffinity(nil, labelSelectors),
-			SecurityContext:      k8sutil.GetDefaultSecurityContext(),
```
saswatamcode, line 122: Why are we removing this?

configuration_go/abstr/kubernetes/thanos/receive/controller.go
```diff
@@ -0,0 +1,198 @@
...
+
+	cmdopt "github.com/observatorium/observatorium/configuration_go/abstr/kubernetes/cmdoption"
+	"github.com/observatorium/observatorium/configuration_go/k8sutil"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	thanosReceiveControllerImage string = "quay.io/observatorium/thanos-receive-controller"
```
saswatamcode, line 14: We can probably keep this exported?

configuration_go/abstr/kubernetes/thanos/receive/receive.go
```diff
@@ -0,0 +1,514 @@
...
+
+// Label represents a single label configuration.
+type Label struct {
+	Key   string
+	Value string
+}
+
+func (l Label) String() string {
+	return l.Key + "=" + l.Value
+}
```
saswatamcode, line 47: Do we need to redefine this? We already use prometheus libs, so I think we can just use that instead?

configuration_go/abstr/kubernetes/thanos/receive/receive.go
```diff
@@ -0,0 +1,514 @@
...
+
+// HashringConfig represents a single hashring configuration.
+type HashringConfig struct {
```
saswatamcode, line 50: Hmm, while I think this is fine, I worry that we will have to keep copying structs like this a lot. I wonder if we should just import/vendor thanos here then?

saswatamcode (dismissed):
Thanks! Mostly LGTM!

Just a few questions/comments.

configuration_go/abstr/kubernetes/thanos/store/store.go
thibaultmg: I had issues for statfulsets having this. The user group is considered invalid and the PV could not be mounted... I don't know where this policy comes from though.

configuration_go/schemas/thanos/option/option.go
thibaultmg: The purpose of this struct is to embed details about the source of the file (CM or secret) if a ".*ConfigFile" option is used. It avoids having some logic about this spread at different parts of the file of a component.
The value is typed by the option being defined. Here is a snippet illustrating usage where value type is `HashRingsConfig`:

```
type receiveHashringConfigFile = option.ConfigFile[HashRingsConfig]

func NewReceiveHashringConfigFile(name string, value HashRingsConfig) *receiveHashringConfigFile {
	return option.NewConfigFile("/etc/thanos/hashring", "hashring.json", name, value)
}

type ReceiveOptions struct {
    	ReceiveHashringsFile                *receiveHashringConfigFile `opt:"receive.hashrings-file"`
}
```

The client defines this option by calling `NewReceiveHashringConfigFile` that defines the CM name and value of type `HashRingsConfig`. From there, everything can be created, from the CM to the VolumesMounts.

Does it make sense?

configuration_go/abstr/kubernetes/thanos/receive/receive.go
thibaultmg: Yes we can. I just wanted to avoid the client to have to import prometheus to define this option. But I don't mind removing that. Do you confirm removal?

configuration_go/abstr/kubernetes/thanos/receive/receive.go
thibaultmg: I agree it is not ideal. The issue with the Thanos version is that it does not implement the Stringer interface to marshal to json. This is needed for cmdopt, because it has no idea about the target being json or yaml for a struct.
And we had to redefine many of these type here already because of missing yaml tags or because they include so exported fields for internal use only...

configuration_go/schemas/thanos/reqlogging/reqlogging.go
thibaultmg: done

configuration_go/schemas/thanos/tracing/client/client.go
thibaultmg: done

configuration_go/schemas/thanos/objstore/objstore.go
thibaultmg: done

configuration_go/schemas/thanos/cache/cache.go
thibaultmg: done

configuration_go/abstr/kubernetes/thanos/receive/controller.go
thibaultmg: done

configuration_go/abstr/kubernetes/thanos/receive/receive.go
thibaultmg: or I could add a tag option for cmdopt defining the target serialisation, but I find it even less ideal.

configuration_go/schemas/thanos/option/option.go
saswatamcode: Ack! I think it does, thanks for explaining!

configuration_go/abstr/kubernetes/thanos/receive/receive.go
saswatamcode: Yup, I think importing prometheus for client using this lib would be ok.

configuration_go/abstr/kubernetes/thanos/receive/receive.go
saswatamcode: Yeah, probably something we need to deal with in the future.

configuration_go/abstr/kubernetes/thanos/receive/receive.go
thibaultmg: Ok, so actually this is trickier. 
When there is an option that can be repeated, it is set as a slice of the type for cmdopt, like `[]Label`. This way, it understands that the option must be repeated for each element. Then, each element of the slice must implement the `Stringer` interface. But in the prometheus labels case, it is defined on the slice type `Labels`:
```
type Labels []Label
func (ls Labels) String() string {
...
```

So I think we must keep it. I don't see another solution here. 😞 
I am reverting the commits I pushed for this.

douglascamata (approved):
LGTM, sir :shipit:

saswatamcode (approved):
LGTM!

PR #546, opened 2023-10-26
Go configs: Add memcached deployment and bucket cache config

thibaultmg:
* Adds memcached deployments
* Sets optional k8s objects as pointers to avoid empty objects generation that causes validation issues (pod affinity, container probes, security cfg)


philipgough (dismissed):
lgtm

philipgough (approved):

PR #547, opened 2023-11-24
Go config first version

thibaultmg:
This big PR: 

- adds a few missing components like alertManager
- Cleans all supporting packages for the go templates generation
- Upgrades go dependencies
- Aligns all components generation with the same layout
- Refactor main file to use new templates

This is kind of the go_config v1. Let me know what you think 🙂


douglascamata (approved):
Awesome work, @thibaultmg. I liked the reorganization of the folders. 

Can't wait to start deploying our things from this. 🚀

PR #549, opened 2024-02-09
Go-config: add tenants-file option to rules syncer

thibaultmg:
The Tenants config struct being in the main package of the syncer, I cannot import it 🤦 . So I am redeclaring it here. That's not a big issue as it is a basic struct for now. Next dev in the syncer, we should move this config into its own sub-package.

### DIFF
--- a/configuration_go/abstr/kubernetes/thanos/ruler/syncer.go
+++ b/configuration_go/abstr/kubernetes/thanos/ruler/syncer.go
@@ -5,9 +5,11 @@ import (
 	"path/filepath"
 
 	"github.com/observatorium/observatorium/configuration_go/kubegen/cmdopt"
+	"github.com/observatorium/observatorium/configuration_go/kubegen/containeropts"
 	kghelpers "github.com/observatorium/observatorium/configuration_go/kubegen/helpers"
 	"github.com/observatorium/observatorium/configuration_go/kubegen/workload"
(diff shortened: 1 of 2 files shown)


philipgough (approved):
lgtm

configuration_go/abstr/kubernetes/thanos/ruler/syncer.go
```diff
@@ -5,29 +5,52 @@ import (
...
+type TenantConfig struct {
+	ID string `yaml:"id"`
+}
+
+func NewTenantsConfigFile(value *TenantsConfig) *containeropts.ConfigResourceAsFile {
+	ret := containeropts.NewConfigResourceAsFile("/etc/thanos-rules-syncer/tenants", "config.yaml", "objstore", "observatorium-rules-syncer-tenants")
+	if value != nil {
+		valueBytes, err := yaml.Marshal(value)
+		if err != nil {
+			panic(err)
```
philipgough, line 33: is panicking here the best solution?

thibaultmg: It is for manifests file generation. So I assume that you want to know right away if something is wrong. 
You are thinking of another approach?

philipgough: no thats all good

